Issue #74, opened 2014-06-26
Multipart form requests need to use \r\n instead of only \n 


ErikPettersson:
Both multiparty and parted (nodejs body parsing middlewares) do not accept Dredd generated multiform posts, due to line endings not being OK.

It seems they should be \r\n and Dredd always sends \n (perhaps because API Blueprint only supports that, so that's the actual data it's told to send). 

So, feature request:
Dredd should make sure that the request are formatted decently. (:

Example request:

```
+ Request (multipart/form-data;boundary=---BOUNDARY)
    + Headers

            Authorization: qwertyqwerty

    + Body

            ---BOUNDARY
            Content-Disposition: form-data; name="json"


            {"name":"test"}
            ---BOUNDARY
            Content-Disposition: form-data; name="image"; filename="filename.jpg"
            Content-Type: image/jpeg

            data
            ---BOUNDARY--
```

(See also: https://github.com/apiaryio/api-blueprint/issues/100)

PR #118, opened 2014-11-07
Multipart form requests need to use \r\n instead of only \n #74

rikardq:
Hello, made a patch for replacing the \n to \r\n for multipart request for issue #74 
Added test. 

Regards Rikard


netmilk:
This is solved in #148

netmilk:
@rikardq It's based on your work, thanks!

PR #125, opened 2015-01-06
Multi blueprint processing

### LINKED ISSUE #89: Better error message for missing blueprint
Get rid with node.js error message and replace with more human understandable message.
 
```
dredd blueprint.md http://api.myservice.tld
ENOENT, open 'blueprint.md'
Error: ENOENT, open 'blueprint.md'

```
vs.
```
Hey--we couldn’t find the blueprint called blueprint.md
```

netmilk:
- Support for multiple blueprints with --path
- Expanding glob blueprint paths
- Support for hooks for multiple requests and responses (examples)
- Support for hooks for anonymous resources and methods
- Added --only cli option for selecting specific transactions to run
- Better error message for missing blueprint, closes #89 
- Removed support for node.js v0.8
- Updated decayng dependencies


README.md
```diff
@@ -147,50 +147,57 @@ If `beforeAll` and `afterAll` are called multiple times, the callbacks are execu
...
-                                                                     [default: true]
-      --level, -l          The level of logging to output. Options: silly, debug,
-                           verbose, info, warn, error.
-                                                                   [default: "info"]
-      --timestamp, -t      Determines whether console output should include
-                           timestamps.
-                                                                    [default: false]
-      --help               Show usage information.
-      --version            Show version number.
+  --dry-run, -y        Do not run any real HTTP transaction, only parse
```
ecordell: It looks like this isn't indented enough to be considered a code block by markdown (you can see in the preview for the file).

bin/dredd
```diff
@@ -33,18 +33,38 @@ if (argv._[1] === undefined) {
...
 }
+
 if (argError) {
   console.error("\n");
   optimist.showHelp(fn=console.error);
   process.exit(1);
 }
 
+// transform path and p argument to array if it's not
+if(!Array.isArray(argv['path'])){
```
ecordell: There are methods in `applyConfiguration` to take care of this, see: https://github.com/apiaryio/dredd/blob/master/src/apply-configuration.coffee#L43

The transformation could be removed from here add two lines added in apply-configuration:
```javascript
 configuration.options.path = coerceToArray(configuration.options.path)
 configuration.options.only = coerceToArray(configuration.options.only)
```

Unless I'm misunderstanding and there's a reason they need to be parsed before the rest of the options?

src/dredd.coffee
```diff
@@ -28,37 +28,85 @@ class Dredd
...
-      
-      runtime = blueprintAstToRuntime result['ast']
-      runtimeError = handleRuntimeProblems runtime
-      return callback(runtimeError, stats) if runtimeError
-
-      @runner.run runtime['transactions'], () =>
-        @transactionsComplete(callback)
+    config.files = []
+
+    # expand all globs
```
ecordell: This is just my opinion, but I think this section would be more readable if every part that starts with a `# descriptive name` comment were replaced with a call to `descriptiveName()` function. A couple of cents from the peanut gallery.

test/fixtures/multiple-examples.apib
```diff
@@ -0,0 +1,37 @@
...
+
+- Request (application/json)
+  + Parameters
+    - id (number, `3`)
+
+- Response 200 (application/json; charset=utf-8)
+
+    [
+      {
+        "type": "bulldozer",
```
kuba-kubula: Should the body be different from the previous Example-Action request-response pair, so it can be actually tested, and the response / request are not taken from the first example-action pair?

bin/dredd
netmilk: This will work only for `only` arg, because some terminals are expanding globs and concating them as CLI arguments and I need [fill](https://github.com/apiaryio/dredd/pull/125/files#diff-1250f065cb350c09960352445f4d42ddR53) those into `option['path']` which couldn't be an array at that moment.

src/dredd.coffee
netmilk: I'm sorry, but I don't get it. What do you mean with `# descriptive name` comment. This part is expanding globs  paths of blueprints.


>  A couple of cents from the peanut gallery.

:sunglasses: LOL!

ecordell: Sorry! I should've described it better.

I just meant that I personally would find it easier to read (since it's a fairly long method) if it were split up into parts - and I thought the comments describing the parts suggested natural places to split.

A quick pseudo-example:

```coffeescript
expandAllGlobs(config)
removeDuplicateFilenames(config)
loadAllFiles(config)
parseAllBlueprints(config)
runAllTransactions(config)
```
Then each "task" would be self-contained in its own method/module/file/etc.

netmilk: :+1:  AKA despaghettisation :)

test/fixtures/multiple-examples.apib
netmilk: Bodies can be the same, they are not subject of test. If they will, it have to change.

Issue #124, opened 2014-12-05
Dredd Hangs When Using Body Parser and Editing Body in Hook


smizell:
If you do the following steps, Dredd will hang when running.

1. Use `body-parser` with express.js.
1. Use a `before` hook
1. In the hook, change the body to some string
1. Run Dredd with that hook

I've created a Gist here to duplicate:
https://gist.github.com/smizell/b054ebdd34a9bbd6eec3

PR #128, opened 2015-01-13
Computing content-length in executeTransasction, fixing #124

netmilk:
Related to #124, computing content-length in executeTransaction not in configureTransaction, because transaction body can be modified in some before hook and content-length sensitive backends can hang forever.

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -112,7 +112,7 @@ class TransactionRunner
       caseInsensitiveMap[key.toLowerCase()] = key
 
     if not caseInsensitiveMap['content-length'] and transaction.request['body'] != ''
-      transaction.request.headers['Content-Length'] = transaction.request['body'].length
+      transaction.request.headers['Content-Length'] = Buffer.byteLength(transaction.request['body'], 'utf8')
 
     requestOptions =
       host: transaction.host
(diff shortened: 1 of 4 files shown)


src/transaction-runner.coffee
```diff
@@ -112,6 +104,16 @@ class TransactionRunner
...
 
+    # Add length of body if no Content-Length present
+    # Doing here instead of in configureTransaction, because request body can be edited in before hook
+
+    caseInsensitiveMap = {}
+    for key, value of transaction.request.headers
+      caseInsensitiveMap[key.toLowerCase()] = key
+
+    if not caseInsensitiveMap['content-length'] and transaction.request['body'] != ''
+      transaction.request.headers['Content-Length'] = transaction.request['body'].length
```
kuba-kubula: Can we be assured that `transaction.request['body']` will always be only String, without any multi-byte chars?

I'd say this will fail when used together with `utf-8` characters.

easily fixed with `Buffer.byteLength(transaction.request['body'], 'utf8')`

kuba-kubula:
Looks good, just the `Buffer.byteLength` stuff would make me worried in future.

pocketmax:
Does anyone know the status of this?

netmilk:
@pocketmax This was fixed, merged and released. Are you experiencing any inconvenience regarding this issue?

PR #144, opened 2015-02-12
Allow to pass config data directly - via "directInput" object

### LINKED ISSUE #143: Use Dredd within Mocha
Hi,

I would like to use Dredd within Mocha.

My idea is that Dredd would be able to do all the heavy lifting for me; by heavy lifting I mean sending HTTP requests and comparing actual responses to the expected ones.

There are two feature requests.

* Ability to pass in a blueprint file

    ```CoffeeScript
    config =
      blueprint: 'FORMAT: ...'

    dredd = new Dredd config
    ```

* Ability to pass in a list of hooks

    ```CoffeeScript
    config =
      hooks:
        'Users > Users Collection > Update a User':
          before: (transaction) ->
            (...)

    dredd = new Dredd config
    ```

What do you think? Does it make sense to use Dredd in such way, @netmilk?

---

## Example

```CoffeeScript
describe 'When I create a new user', ->
  user = null

  before (done) ->
    user = new User()
    user.save done

  describe 'And I run Dredd against the API', ->
    returnedStats = null

    before (done) ->
      blueprint = """
      (...)

      ### Update a User [PUT]

      + Request (application/json)

              {
                "name": "František"
              }

      + Response 200 (application/json)

              {
                "name": "František"
              }

      + Request (application/json)

              {
                "age": "19"
              }

      + Response 200 (application/json)

              {
                "age": "19"
              }
      """

      config = {
        blueprint
        server: '127.0.0.1:9999'
        hooks:
          'Users > Users Collection > Update a User':
            before: (transaction) ->
              transaction.headers['X-User-Token'] = user.get 'token'
      }

      dredd = new Dredd config
      dredd.run (err, stats) ->
        returnedStats = stats
        done err

    it 'Everything should work fine', ->
      assert.ok returnedStats
```

kuba-kubula:
Closes #143

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "advisable": "~0.2.0",
     "async": "~0.9.0",
+    "clone": "^0.2.x",
     "coffee-script": "^1.9.0",
     "file": "~0.2.2",
     "gavel": "~0.4.0",
@@ -35,7 +36,6 @@
   "devDependencies": {
     "body-parser": "^1.10.2",
     "chai": "^1.10.0",
-    "clone": "^0.x",
     "codo": "2.0.9",
     "coffee-coverage": "~0.4.4",
     "coffee-errors": "~0.8.6",
--- a/src/dredd.coffee
+++ b/src/dredd.coffee
@@ -34,9 +34,23 @@ class Dredd
     config = @configuration
     stats = @stats
 
+    configDataIsEmpty = true
+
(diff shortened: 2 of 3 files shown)


Baggz:
<3

netmilk:
Why is this feature not using already existing `options.data` property? I strongly prefer using existing options with same functionality over adding new one.

netmilk:
Nice!

PR #145, opened 2015-02-13
Load blueprint-file from provided URL from http/https

kuba-kubula:
- with a timeout of 5 secs

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "optimist": "~0.6.1",
     "protagonist": "~0.17.1",
     "proxyquire": "^1.3.1",
-    "request": "",
+    "request": "^2.53.0",
     "setimmediate": "^1.0.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
     "optimist": "~0.6.1",
     "protagonist": "~0.17.1",
     "proxyquire": "^1.3.1",
+    "request": "",
```
netmilk: Please set specific version here.

netmilk:
Nice, thanks! But I'm nitpicking... :) This PR extends functionality of `Dredd` class, but you are not introducing any new unit tests for this in `Dredd` class unit tests, despite the fact you have tested this new functionality very nicely in the integration tests for the CLI, which I really appreciate. .

netmilk:
I also miss some update of README taking account of this change.

PR #148, opened 2015-02-16
Adding CR to multipart requests based on multipart content-type header, ...

### LINKED ISSUE #74: Multipart form requests need to use \r\n instead of only \n

netmilk:
...Closes #74 #118 #29

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -227,9 +227,10 @@ class TransactionRunner
     caseInsensitiveRequestHeaders['content-type']?.indexOf("multipart") > -1
 
   replaceLineFeedInBody: (transaction, requestOptions) =>
-    transaction.request['body'] = transaction.request['body'].replace(/\n/g, '\r\n')
-    transaction.request['headers']['Content-Length'] = transaction.request['body'].length
-    requestOptions.headers = transaction.request['headers']
+    if transaction.request['body'].indexOf('\r\n') == -1
+      transaction.request['body'] = transaction.request['body'].replace(/\n/g, '\r\n')
+      transaction.request['headers']['Content-Length'] = Buffer.byteLength(transaction.request['body'], 'utf8')
+      requestOptions.headers = transaction.request['headers']
 
 
 module.exports = TransactionRunner
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -17,7 +17,7 @@ logger = require './logger'
 
 
 String::startsWith = (str) ->
-    return this.slice(0, str.length) is str
+  return this.slice(0, str.length) is str
 
 class TransactionRunner
   constructor: (@configuration) ->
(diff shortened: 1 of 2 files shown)


src/transaction-runner.coffee
```diff
@@ -217,4 +220,16 @@ class TransactionRunner
         configuration.emitter.emit 'test error', error, test if error
         return callback()
 
+  isMultipart: (requestOptions) =>
+    caseInsensitiveRequestHeaders = {}
+    for key, value of requestOptions.headers
+      caseInsensitiveRequestHeaders[key.toLowerCase()] = value
+    caseInsensitiveRequestHeaders['content-type']?.indexOf("multipart") > -1
```
kuba-kubula: I'd love to use https://www.npmjs.com/package/caseless here.

`caseless(requestOptions.headers).get('content-type')?.indexOf('multipart') > -1` ... but I can see we might refactor more parts of it soon.

src/transaction-runner.coffee
```diff
@@ -217,4 +220,16 @@ class TransactionRunner
...
+
+  replaceLineFeedInBody: (transaction, requestOptions) =>
+    transaction.request['body'] = transaction.request['body'].replace(/\n/g, '\r\n')
+    transaction.request['headers']['Content-Length'] = transaction.request['body'].length
```
kuba-kubula: please use `Buffer.byteLength` here too (as seen before)

src/transaction-runner.coffee
```diff
@@ -217,4 +220,16 @@ class TransactionRunner
...
         return callback()
 
+  isMultipart: (requestOptions) =>
+    caseInsensitiveRequestHeaders = {}
+    for key, value of requestOptions.headers
+      caseInsensitiveRequestHeaders[key.toLowerCase()] = value
+    caseInsensitiveRequestHeaders['content-type']?.indexOf("multipart") > -1
+
+  replaceLineFeedInBody: (transaction, requestOptions) =>
+    transaction.request['body'] = transaction.request['body'].replace(/\n/g, '\r\n')
```
kuba-kubula: My only hope is that if you do modify the body during hooks, and you do this ^^^ `replace` by yourself, you won't introduce double `\r\r` into the body.
E.g.: 
1. user has hooks to modify (replace `\n` with `\r\n`) request body before transaction happens
2. we will do this replacement again

So I'll propose to check that the body does not contain any `\r` character. If it does, we don't have to do the replacement by ourself.

cc. @netmilk

netmilk: Added scenario for this

src/transaction-runner.coffee
netmilk: Fixed

PR #154, opened 2015-02-23
Passing custom parameters to Reporters via JS API, logic from bin script to CoffeeScript

kuba-kubula:
Because `clone` cannot clone `process.stdout`, custom configuration is passed to `Dredd` class as a second argument

### DIFF
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -1,0 +1,155 @@
+path = require 'path'
+optimist = require 'optimist'
+console = require 'console'
+
+parsePackageJson = require './parse-package-json'
+Dredd = require './dredd'
+
+version = parsePackageJson path.join(__dirname, '../package.json')
+
+class DreddCommand
+  constructor: (options = {}, @cb) ->
+    # support to call both 'new DreddCommand()' and/or 'DreddCommand()'
+    # without the "new" keyword
+    if not (@ instanceof DreddCommand)
+      return new DreddCommand(options, @cb)
+
+    {@exit, @custom} = options
+
+    @custom ?= {}
+
+    if not @custom.cwd or typeof @custom.cwd isnt 'string'
+      @custom.cwd = process.cwd()
+
+    if not @custom.argv or not Array.isArray @custom.argv
+      @custom.argv = []
+
+    @finished = false
+
+  warmUp: ->
+    @finished = false
+
+    @optimist = optimist(@custom['argv'], @custom['cwd'])
+
+    @optimist.usage(
+      """
+      Usage:
+        dredd <path or URL to blueprint> <api_endpoint> [OPTIONS]
+
+      Example:
+        dredd ./apiary.md http://localhost:3000 --dry-run
+      """
+    )
+    .options(Dredd.options)
+    .wrap(80)
+
+    @argv = @optimist.argv
+
+    argError = false
+
+    if not @cb
+      @exit and (@exit is process.exit) and @sigIntEventAdd = true
+      @_processExit = @exit or process.exit
+    else
+      @_processExit = ((code) ->
+        if @finished then return
+        @finished = true
+        if @sigIntEventAdded
+          process.removeEventListener 'SIGINT', @commandSigInt
+        @cb code
+        return @
+      ).bind @
+
+    if @argv.help is true
+      @optimist.showHelp(console.error)
+      return @_processExit(0)
+
+    if @argv.version is true
+      console.log version
+      return @_processExit(0)
+
+    if not @argv._[0]?
+      console.error("\nError: Must specify path to blueprint file.")
+      argError = true
+
+    if not @argv._[1]?
+      console.error("\nError: Must specify api endpoint.")
+      argError = true
+
+    if argError
+      console.error("\n")
+      @optimist.showHelp(console.error)
+      return @_processExit(1)
+
+    # transform path and p argument to array if it's not
+    if !Array.isArray(@argv['path'])
+      @argv['path'] = @argv['p'] = [@argv['path']]
+
+    configurationForDredd = @initConfig()
+    @dreddInstance = @initDredd configurationForDredd
+    return @
+
+  lastArgvIsApiEndpoint: ->
+    # some shells are automatically expanding globs and concating result as arguments
+    # so I'm taking last argument as API endpoint server URL and removing it forom optimist's args
+    @server = @argv._[@argv._.length - 1]
+    @argv._.splice(@argv._.length - 1, 1)
+    return @
+
+  takeRestOfParamsAsPath: ->
+    # and rest of arguments concating to 'path' and 'p' opts, duplicates are filtered out later
+    @argv['p'] = @argv['path'] = @argv['path'].concat(@argv._)
+    return @
+
+  initConfig: ->
+    @
+    .lastArgvIsApiEndpoint()
+    .takeRestOfParamsAsPath()
+
+    configuration =
+      'server': @server
+      'options': @argv
+
+    # push first argument (without some known configuration --key) into paths
+    configuration.options.path ?= []
+    configuration.options.path.push @argv._[0]
+
+    configuration.custom = @custom
+
+    return configuration
+
+  initDredd: (configuration) ->
+    dredd = new Dredd configuration
+    return dredd
+
+  commandSigInt: ->
+    console.log "\nShutting down from SIGINT (Ctrl-C)"
+    @dreddInstance.transactionsComplete => @_processExit(0)
+
+  takeoff: ->
+    if @finished then return
+
+    @runDredd @dreddInstance
+
+  runDredd: (dreddInstance) ->
+    if @sigIntEventAdd
+      # handle SIGINT from user
+      @sigIntEventAdded = !@sigIntEventAdd = false
+      process.on 'SIGINT', @commandSigInt
+
+    dreddInstance.run (error, stats) =>
+      if error
+        if error.message
+          console.error error.message
+        if error.stack
+          console.error error.stack
+        return @_processExit(1)
+      if (stats.failures + stats.errors) > 0
+        @_processExit(1)
+      else
+        @_processExit(0)
+      return
+
+    return @
+
+exports = module.exports = DreddCommand
--- a/src/configure-reporters.coffee
+++ b/src/configure-reporters.coffee
@@ -44,7 +44,7 @@ configureReporters = (config, stats, tests) ->
       when 'markdown'
         mdReporter = new MarkdownReporter(emitter, stats, tests, path, config.options.details)
       when 'apiary'
-        apiaryReporter = new ApiaryReporter(emitter, stats, tests)
+        apiaryReporter = new ApiaryReporter(emitter, stats, tests, config)
       # else
       #   Cannot happen, due to 'intersection' usage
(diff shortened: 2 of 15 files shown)


src/configure-reporters.coffee
```diff
@@ -57,14 +57,16 @@ configureReporters = (config, stats, tests) ->
 
   if usedFileReporters.length > 0
     usedFileReportersLength = usedFileReporters.length
-    if reporters.indexOf('apiary') != -1
+    if reporters.indexOf('apiary') isnt -1
       usedFileReportersLength = usedFileReportersLength - 1
-      if process.env['DREDD_REST_TOKEN'] == undefined or process.env['DREDD_REST_SUITE'] == undefined
```
netmilk: I thought this part will be moved and handled in apiary-reporter class

src/dredd.coffee
```diff
@@ -19,22 +19,33 @@ configureReporters = require './configure-reporters'
...
+removeDuplicates = (arr) ->
+  arr.reduce (alreadyProcessed, currentItem) ->
+    if alreadyProcessed.indexOf(currentItem) is -1
+      return alreadyProcessed.concat currentItem
+    return alreadyProcessed
+  , []
+
 class Dredd
-  constructor: (configOrigin) ->
+  constructor: (configOrigin, customConfig) ->
```
netmilk: Why changing/extending the interface here? I thought custom configuration will be maintained under `configOrigin['options']['custom']` and it can handle an object when using JS API directly and it will contain an array of encoded key-value pairs if initiated via CLI `--custom` interface.

src/dredd.coffee
```diff
@@ -19,22 +19,33 @@ configureReporters = require './configure-reporters'
 
 CONNECTION_ERRORS = ['ECONNRESET', 'ENOTFOUND', 'ESOCKETTIMEDOUT', 'ETIMEDOUT', 'ECONNREFUSED', 'EHOSTUNREACH', 'EPIPE']
 
+removeDuplicates = (arr) ->
```
netmilk: :thumbsup:

src/configure-reporters.coffee
netmilk: And tested there :)

src/dredd.coffee
netmilk: And this logic will be handled and tested under `apply-configuration`

src/command.coffee
```diff
@@ -0,0 +1,165 @@
...
+    # push first argument (without some known configuration --key) into paths
+    configuration.options.path ?= []
+    configuration.options.path.push @argv._[0]
+
+    configuration.options.custom ?= {}
+    configuration.options.custom['env'] = @custom['env'] or process.env
+    configuration.options.custom['stdout'] = @custom['stdout'] or process.stdout
+    configuration.options.custom['cwd'] = @custom['cwd'] or process.cwd()
+
+    custom = configuration.options.custom
```
netmilk: This does not make sense for me. Why it can't custom config passed in the configuration object? Isn't it extension of the interface only for purpose of testing?

src/reporters/apiary-reporter.coffee
```diff
@@ -18,35 +19,41 @@ class ApiaryReporter
...
-      apiUrl: process.env['APIARY_API_URL'] || "https://api.apiary.io"
-      apiToken: process.env['APIARY_API_KEY'] || null
-      apiSuite: process.env['APIARY_API_NAME'] || 'public'
+      apiUrl: @env('APIARY_API_URL') || "https://api.apiary.io"
+      apiToken: @env('APIARY_API_KEY') || null
+      apiSuite: @env('APIARY_API_NAME') || 'public'
 
     logger.info 'Using apiary reporter.'
 
+  env: (property) ->
```
netmilk: I prefer to rename it to something like `customOrEnv`

src/reporters/apiary-reporter.coffee
```diff
@@ -18,35 +19,41 @@ class ApiaryReporter
...
+    if not property?
+      return Object.keys(@config?.options?.custom?.env or process.env)
```
netmilk: Does this mean that JS API of Dredd must be initiated with the config for custom configuration something like:
```
config = 
  options:
    custom:
      env: 
        'APIARY_API_KEY': "secretkey3456789"
```

I don't like that encapsulation, under `env` key.

src/reporters/apiary-reporter.coffee
```diff
@@ -18,35 +19,41 @@ class ApiaryReporter
...
+    return returnedValue
+
   configureEmitter: (emitter) =>
     emitter.on 'start', (blueprintsData, callback) =>
       @uuid = uuid.v4()
       @startedAt = Math.round(new Date().getTime() / 1000)
 
-      ciVars = [/^TRAVIS/, /^CIRCLE/, /^CI/, /^DRONE/]
-      envVarNames = Object.keys process.env
+      ciVars = /^(TRAVIS|CIRCLE|CI|DRONE)/
```
netmilk: Not needed refactoring imho.

src/reporters/dot-reporter.coffee
```diff
@@ -2,10 +2,11 @@ logger = require './../logger'
 prettifyResponse = require './../prettify-response'
 
 class DotReporter
-  constructor: (emitter, stats, tests, details) ->
+  constructor: (emitter, stats, tests, details, config) ->
     @type = "dot"
     @stats = stats
     @tests = tests
+    @stdout = config?.options?.custom?.stdout or process.stdout
```
netmilk: I'm sorry, but I don't get this change. For what is it good for?

src/reporters/nyan-reporter.coffee
```diff
@@ -4,14 +4,15 @@ logger = require './../logger'
 prettifyResponse = require './../prettify-response'
 
 class NyanCatReporter
-  constructor: (emitter, stats, tests) ->
+  constructor: (emitter, stats, tests, config) ->
```
netmilk: ditto

src/transaction-runner.coffee
```diff
@@ -34,7 +34,7 @@ class TransactionRunner
     async.mapSeries transactions, @configureTransaction, (err, results) ->
       transactions = results
 
-    addHooks {}, transactions, @configuration.emitter
+    addHooks {}, transactions, @configuration.emitter, @configuration.options.custom
```
netmilk: :+1:

test/fixtures/multifile/multifile_hooks.coffee
```diff
@@ -1,5 +1,7 @@
 {after} = require 'hooks'
 
+console = require 'console'
```
netmilk: I don't get this. Isn't `console` in global scope, yet?

test/fixtures/test2_all.js
```diff
@@ -1,6 +1,7 @@
-var hooks;
+var hooks, console;
```
netmilk: ditto

test/fixtures/test2_events.js
```diff
@@ -1,4 +1,5 @@
 var hooks = require('hooks');
+var console = require('console');
```
netmilk: ditto

test/fixtures/test2_hooks.js
```diff
@@ -1,7 +1,8 @@
 // Generated by CoffeeScript 1.6.3
-var hooks;
+var hooks, console;
```
netmilk: ditto

test/fixtures/test_all.coffee
```diff
@@ -1,5 +1,7 @@
 {after,afterAll} = require 'hooks'
 
+console = require 'console'
```
netmilk: ditto

test/unit/command-test.coffee
```diff
@@ -0,0 +1,770 @@
...
+    custom: custom
+  }, (code) ->
+    if not finished
+      finished = true
+      exitStatus = (code ? 0)
+      cb null, stdout, stderr, (code ? 0)
+  ).warmUp().takeoff()
+  return
+
+describe "DreddCommand class", () ->
```
netmilk: Theese below are not unit tests for logic handled in `DreddCommand` class.

netmilk:
I'm sorry, but:
- I can't find out any clear purpose of this pull request. It's not explained anywhere and it should be evident.
- I assume it's a combination of:
  1. refactoring of layer handling CLI
  2. extension of Dredd class JS API to support custom configuration data e.g. for reporters

My 0,02$:
- Please **never** refactor and introduce new functionality in one PR
- It's fine that junk form `bin/dredd` was refactored and sliced into separate, more readable, JS testable layer under `DreddCommand` class
- It's **not** fine that `test/unit/command-test.coffee` does not test functionality in `DreddCommand`, it's only a copy of CLI integration test which is end-to-end testing whole Dredd functionality through CLI interface. Unit tests for `DreddCommand` should test functionality and units in `DreddComand` class, such is:

  - Handling expanded globs from shell and filling `option['path']`
  - Showing help
  - Showing version
  - Error handling on missing arguments for blueprint and server
  - Running Dredd
  - Error handling for `Dredd#run`

- It's** not** fine extending surface of `DreddCommand` class interface only for purpose of testing, it should be through mocking output from `process.cwd` etc..

kuba-kubula:
@netmilk Understood most of the problems about "why is there console, why is there the `config.options.custom.env`.

Unfortunately, you can't simply mock `process` or even `process.stdout`. Those two can't be even cloned. If you try to mock these two, you are facing problems deep in your dependencies. Really.
There are some articles about it, programers suggest to actually pass `console` or `stdout` into the script, and work on top of that. That is the reason I am passing `stdout` or `console` separately.

Other option would be to use `config.custom` instead of taking it from the really deep `config.options.custom`.

Why I chose `custom.env`? Because there are other possible things you actually want to wrap. One is env, the other one would be a `stdout`, then maybe anything else like `process.exit` function? The list gets longer really soon.

netmilk:
Ok, thanks for the explanation. In my opinion it's not needed at all, because if you want to test behavior of  `DreddCommand` class, you can spy on `console.log` and use `proxyquire` with stubs and spies for loading `dredd-command` module. Get rid with it please and write proper unit tests.

src/transaction-runner.coffee
```diff
@@ -16,9 +16,6 @@ packageConfig = require './../package.json'
 logger = require './logger'
 
 
-String::startsWith = (str) ->
```
netmilk: :thumbsup:

src/reporters/apiary-reporter.coffee
```diff
@@ -15,35 +15,51 @@ class ApiaryReporter
...
+      apiUrl: @_get 'apiaryApiUrl', 'APIARY_API_URL', 'https://api.apiary.io'
+      apiToken: @_get 'apiaryApiKey', 'APIARY_API_KEY', null
+      apiSuite: @_get 'apiaryApiName', 'APIARY_API_NAME', null
 
     logger.info 'Using apiary reporter.'
+    if not @configuration.apiToken? and not @configuration.apiSuite?
+      logger.warn "Apiary reporter environment variable APIARY_API_KEY or APIARY_API_NAME not defined."
+    @configuration.apiSuite ?= 'public'
+
+  _get: (customProperty, envProperty, defaultVal) ->
```
netmilk: Get what? I prefer to name methods/functions based on what they are doing.

PR #156, opened 2015-02-26
Feature: Custom user assertions in hooks to fail a transaction

netmilk:
- ability to fail a transaction in a hook
- get rid with `advisable` in hook's infrastructure, it does not work with `sinon`
- support for writing custom assertions with `chai`


test/unit/transaction-runner-test.coffee
```diff
@@ -538,3 +530,486 @@ describe 'TransactionRunner', ()->
...
+              loggerStub.info "second"
+              cb()
+          ]
+
+      afterEach () ->
+        loggerStub.info.restore()
+
+      it 'should run all hooks', (done) ->
+        runner.executeAllTransactions [transaction], hooksStub, () ->
+          assert.ok loggerStub.info.calledWith "first"
```
kuba-kubula: you are not checking order of calls.
I'd vote to use `loggerStub.info.firstCall.calledWith 'first'`or similar.

PR #172, opened 2015-04-12
Sphinx/RTD kickoff

Almad:


netmilk:
Used `mkdoc` instead.

PR #173, opened 2015-04-12
Improve dx

netmilk:
- add `dredd init` command with interactive configuration
- saving and loading configuration from `dredd.yml`
- added `--server` argument for running backend server before test


src/config-utils.coffee
```diff
@@ -0,0 +1,44 @@
...
+configUtils = {}
+
+configUtils.save = (args, path) ->
+  path = './dredd.yml' unless path?
+
+  args['blueprint'] = args['_'][0]
+  args['endpoint'] = args['_'][1]
+
+  for key, value of args
+    delete args[key] if key.length == 1
```
kuba-kubula: I'd vote to use `options.coffee` file as source for _aliases_ here. It might resolve any possible problems with `argv`

Also - mutating args is probably not the best idea.

src/config-utils.coffee
```diff
@@ -0,0 +1,44 @@
...
+  delete args['_']
+
+  yamlArgs = yaml.dump args
+  fs.writeFileSync path, yamlArgs
+
+
+configUtils.load = (path) ->
+  path = './dredd.yml' unless path?
+
+  yamlData = fs.readFileSync path
```
kuba-kubula: It'd be good to check for presence of the file first via `fs.existsSync`

kuba-kubula: Ok, please ignore... I see it down in the code now

src/dredd-command.coffee
```diff
@@ -60,38 +77,124 @@ class DreddCommand
...
+    return if @finished
+
+    @loadDreddFile()
+    return if @finished
+
+    @checkRequiredArgs()
+    return if @finished
+
+    @moveBlueprintArgToPath()
+    return if @finished
```
kuba-kubula: these pieces probably can be DRYied out.

PR #176, opened 2015-04-15
Use Read the Docs

smizell:
This is the initial commit for using read the docs. 

http://dredd.readthedocs.org/en/latest/

Once this looks good, I'll work on cleaning them up and making it look correct on RTD.


smizell:
@netmilk Here we go. This is kind of the first run in my opinion. It would be good to have @DeeStrong read over this and make sure it looks good. We can do all that editing in another PR or do it here.

netmilk:
@smizell , would you mind adding brief manual how to install & run MKdocs locally?

smizell:
@netmilk Great idea. I will put it in the README.

smizell:
@netmilk This has done with commits squashed.

smizell:
BTW, I simply linked to the installation instructions for MkDocs so you don't have to keep the instructions in sync.

PR #178, opened 2015-04-22
Use pitboss-ng npm instead of pitboss fork-git-url

kuba-kubula:
- brings in `pitboss-ng` with support for `nodejs 0.12` / `0.10` / `iojs`


kuba-kubula:
@netmilk thanks

PR #199, opened 2015-05-25
Completed support for API Blueprint 1A8

netmilk:
+ updated gavel fixing wrong header mismatching values error message


src/blueprint-ast-to-runtime.coffee
```diff
@@ -57,7 +58,12 @@ blueprintAstToRuntime = (blueprintAst, filename) ->
           }
 
         # expand URI parameters
-        uriResult = expandUriTemplateWithParameters resource['uriTemplate'], parameters
+        if action.attributes?.uriTemplate? and action.attributes?.uriTemplate != ''
```
kuba-kubula: this could be easily written as `if action.attributes.uriTemplate`, or am I missing any specialities here?
Probably `"0"`?

netmilk: There is empty string `''` in the blueprint when it's not present in the code.

kuba-kubula: Empty string equals to false when used in if conditions.

PR #218, opened 2015-06-10
Public:true/false Apiary reporter, based on token and api-name (suite) existence

kuba-kubula:


kuba-kubula:
@netmilk please read and review and possibly merge

PR #219, opened 2015-06-10
Connection errors in gavel error interface

netmilk:


src/transaction-runner.coffee
```diff
@@ -420,6 +420,8 @@ class TransactionRunner
 
               test.message = message
               test.results = result
+              test.results['general'] = []
```
kuba-kubula: @netmilk `results` or `result`? __blocking__

netmilk: `results`

netmilk: `test.results = result`

src/reporters/apiary-reporter.coffee
```diff
@@ -132,10 +132,16 @@ class ApiaryReporter
       data = @_transformTestToReporter test
       data.result = 'error'
 
+      data['resultData']['result']['general'] ?= []
+
       if CONNECTION_ERRORS.indexOf(error.code) > -1
-        data.resultData.errors.push "Error connecting to server under test!"
+        data['resultData']['result']['general'].push {
```
kuba-kubula: Maybe a comment about "using similar interface as Gavel to keep better interoperability" would be nice.

PR #228, opened 2015-06-23
Updated the MkDocs config from the deprecated format

d0ugal:


netmilk:
@d0ugal Thanks a lot! 🍭

PR #222, opened 2015-06-16
Report skipped and programatically failed transactions

netmilk:

PR #229, opened 2015-06-23
Clean beforeResults for every gavelResult iteration

kuba-kubula:
follow up to #222

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -290,9 +290,7 @@ class TransactionRunner
           transaction['results'] ?= {}
           transaction['results']['general'] ?= []
           transaction['results']['general']['results'] ?= []
-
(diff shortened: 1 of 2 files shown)


kuba-kubula:
@netmilk just one small, but important fix regarding `beforeResults`

src/transaction-runner.coffee
```diff
@@ -506,16 +504,17 @@ class TransactionRunner
...
+            for entityResult in data['results'] or []
               message += resultKey + ": " + entityResult['message'] + "\n"
 
         test['message'] = message
 
         transaction['results'] ?= {}
 
-        for key, value of gavelResult
+        for own key, value of gavelResult or {}
+          beforeResults = null
```
netmilk: 🙆

PR #241, opened 2015-07-08
Use vector badges in the README

kylef:
Mostly to make it look better when using a retina device.

Before:

screen shot 2015-07-08 at 09 03 29

After:

screen shot 2015-07-08 at 09 05 44


netmilk:
@kylef, unbelievable, how can you know I today switched to a new MBP?

w-vi:
Looks legit, merging

PR #262, opened 2015-08-20
Adding pretty JSON

obihann:
- addition of [prettyjson](https://github.com/rafeca/prettyjson) to `src/prettyify-response.coffee` so that body and schema objects that are of content type `application/json' are easier to read
- update to how we check content type, since types such as `application/json; charset=utf-8` are valid we should not do an exact match on `application/json` and instead do an indexOf
- allowing `schemaBody` to be passed through prettyjson along with `schema`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "dependencies": {
     "advisable": "~0.2.0",
     "async": "^0.9.x",
+    "blueprint-transactions": "^0.0.2",
     "chai": "^2.1.0",
     "clone": "^1.0.0",
     "coffee-script": "^1.9.1",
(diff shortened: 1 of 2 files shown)


honzajavorek:
@obihann I'm sorry we did not respond for a longer time :cold_sweat: Thanks for your work! :+1: As `master` is fixed now, would you mind to rebase with it and try whether tests pass?

honzajavorek:
And since you're already touching the content type logic in your PR, I'd prefer if it would be able to deal also with content types such as `application/hal+json`, `application/vnd.geo+json` and similar ones. What do you think?

obihann:
@honzajavorek sure, I'll rebase this and see how it goes. I'll also try to add those content types in.

honzajavorek:
@obihann Awesome! If I may provide hints, I'd use [media-typer](https://www.npmjs.com/package/media-typer) for the heavy lifting.

netmilk:
Looking forward @obihann! Thanks a lot for all your contributions!

honzajavorek:
I think we should incorporate this as part of #765.

honzajavorek:
I'm very sorry we were not able to get back to this. I need to close this as we're going to work on #705 first, which requires having no pending Pull Requests. We'll get back to this proposal as soon as we're revamping the CLI reporter: #765

Thanks for all the work! I'll do my best for this contribution not to be forgotten!

revett:
@honzajavorek Have you come back to this feature yet?

honzajavorek:
@revett We thought about revamping the whole CLI reporter: #765

revett:
@honzajavorek Ah okay 👍That seems like quite a large piece of work but this issue is a simple fix. If I put in a PR for this, would it be merged before #765?

honzajavorek:
Yes, that would be nice!

PR #275, opened 2015-09-10
Apiary reporter not throwing anymore, only passing error to callback

miiila:
Introducing _performRequestAsync to return callback filled with an error instead of fatal error in Apiary reporter.

Introduced change is triggered by need of assertion failure just after reporter start.

### DIFF
--- a/src/dredd.coffee
+++ b/src/dredd.coffee
@@ -186,13 +186,25 @@ class Dredd
         return callback(runtimeError, stats) if runtimeError
         cb()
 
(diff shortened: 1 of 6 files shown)


w-vi:
@netmilk ... possibly squash to single commit? Or two commits one from miiila and one from you?

netmilk:
@wvi, squashed. 

Seal of approval?

w-vi:
@netmilk approved!

PR #285, opened 2015-10-01
replacing drafter.js with protagonist

obihann:


obihann:
Not complete yet, just starting the pull request so its ready.

obihann:
@w-vi Just to make sure I'm not going in a wrong direction, the "do not merge" label is only here because of my comment and the failing tests, correct? Or am I just handling this the wrong way?

w-vi:
@obihann It's WIP right? So this is here to make sure that nobody tries to merge it until you say it's done, than I will remove the `DO NOT MERGE` label.

w-vi:
@obihann Also another thing is that dredd now goes through big overhaul internally, so if you are going to make bigger changes it will probably need rebase ... @netmilk could you please comment on this.

obihann:
@w-vi makes sense thanks.

obihann:
Closing this because after hearing from @netmilk and @pksunkara this is not the solution I need for my problems, and is much more complicated than I thought.

Issue #283, opened 2015-09-29
Update dredd to support nullable attribute


obihann:
Dredd should properly test against the new `nullable` attribute supported by MSON and drafter.

md
`+ s1 (string, nullable)`

json
`{"s1":null}`

I can help with this however I am not 100% sure if I need to make changes to drafter.js or protagonist?

Thanks.

netmilk:
Hi @obihann,

you're right, this a Drafter issue. Drafter compiles JSON schema from MSON and this JSON schema is used for Gavel validation of server response. Since this feature will be introduced in Drafter it's only a matter of bumping Drafter version in Dredd to support it.

obihann:
Awesome! @pksunkara when can can we expect a new release of drafter? I know nullable is in the master branch since I was so involved in that, but it seems that we need a new tag release so that Dredd can point specifically to that.

pksunkara:
Unfortunately, we can't announce nullable support until https://github.com/apiaryio/drafter/issues/121 is done. But other than that, we have a bit of work currently happening after which there will be a new drafter release.

obihann:
@pksunkara thanks, I'll try to take a look at that tonight and help you guys out a bit.

@netmilk I just want to verify all that is holding this up is in fact [apiaryio/drafter](https://github.com/apiaryio/drafter), I ask this because I understood dredd uses [apiaryio/drafter.js](https://github.com/apiaryio/drafter.js) (which is being replaced by [apiaryio/protagonist](https://github.com/apiaryio/protagonist))?

obihann:
I think I have my confusion figured out @netmilk... Dredd uses drafter.js which is a wrapper (?) for protagonist which uses drafter via a git module. Right?

pksunkara:
Dredd uses drafter.js which uses protagonist 0.x. We need to move dredd to protagonist 1.x.

obihann:
That makes perfect sense, thanks. I'm kind of afraid to ask but what is waiting on moving dredd to protagonist 1.x?

Thanks.

pksunkara:
I think @netmilk will be more suited to answer this.

obihann:
Sounds good, thanks.

obihann:
@netmilk I'm running into some difficulty right now testing an update to dredd so that it will use drafter.js and the latest (currently master branch) version of drafter. I have a feeling this may be because (as @pksunkara) dredd needs to use protagonist 1.x directly and eliminate the reference to drafter.js.

[...]

PR #286, opened 2015-10-07
WIP: Use Protagonist instead of Drafter.js

tobiashm:
Drafter.js is now deprecated, and has been replaced by Protagonist.

This commit is just a naive replacement of `new Drafter().make` with `protagonist.parse` for parsing the API Blueprint documents.
A lot of tests are currently failing, so probably the output of the two methods are not compatible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,10 +19,10 @@
   "dependencies": {
     "advisable": "~0.2.0",
     "async": "^0.9.x",
+    "blueprint-transactions": "^0.0.2",
     "chai": "^2.1.0",
     "clone": "^1.0.0",
     "coffee-script": "^1.9.1",
-    "drafter": "~0.2.5",
     "file": "~0.2.2",
     "gavel": "0.5.2",
     "glob": "^5.0.3",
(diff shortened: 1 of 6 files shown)


obihann:
I started a similar PR #285 but after talking to @netmilk and @pksunkara I learnt that their is a big difference between drafter.js and protagonist, specifically, drafter does not render json the same as drafter.js does. Check out #283 and you can see the full conversation. I think this may have to wait until drafter us updated to match the same functionality as drafter.js, @pksunkara should be able to speak to a timeline or what is needed for that, but I think that is our first step before we do this.

tobiashm:
Okay. Didn't look at closed PR before starting this.
I'll look more into the differences, and where they're going with Dredd.
I still think this needs to happen; Dredd will currently not build with Node 4, due to the version of protagonist that Drafter.js depends on.
Hopefully some of the apiary.io folks will help out :-)

obihann:
I'm stuck waiting for it as well but more because of the new `nullable` attribute that will be added. As I understand it drafter has to be changed to include JSON schema rendering, after that they can release 1.1.0 of drafter, then update protagonist to that version, and lastly dredd.

They have setup a [1.1.0 milestone](https://github.com/apiaryio/drafter/milestones/1.1.0) for drafter however I'm not sure if that list of issues includes the JSON schema rendering. 

Hopefully that helps :+1:

kylef:
Now that Protagonist was updated with JSON Schema generation this is now possible, and it's made it into master of Dredd via #309.

Thanks very much for your effort here @obihann and @tobiashm and nice to see this has made it in :balloon:.

tobiashm:
@kylef good news! Thanks

PR #288, opened 2015-10-16
Add info about supported Node.js versions

w-vi:


o5:
`Support for node.js 4.x.x is on the way`

When? :)

netmilk:
Hi @o5,

the upgrade is waiting for full MSON support and its JSON schema representations rendering in the `protagonist` dependency. Once it will be available it will be upgraded in Dredd as soon as possible, added build for Node 4 and released.

PR #306, opened 2015-12-10
Fixed typo

netmilk:


src/interactive-config.coffee
```diff
@@ -72,7 +72,7 @@ interactiveConfig.prompt = (config = {}, callback) ->
           (answers['apiary'] == true && ! config['custom']?['apiaryApiName']?) &&
           (answers['apiary'] == true && answers['apiaryApiKey'] != '')
         )
-      else
+      elsebe
```
w-vi: Whats this doing? :)

netmilk: Lol

honzajavorek:
@netmilk I think it is worth squashing these two commits, but YMMV

kylef:
Huge :+1: on squashing these commits, it's a good practise to have since otherwise [git-bisect](https://git-scm.com/docs/git-bisect) is difficult to use.

honzajavorek:
Squashed the commits for @netmilk, rebased with `master` & merging. :balloon:

PR #307, opened 2015-12-14
Fixing the example in Ruby Hooks docs

netmilk:


honzajavorek:
Is this some special Markdown syntax supported by Sphinx? Never seen this before in Markdown.

w-vi:
It's an example block triple back tick watch it closely

kylef:
@honzajavorek This is in a code-block, it's valid Ruby change. :+1: on merging, this was incorrect before.

honzajavorek:
Ah! I get it (facepalm) I thought it's something different than it is. Thanks for clarification.

PR #310, opened 2015-12-16
Test with node 4 and 5

ascripcaru:


honzajavorek:
This is awesome! Can't wait! :gift:

.travis.yml
```diff
@@ -4,6 +4,17 @@ node_js:
...
+  - "4"
+  - "5"
+
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - gcc-4.8
+      - g++-4.8
```
kylef: I wonder if it would be better to instead use the trusty image on Travis (https://docs.travis-ci.com/user/trusty-ci-environment/). Since it already has modern compilers it should speed up the CI time and also removes many lines from the .travis.yml file.

What do you think?

ascripcaru: I'll give it a try

ascripcaru:
@honzajavorek now the bastard version of node that fails without a good reason is 4. if you could give it a restart? if you like the approach I'm glad that I could help

kylef:
I hit rebuilt and it's passing now, not sure what's flaky with it.

.travis.yml
```diff
@@ -1,11 +1,14 @@
+sudo: required
+dist: trusty
```
netmilk: @ascripcaru Hi, why are we locking build to some particular Ubuntu release here?

.travis.yml
```diff
@@ -1,11 +1,14 @@
+sudo: required
```
netmilk: @ascripcaru Why is sudo required?

.travis.yml
kylef: @netmilk See my comment here: https://github.com/apiaryio/dredd/pull/310#discussion_r47811905 it's basically the new image on Travis CI that has modern compilers. The default is much older.

PR #313, opened 2015-12-17
Support node 4 and 5 in readme

netmilk:


w-vi:
@netmilk  it works also on iojs 2.x.x so i would update the readme thatit supports iojs too

honzajavorek:
Sorry for hijacking the PR, but I felt like I could help.

- Addressed @w-vi's comment.
- Reworded the sentence and ordered the versions so the newest ones are mentioned first.
- Removed `~0.8.15` since we do not seem to test against it so we apparently can't guarantee Dredd will work with it.

w-vi:
@hj can you squash it? 3 commits for changing README feel a little bit too many :)

honzajavorek:
@w-vi rebased

PR #352, opened 2016-01-20
Fix syntax typo in docs example

arcz:
Spotted another typo


honzajavorek:
Awesome, thanks! :gem:

PR #359, opened 2016-02-03
Run dependent integration builds on every release

honzajavorek:
This PR introduces a script, which will trigger dependent integration builds for every hook handler implementation. Such setup allows us to test whether new version of Dredd won't break the interface for existing hook handlers.

#### What?

1. Every time commit is pushed to GitHub, Travis CI automatically starts a new build.
2. The build is defined by contents of `.travis.yml`. It runs regular tests and then runs the new script introduced in this PR, `npm run test:hook-handlers`.
3. The script...
    1. ...makes sure it is ran for just one build run within the build matrix.
    2. ...checks whether hook handler integration tests were triggered or not. **If the tested commit has tag or its commit message contains words "tests hook handlers", it continues.** Otherwise it skips the tests (ends immediately with 0 exit status code).
    3. ...creates special dependent integration branches for each hook handler implementation. In these branches, it takes whatever is in master branch of the hook handler repository with the only difference that instead of `npm i -g dredd` it links the Dredd from currently tested commit.
    4. ...pushes these dependent branches to GitHub, which triggers dependent Travis CI builds.
    5. ...polls for results of the dependent Travis CI builds.
    6. ...evaluates results of dependent builds and if any of them didn't pass, the script exits with non-zero code.
    7. ...makes sure it deletes the dependent git branches from GitHub before exiting.

#### Tested hook handler implementations

- https://github.com/apiaryio/dredd-hooks-python/
- https://github.com/apiaryio/dredd-hooks-ruby/
- https://github.com/ddelnano/dredd-hooks-php/

#### Known issues

- If `master` branch of hook handler repository becomes red, this whole build will fail and we won't be able to release new version of Dredd (not with green status). Instead, we should integrate with commit corresponding to the latest release of the hook handler implementation. As it's not trivial, I leave that up to the situation when we actually bump into such problem (if ever).
- If the main Travis CI build where the script is being ran gets canceled, the script won't cleanup the dependent branches on GitHub. It has solutions, but again, I leave that up to the situation when we actually bump into such problem (if ever).


w-vi:
@honzajavorek please document it bit more in the README or maybe better in the `test-hook-handlers.coffee` comments. What it actually does is not obvious, add the caveats as well to make sure it is know to everybody.

w-vi:
otherwise I am giving it go ... :fireworks:

honzajavorek:
@w-vi Thanks, awesome! Do you think 87c2fd4 solves the problem?

w-vi:
@honzajavorek perfect :clap:  so now it is :checkered_flag:

honzajavorek:
Awwww :smile_cat: Thanks!

Issue #321, opened 2015-12-22
Add PHP and Go to `dredd init` and support them it in the --language


netmilk:
Dredd interactive config doesn't offer Go and PHP as a language option and doesn't support them in the `--langage` switch.

honzajavorek:
This was closed by #362.

PR #362, opened 2016-02-09
Add more hook languages into dredd init

abtris:
By http://dredd.readthedocs.org/en/latest/ are PHP, Go available in Dredd but missing in `dredd init`

### DIFF
--- a/src/dredd-command.coffee
+++ b/src/dredd-command.coffee
@@ -123,6 +123,8 @@ class DreddCommand
           console.log "  $ pip install dredd_hooks"
         else if config['language'] == 'php'
           console.log "  $ composer require ddelnano/dredd-hooks-php:~1.0.0 --dev"
+        else if config['language'] == 'perl'
+          console.log "  $ cpanm Dredd::Hooks"
         else if config['language'] == 'go'
           console.log "  $ go get github.com/snikch/goodman"
 
--- a/src/dredd-command.coffee
+++ b/src/dredd-command.coffee
@@ -121,6 +121,10 @@ class DreddCommand
           console.log "  $ gem install dredd_hooks"
         else if config['language'] == 'python'
           console.log "  $ pip install dredd_hooks"
+        else if config['language'] == 'php'
+          console.log "  $ composer require ddelnano/dredd-hooks-php:~1.0.0 --dev"
+        else if config['language'] == 'go'
+          console.log "  $ go get github.com/snikch/goodman"
 
         console.log "  $ dredd"
         console.log ""
--- a/test/unit/dredd-command-test.coffee
+++ b/test/unit/dredd-command-test.coffee
@@ -254,7 +254,7 @@ describe "DreddCommand class", () ->
       it 'prints out version', ->
         assert.include stdout, "#{packageJson.name} v#{packageJson.version}"
 
-    describe '"init"', ->
+    describe '"init" (nodejs)', ->
       before (done) ->
         sinon.stub interactiveConfigStub, 'run', (argv, cb) ->
           cb({language: 'nodejs'})
@@ -272,7 +272,7 @@ describe "DreddCommand class", () ->
       it 'should save configuration', ->
         assert.isTrue configUtilsStub.save.called
 
-    describe '"init"', ->
+    describe '"init" (python)', ->
       before (done) ->
         sinon.stub interactiveConfigStub, 'run', (argv, cb) ->
           cb({language: 'python'})
@@ -291,7 +291,7 @@ describe "DreddCommand class", () ->
         assert.isTrue configUtilsStub.save.called
 
 
-    describe '"init"', ->
+    describe '"init" (php)', ->
       before (done) ->
         sinon.stub interactiveConfigStub, 'run', (argv, cb) ->
           cb({language: 'php'})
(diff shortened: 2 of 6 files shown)


honzajavorek:
Thanks! There is an issue for this: #321 It mentions we should also add it to the `--language` switch. Could you please look into that?

abtris:
ok, I look into.

abtris:
I'm not sure how add [go lang](http://dredd.readthedocs.org/en/latest/hooks-go/) into `hooks-worker-client`.

abtris:
@honzajavorek I think this looks ready. Please make review.

honzajavorek:
Sorry for delay. Could you please also add Perl :camel: from #364? Thanks! :heart_eyes:

abtris:
@honzajavorek I add perl, but I'm not sure if this works.

honzajavorek:
It should, presence of the `dredd-hooks-{language}` command is tested also by the Cucumber test suite. Also see https://github.com/ungrim97/Dredd-Hooks/blob/master/script/dredd-hooks-perl.

src/dredd-command.coffee
```diff
@@ -121,6 +121,12 @@ class DreddCommand
           console.log "  $ gem install dredd_hooks"
         else if config['language'] == 'python'
           console.log "  $ pip install dredd_hooks"
+        else if config['language'] == 'php'
+          console.log "  $ composer require ddelnano/dredd-hooks-php:~1.0.0 --dev"
+        else if config['language'] == 'perl'
+          console.log "  $ cpanm Dredd::Hooks"
+        else if config['language'] == 'go'
+          console.log "  $ go get github.com/snikch/goodman"
```
honzajavorek: Does this really work? It's not in the README of the project (there are no installation guidelines :( ).

src/dredd-command.coffee
```diff
@@ -121,6 +121,12 @@ class DreddCommand
           console.log "  $ gem install dredd_hooks"
         else if config['language'] == 'python'
           console.log "  $ pip install dredd_hooks"
+        else if config['language'] == 'php'
+          console.log "  $ composer require ddelnano/dredd-hooks-php:~1.0.0 --dev"
```
honzajavorek: Why we do specify version here? And what does the `--dev` mean?

netmilk:
@abtris @honzajavorek Guys! Beware! I suspect there's a typo in the original `doc/hooks-perl.md` documentation for the Perl hooks. It should be `cpan` no `cpanm`. 

@w-vi, would you mind confirming?

@honzajavorek If it is a typo, would you mind fixing it in the doc?

abtris:
@netmilk both exists `cpan` and `cpanm`. I check it. But I'm not sure what is right.

test/unit/hooks-worker-client-test.coffee
```diff
@@ -176,6 +176,60 @@ describe 'Hooks worker client', () ->
...
+              logs.push args.join(" ")
+
+            consoleStub.log.restore()
+            whichStub.which.restore()
+
+            assert.include logs.join(", "), "composer require ddelnano/dredd-hooks-php"
+            done()
+          , 2200
+
+
```
honzajavorek: Missing test for Perl.

src/dredd-command.coffee
abtris: Similar `devDependecies` in nodejs.

test/unit/dredd-command-test.coffee
```diff
@@ -272,6 +272,61 @@ describe "DreddCommand class", () ->
       it 'should save configuration', ->
         assert.isTrue configUtilsStub.save.called
 
+    describe '"init"', ->
```
honzajavorek: Aren't these tests redundant? I'm not sure how important the language is for this particular test. I think it tests just basic functionality of the `dredd init` command itself.

src/dredd-command.coffee
honzajavorek: Ah, thanks!

test/unit/dredd-command-test.coffee
abtris: Yes, but helping with coverage for lines with others options (php, python, perl).

honzajavorek:
I can fix `cpan` / `cpanm` in #368 if there are mistakes. Maybe @ungrim97 himself can give us a hint?

test/unit/dredd-command-test.coffee
honzajavorek: I don't know what's best practice in that case :( Having several tests just to keep coverage sounds like smell to me, but YMMV. Calling @netmilk for an opinion.

If we are keeping them, can we at least rename [this](https://github.com/apiaryio/dredd/pull/362/files#diff-9583335aa63e76771edff5e02cf4a0c2R257) to

```diff
- describe '"init"', ->
+ describe '"init" (nodejs)', ->
```

and then add the language also to the new ones?


```diff
- describe '"init"', ->
+ describe '"init" (python)', ->
```

etc.

ungrim97:
cpanm is the preferred installation manager for perl (https://metacpan.org/pod/App::cpanminus). If you are using the perl pre installed on the system rather than installing it via perlbrew/plenv then it is likely not already installed. 

cpan is the client that comes packaged with system builds of perl and should work fine (cpan Dredd::Hooks) but may require sudo access if its installing against system perl.

travis (and most perl devs) install perl using either [perlbrew](http://perlbrew.pl/) or [plenv](https://github.com/tokuhirom/plenv) and cpanm to install modules to.

honzajavorek:
@ungrim97 Thanks for clarification! :camel: I'd go with `cpanm` then the same way we rely on people being able to get `pip` on their machine, regardless what they're provided with by default.

w-vi:
`cpanm` FTW http://www.cpan.org/modules/INSTALL.html

honzajavorek:
Thanks! :balloon:

PR #363, opened 2016-02-11
Exclude documentation files and directories from the test coverage

kuba-kubula:


netmilk:
I'm not sure if this is an actual solution for the confusing message about decreasing coverage when editing Documentation. Coverage should be calculated from the annotated JS code and no Markdown files can be annotated. But it's nothing bad to exclude the documentation explicitly.

kuba-kubula:
@netmilk so is it free to merge?

honzajavorek:
@kuba-kubula Does it support something like `*.md`? Otherwise seems OK to me, but I am also unsure about how `coffeecoverage` treats this and whether it's going to help. Did you try it out?

w-vi:
It didn't change the coverage at all, you can have a look

netmilk:
I think it has no effect. Closing.

kuba-kubula:
Unfortunately Coveralls is in READ-ONLY mode due to database Upgrade event today... :(

netmilk:
@kuba-kubula Please elaborate, I'm affraid I don't understand what do you mean.

honzajavorek:
@netmilk I think he is trying to make some actual changes in README to see whether they affect coverage with the new setting. However, he's unable to evaluate the attempt as Coveralls won't report back now.

netmilk:
@kuba-kubula @honzajavorek I meant this:

> Unfortunately Coveralls is in READ-ONLY mode due to database Upgrade event today... :(

PR #370, opened 2016-02-17
What if we could link SVG directly?

honzajavorek:
Attempt to make the README image more retina-friendly.


honzajavorek:
Nope, GitHub [serves the SVG with incorrect Content-Type](http://stackoverflow.com/q/13808020/325365) and the image won't render :cry:

PR #332, opened 2015-12-28
Big refactoring and optimisation of hook handler spawner and client

netmilk:
- [x] Removed proxyquire hell, using clean, asynchronous code (#235)
- [x] Gracefully handling handler crashes and kills and lot of corner-cases (#238)
- [x] Removed console.logging using logger or error messages in hook handler spawner
- [x] Trying to terminate server first, then killing, waiting for exit. Should prevent leaking server process.
- [x] All spawn, wait and kill variables in constants for future configuration from CLI (#328, #235)
- [x] Gracefully ending the server event though Dredd experiences exception (#330)
- [x] Flaky tests workaround, hopefully. (#282)
- [x] Tested against each hook handler test suites (gets enabled by https://github.com/apiaryio/dredd/pull/359)
- [x] Removed stack tracing for non-exceptions in CLI
- [x] 2x faster test-suite

- [x] Rebased into one commit
- [x] Rebased against master

before:
```
  734 passing (4m)


real    4m20.975s
user    2m24.796s
sys     0m5.835s
```

after:
```
  820 passing (2m)


real	2m15.007s
user	0m43.701s
sys     0m5.653s
```

PR #379, opened 2016-02-24
Filling gaps in CLI integration tests & fixing coverage so it's able to pick up tests using Dredd binary

honzajavorek:
### CLI integration tests 

Filling gaps in CLI integration tests and raising coverage after merging #332. Fixed coverage so it picks up tests using `./bin/dredd`. Moved some tests from the giant `cli-test.coffee` into their separate files:

-  CLI - API Description Document
-  CLI - Reporters
-  CLI - Server Process

Further work on cleanup of CLI integration tests is to be discussed with @netmilk and addressed some day in the future.

### HTML cov removed

I had to remove the ability to generate simple HTML file with coverage, because it's hardcoded functionality of Mocha and it would be a lot of work and hassle to generate it on our own from the merged LCOV file. There is a way how to generate local HTML report, though:

```
$ brew install lcov
$ genhtml ./cov.info --output-directory ./coverage
$ open ./coverage/index.html
```

The report generated this way is not one file and it's extremely ugly, but... ¯\\\_(ツ)\_/¯

### Aftermath

- Coverage before #332: ~95.3%
- Coverage after #332: ~93%
- Coverage this PR brings: ~95.6%


netmilk:
This is amazing!

bin/dredd
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env node
+
```
netmilk: This is awesome! Would it be possible to keep it here in this file and enable the functionality by exporting some ENV variable? I would love to have one file. I'm afraid diverging of those two might happen.

bin/dredd-coverage
```diff
@@ -0,0 +1,48 @@
...
+    var runner = new EventEmitter();
+
+    // Monkey-patching 'process.stdout.write' to catch all stdout in a variable
+    var content = '';
+    var write = process.stdout.write;
+    process.stdout.write = function(stdout) {
+      content += stdout;
+    };
+
+    // Collecting the stats
```
netmilk: This is awesome! Would it be possible to have it in the original `bin/dredd` file and enable the functionality by exporting some ENV variable? I would love to have one file because I'm afraid diverging of these two might happen.

honzajavorek: Awww! :smirk_cat: Good idea. This didn't occur to me. Will do.

honzajavorek:
@netmilk Updated, now `bin/dredd` and `bin/dredd-coverage` are merged into one.

netmilk:
:+1:

netmilk:
@honzajavorek I'm not sure if Coveralls can see the same coverage as you locally: https://coveralls.io/builds/5255948

> Coverage this PR brings: ~95.6%

honzajavorek:
@netmilk That smells :disappointed: This means I had to break the coverage report somewhere (in the last commit? uh..), because in the past Coveralls were catching up correctly on this branch:

screen shot 2016-03-01 at 15 31 21

Will investigate. Thanks for spotting this!

honzajavorek:
@netmilk addressed in https://github.com/apiaryio/dredd/pull/392

PR #382, opened 2016-02-26
Pass in options to hook worker

xvalentino:


netmilk:
@xvalentino This is awesome! Thank you very much for the contribution! I'll have a look and I might have some comments. :)

src/add-hooks.coffee
```diff
@@ -130,7 +130,8 @@ addHooks = (runner, transactions, callback) ->
       # Worker client will start the worker server and pass the "hookfiles" options as CLI arguments to it
       else
         # start hooks worker
-        hooksWorkerClient = new HooksWorkerClient(runner)
+        hookTimeout = runner?.configuration?.options?['hook-timeout'] || 5000
+        hooksWorkerClient = new HooksWorkerClient(runner, hookTimeout)
```
netmilk: The `configuration` object is already passed  to the `HooksWorkerClient` on the `runner` object. I think there's no need to add ab additional argument.

src/hooks-worker-client.coffee
```diff
@@ -8,9 +8,6 @@ generateUuid = require('node-uuid').v4
 logger = require './logger'
 which = require './which'
 
-
-HOOK_TIMEOUT = 5000
```
netmilk: I think we can get rid of these constants at all and move the defaults to the constructor and override them if the configuration is present.

netmilk: The reason is I'd like to make almost all of them configurable.

src/options.coffee
```diff
@@ -131,4 +131,8 @@ options =
   version:
     description: "Show version number.\n"
 
+  'hook-timeout':
```
netmilk: I'm not sure if the `hook-timeout` configuration parameter isn't confusing because it applies to the `HookWorkerClient` only and it doesn't apply for native Node.js hooks.

I suggest:
a) to make it work for the native Node.js hooks as well
b) rename it  ti something` like: `hook-worker-timeout`

Option a) makes more sense in my opinion

src/hooks-worker-client.coffee
```diff
@@ -24,7 +21,7 @@ HANDLER_MESSAGE_DELIMITER = "\n"
 
 
 class HooksWorkerClient
-  constructor: (@runner) ->
+  constructor: (@runner, @timeout) ->
```
netmilk: Mentioned here: https://github.com/apiaryio/dredd/pull/382/files#r54400093

test/unit/hooks-worker-client-test.coffee
```diff
@@ -28,7 +28,7 @@ TransactionRunner = require '../../src/transaction-runner'
 hooksWorkerClient = null
 
 loadWorkerClient = (callback) ->
-  hooksWorkerClient = new HooksWorkerClient(runner)
+  hooksWorkerClient = new HooksWorkerClient(runner, 5000)
```
netmilk: Mentioned here: https://github.com/apiaryio/dredd/pull/382/files#r54400093

src/hooks-worker-client.coffee
```diff
@@ -9,23 +9,19 @@ logger = require './logger'
...
-
-HANDLER_HOST = 'localhost'
-HANDLER_PORT = 61321
-HANDLER_MESSAGE_DELIMITER = "\n"
-
-
 class HooksWorkerClient
   constructor: (@runner) ->
-    @language = @runner.hooks?.configuration?.options?.language
+    options = @runner.hooks.configuration.options
```
xvalentino: The tests use @runner.hooks.configuration, is there a reason not to use @runner.configuration?

romansklenar:
:+1:

src/hooks-worker-client.coffee
netmilk: I'm so sorry for the delayed response. I'd use `@runnner.configuration`.

I'd be up for removing `hooks.configuration` to prevent data duplication and having a single source of truth in the future.

netmilk: But it's not worth redoing it for now. :)

src/options.coffee
```diff
@@ -131,4 +131,32 @@ options =
   version:
     description: "Show version number.\n"
 
+  'hook-worker-timeout':
```
netmilk: This is amazing. Would you mind add the description to all of them and updating the documentation of the CLI and JS interface.

src/hooks-worker-client.coffee
```diff
@@ -9,23 +9,19 @@ logger = require './logger'
...
+    @language = options.language
+    @timeout = options['hook-worker-timeout'] || 5000
+    @connect_timeout = options['hook-worker-connect-timeout'] || 1500
+    @connect_retry = options['hook-worker-connect-retry'] || 500
+    @after_connect_wait = options['hook-worker-after-connect-wait'] || 100
+    @term_timeout = options['hook-worker-term-timeout'] || 5000
+    @term_retry = options['hook-worker-term-retry'] || 500
+    @handler_host = options['hook-worker-handler-host'] || 'localhost'
+    @handler_port = options['hook-worker-handler-port'] || 61321
+    @handler_message_delimiter = options['hook-worker-handler-message-delimiter'] || '\n'
```
netmilk: I'm sorry, haven't noticed this one originally. Please exclude this from the options. The message delimiter should be private property. It's not anything that should change.

netmilk: Just remove the options overriding, please. The configuration can totally be in the constructor.

netmilk:
@xvalentino Thank you for addressing my comments! It's perfect. I think it's good to merge if we update the documentation on CLI and JS interface and remove the possibility to configure the message delimiter.

We might also consider mentioning those configuration options in the documentation of the [TCP language bridge](https://github.com/apiaryio/dredd/blob/master/docs/hooks-new-language.md). It can be helpful for debugging when implementing a new language.

src/hooks-worker-client.coffee
xvalentino: Sounds good to me!

src/options.coffee
xvalentino: Cool, I'm not totally sure what a majority of the variables mean, I can look into it later and ask for specific ones. If there are any other names as well that'd make more sense just let me know!

src/hooks-worker-client.coffee
xvalentino: Will do! Thanks!

src/hooks-worker-client.coffee
honzajavorek: @netmilk I tried to remove `runner.hooks.configuration`, but it seems it's there on purpose and it's part of the Node.js API: http://dredd.readthedocs.org/en/latest/hooks-nodejs/#api-reference

PR #385, opened 2016-02-29
Link to hooks template explaining the weird ENV variable

netmilk:


honzajavorek:
@netmilk Thanks very much! During #379 I bumped exactly into this - complete magic. I think the comment clarifies the situation a lot.

The linter doesn't like trailing whitespace :stuck_out_tongue_winking_eye: Fixed it for you :heart: Will merge once it passes.

src/hooks-worker-client.coffee
```diff
@@ -301,7 +301,8 @@ class HooksWorkerClient
 
     @runner.hooks.afterAll (transactions, hookCallback) =>
 
-      # This is needed for transaction modification integration tests.
+      # This is needed for transaction modification integration tests:
+      # https://github.com/apiaryio/dredd-hooks-template/blob/master/features/execution_order.feature
```
kylef: master? should probably tie this link to a specific tag or commit because it's a moving target. This could be a 404 in the future

honzajavorek: This won't easily move as all the hooks handler implementations use the `dredd-hooks-template` repo as a base, so we don't change it much (actually, I'd even say we don't change it at all). I'd use commit if there would be a link to specific line, but this way I'd say it's as stable as any other internet link.

PR #399, opened 2016-03-08
Pinned Protagonist to 1.2.6.

### LINKED ISSUE #394: Dredd should have completely strict dependency on Drafter aka Protagonist
Currently, every minor version of Protagonist brings various regressions etc., but the package.json as it is now allows users to install the latest version. Any new install of Dredd evaluates the Protagonist dependency to it's latest version. Thus I think we should pin it to be sure about what users have installed and to have control over what combination we consider in Dredd as stable. Once Drafter stabilizes, we can make this more loose again.

@kylef @netmilk

honzajavorek:
Fixes #394.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "node-uuid": "~1.4.2",
     "optimist": "~0.6.1",
     "pitboss-ng": "^0.3.1",
-    "protagonist": "^1.2.5",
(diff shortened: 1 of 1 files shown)


honzajavorek:
We agreed with @netmilk won't use latest Protagonist, but the last one which actually passes our regression tests.

PR #380, opened 2016-02-25
Regression tests for #354

honzajavorek:
This Pull Request introduces tests for #354. Until these tests pass with some version of Protagonist, https://github.com/apiaryio/protagonist/issues/133 can't be considered as resolved.

I created Data Structure, which uses the `Include` statement. I test several combinations of where the Data Structure is defined and from where it is referenced. Following table displays which combinations properly generate JSON and JSON Schema assets when using Protagonist `v1.3.0-pre.0`:

| :open_mouth: | Defined in Resource | Defined as Data Structure |
|---|---|---|
| **Referenced from Payload** | JSON Schema: :heavy_multiplication_x: JSON: :heavy_multiplication_x: | JSON Schema: :heavy_check_mark: JSON: :heavy_check_mark: |
| **Referenced from Action**  | JSON Schema: :heavy_multiplication_x: JSON: :heavy_multiplication_x: | JSON Schema: :heavy_check_mark: JSON: :heavy_check_mark: |

See also https://github.com/apiaryio/dredd/issues/354#issuecomment-188705444 for further details.

-----------

**Update:** Per request in https://github.com/apiaryio/drafter/issues/248#issuecomment-194731665, there is now minimal app reproducing the bug:

### https://github.com/apiaryio/protagonist-include-issue

Issue #381, opened 2016-02-25
Dredd fails after "ExtendElement must be composed from Elements of same type" warning


amuino:
When parsing our blueprint, which uses `Include` in MSON Data Structures, we get the following output:

```
$ dredd
Configuration dredd.yml found, ignoring other arguments.
Starting server with command: false
Waiting 0 seconds for server command to start...
warn: Parser warning in file 'doc/api/public.apib.md': (4) ExtendElement must be composed from Elements of same type
/Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/node_modules/blueprint-transactions/lib/blueprint-ast-to-runtime.js:25
  if (blueprintAst['name'] !== "") {
                  ^

TypeError: Cannot read property 'name' of undefined
  at blueprintAstToRuntime (/Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/node_modules/blueprint-transactions/lib/blueprint-ast-to-runtime.js:25:19)
  at Object.module.exports.compile (/Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/node_modules/blueprint-transactions/lib/blueprint-transactions.js:11:14)
  at Dredd.compileTransactions (/Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/lib/dredd.js:285:44)
  at /Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/lib/dredd.js:117:26
  at /Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/lib/dredd.js:272:16
  at done (/Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/node_modules/async/lib/async.js:132:19)
  at /Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/node_modules/async/lib/async.js:32:16
  at /Users/amuino/.nvm/versions/node/v5.7.0/lib/node_modules/dredd/lib/dredd.js:245:18
```

Initial debugging points to either `drafter` or `protagonist` (or both) as the culprit.

Drafter detects [an error on the blueprint](https://github.com/apiaryio/drafter/blob/8dd33e76654901a7ea076fbd667fdef5b9962b27/src/refract/Element.cc#L182) but it is reported to dredd as successful parse with warnings. However, the parse result has no elements to process:

```{ _version: '2.2',
  warnings:
[...]

Issue #354, opened 2016-01-21
2nd level MSON structure not implemented in JSON schema?


mekanixdk:
This might be a dupe of https://github.com/apiaryio/dredd/issues/319 . But as it is marked at fixed I am thinking it might be something else as dredd/protagonist are all up to date.

I am seeing this from a little testcase I made (actually I expanded a little on http://dredd.readthedocs.org/en/latest/example/):

```
# dredd  blueprint.apib http://localhost:3000
info: Beginning Dredd testing...
fail: GET / duration: 38ms
info: Displaying failed tests...
fail: GET / duration: 38ms
fail: body: JSON schema is not valid draft v4! Invalid type: string (expected object) at path "/properties/type"

request:
body:

headers:
    User-Agent: Dredd/1.0.4 (Darwin 15.2.0; x64)

uri: /
method: GET


expected:
headers:
    Content-Type: application/json; charset=utf-8

body:
{
  "message": "Hello Word!",
  "myobject": {
    "item": ""
  }
}
statusCode: 200
bodySchema: {
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "message": {
      "type": "string"
    },
    "type": "object"
  },
  "required": [
    "message"
  ]
}


actual:
statusCode: 200
headers:
    x-powered-by: Express
    content-type: application/json; charset=utf-8
    content-length: 54
    etag: W/"36-qupWQ33wo1wPtGnodw5i6g"
    date: Thu, 21 Jan 2016 20:51:46 GMT
    connection: close

body:
{"message":"Hello World!","myobject":{"item":"value"}}



complete: 0 passing, 1 failing, 0 errors, 0 skipped, 1 total
complete: Tests took 46ms
```

app.js:
```javascript
var app = require('express')();

app.get('/', function (req, res) {
  res.json({message: 'Hello World!', myobject: {item: "value"}});
})

var server = app.listen(3000);
```

blueprint.apib
```md
FORMAT: 1A

# API

# Group X

## GET /

+ Response 200 (application/json; charset=utf-8)

    + Attributes (Testing)

# Data Structures

## Testing (object)
+ message: Hello Word! (string, required)
+ Include Level Two

## Level Two (object)
+ myobject (object, required)
  + item (string, required)
```

I am quite new to API Blueprint so I am possibly missing something here.

philsturgeon:
I think I'm also suffering this same problem.

w-vi:
[...]

PR #402, opened 2016-03-08
Pinning Protagonist

### LINKED ISSUE #394: Dredd should have completely strict dependency on Drafter aka Protagonist

honzajavorek:
Pinning Protagonist to the latest version working with our regression tests.

----------------

- Closes https://github.com/apiaryio/dredd/pull/380.
- Fixes #394.
- Should also address https://github.com/apiaryio/dredd/issues/354 and https://github.com/apiaryio/dredd/issues/381.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "node-uuid": "~1.4.2",
     "optimist": "~0.6.1",
     "pitboss-ng": "^0.3.1",
-    "protagonist": "^1.2.5",
+    "protagonist": "^1.2.6",
     "proxyquire": "~1.7.x",
     "request": "^2.53.0",
     "setimmediate": "^1.0.2",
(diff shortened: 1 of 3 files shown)


honzajavorek:
This did not prove as a solution. Even tried to [return to drafter.js](https://github.com/apiaryio/dredd/tree/honzajavorek/return-to-drafter), but the result wasn't something we could use. Dropping this idea and PR. All we can do is pin protagonist to the latest version as in #399 and urge fixing Drafter bugs.

PR #413, opened 2016-03-12
Linter of CLI docs

### LINKED ISSUE #377: Documentation of CLI parameters should be generated
At least following `-a/--language` isn't appearing in [docs](http://dredd.readthedocs.org/en/latest/usage/), but is available in current Dredd version.

Maybe we could even make a little script which takes the output of `dredd --help` and puts it into docs when building them, so we don't have to care about this manually.

honzajavorek:
Fixes #377.


honzajavorek:
Closed in favor of #464.

PR #420, opened 2016-03-15
Cosmetic changes in README

honzajavorek:
Node versions are probably not so important information they would have to take up space in README.


miiila:
:+1:

Issue #421, opened 2016-03-17
Allow hiding the values of certain headers on failures


realityking:
The source for our API documentation is public and tested on TravisCI. When a test failure occurs it includes the authentication token which is a security issue.

The best solution from our end would be, if we could pass a list of headers whose value should never be printed to the CLI output.

honzajavorek:
This makes sense to me! Thanks for the idea. What would be @netmilk's thoughts on this?

honzajavorek:
@realityking We're in the process of designing how removing sensitive data from Dredd reports could work in https://github.com/apiaryio/dredd/issues/623#issuecomment-246737489. Would you mind looking at the proposal and comment / give us feedback if that would work for you?

PR #426, opened 2016-03-22
Add support for hiding sensitive headers in reports.

realityking:
This is an implementation for #421.

This is very much WIP. Stuff do to includes:

* [ ] Write tests
* [ ] Apiary reporter doesn't censor tests

I'd appreciate it if somebody could look over the code I do have and some help/guidance on the open tasks.

### DIFF
--- a/src/apply-configuration.coffee
+++ b/src/apply-configuration.coffee
@@ -31,6 +31,7 @@ applyConfiguration = (config) ->
       output: null
       debug: false
       header: null
+      'private-header': null
       user: null
       'inline-errors':false
       details: false
@@ -68,6 +69,7 @@ applyConfiguration = (config) ->
   configuration.options.reporter = coerceToArray(configuration.options.reporter)
   configuration.options.output = coerceToArray(configuration.options.output)
   configuration.options.header = coerceToArray(configuration.options.header)
+  configuration.options['private-header'] = coerceToArray(configuration.options['private-header'])
   configuration.options.method = coerceToArray(configuration.options.method)
   configuration.options.only = coerceToArray(configuration.options.only)
   configuration.options.path = coerceToArray(configuration.options.path)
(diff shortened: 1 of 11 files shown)


honzajavorek:
@realityking Thanks very much! We'll take a look at it with @netmilk and let you know.

netmilk:
Hey! I think It's a super useful feature and we totally need it, but I'm not 100% if this is the right way how to implement it.

A super quick workaround would be to use a `afterEachValidation` hook and replace the header value there.

```js
var hooks = require('hooks');

SECRET_HEADERS = ['x-my-secret-header'];

hooks.afterEachValidation(transaction, callback){
  for(var key in transaction['real']]['headers']) {
    if(SECRET_HEADERS.indexOf(key) > -1){
      transaction['real']]['headers'][key] = "***HIDDEN SECRET DATA***";
    }
  }
}
```

realityking:
Interesting workaround, didn't think of that approach.

Out of curiosity, what downsides do you see with the approach I took?

realityking:
I finally got around to testing @netmilk's workaround. Turns out there's no `afterEachValidation` hook but it does work with the `afterEach` hook.

My code:

```js
var hooks = require('hooks');
const SECRET_HEADERS = ['Authorization'].map(header => header.toLowerCase());

hooks.afterEach(function(transaction, done) {
  Object.keys(transaction['request']['headers']).forEach(function(key) {
    if(SECRET_HEADERS.indexOf(key.toLowerCase()) > -1){
      transaction['request']['headers'][key] = "***HIDDEN SECRET DATA***";
    }
  });

  done();
});
```

From my perspective this can now be closed. 👍

honzajavorek:
I'm glad to hear that! Thanks for giving it a try and getting back. I think I'll create another issue anyway to add this to docs.

PR #427, opened 2016-03-24
Little Polishing

honzajavorek:
See commit messages.


w-vi:
@honzajavorek  seems like `coffee-errors` are still needed.
From the Travis build log `Error: Cannot find module 'coffee-errors'`

honzajavorek:
@w-vi Thanks! Should be fixed now. It was required in some files. I removed the requires as the thing shouldn't really be needed with coffee-script 1.7.0+.

PR #451, opened 2016-04-20
Remove 'setimmediate' shim

honzajavorek:
Node 0.9 isn't supported by Dredd anymore. The shim isn't needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "protagonist": "1.3.0",
     "proxyquire": "~1.7.x",
     "request": "^2.53.0",
-    "setimmediate": "^1.0.2",
     "spawn-args": "^0.1.0",
     "spawn-sync": "^1.0.11",
     "uri-template": "~1.0.0",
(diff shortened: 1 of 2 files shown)


w-vi:
In docs I see support for node 0.8 https://github.com/apiaryio/dredd/blob/master/docs/quickstart.md#quickstart

so how is it?

honzajavorek:
Hm, good catch! Docs are wrong. [QA is the source of truth.](https://github.com/apiaryio/dredd/blob/master/.travis.yml#L7:L12)

w-vi:
This PR should fix the docs too then :)

honzajavorek:
@w-vi Just created another one: https://github.com/apiaryio/dredd/pull/452

w-vi:
cool, so I think this is good to go even if the coverage has decreased :)

honzajavorek:
@w-vi Thanks! Yeah, it would be tough to achieve the same coverage with this one! :smile: It probably dropped by 0.00001 or something.

w-vi:
0.002 so a serious drop :D

PR #452, opened 2016-04-20
Remove supported Node versions from docs

honzajavorek:
I already removed similar paragraphs from index.md and README.md, this slipped through the net. We support all recent versions of Node and we will always strive to do so. I consider such information redundant and hard to maintain in docs. If somebody wants to know, I'm okay with replying individually and sending them to CI config as a source of truth.


w-vi:
Another coverage decrease, more of this and you will end up with no coverage at all ;)

PR #453, opened 2016-04-20
Delete ApiaryReporter.md

honzajavorek:
I don't know what this file is about, but I believe it's out of sync with any kind of reality for a long time and doesn't describe anything which would be used anywhere in Dredd or Apiary.

@w-vi @netmilk @kuba-kubula Could you please confirm my speculation (by merging this)?


w-vi:
I don't know but apiary reporter still exists and the settings are still the same aren't they? Is this information described somewhere else?

honzajavorek:
@w-vi There's [ApiaryReportingApi.md](https://github.com/apiaryio/dredd/blob/master/ApiaryReportingApi.md).

w-vi:
@honzajavorek that is API Blueprint of the apiary reporter not the client side settings.

honzajavorek:
@w-vi I see. Then I think this should live in the Dredd Documentation as it has nothing to do with the actual API, it's about usage of Dredd's Apiary reporter.

w-vi:
@honzajavorek  :+1:

honzajavorek:
Closing in favor of #464.

Issue #186, opened 2015-05-09
Modifying the request URI in a dredd hook doesn't work


jolleon:
I'm trying to write a dredd *before* hook so that my DELETE action can target the resource that was created by a previous POST.

I've been following [the doc here](https://github.com/apiaryio/dredd/blob/e5e4a5ccab256e2ebde529e8c69b7d1d30294097/docs/hooks.md#sharing-data-between-steps-in-request-stash), but updating `transaction.request.uri` doesn't seem to work (BTW the doc is outdated and refers to `url` while the field seems to be `uri` - I've tried setting both to a different value without success). The updated url does show up on the report that dredd uploads to apiary, but my server is still receiving the old URL.

For the record I'm able to successfully update the request body and headers in a *before* hook, I think the bug is only in the URL/URI.

netmilk:
Hi @jolleon thanks for the report! I'm excited you're using Dredd and the test inspector in Apiary. I'll have a look on it asap.

ecordell:
I think I can help out here - there are a couple of idiosyncrasies with changing the request url. The actual attribute used to construct the request is `fullPath`. This is how I change the url in before hooks:

```javascript
before("Hook Name", (transaction) => {
  transaction.fullPath = "/new/path";
  return transaction.uri = transaction.fullPath;
});
``` 

`transaction.fullPath` sets the path the request will be sent to, and `transaction.request.uri` sets the path that will be recorded in logs and reports. It's a little confusing - I think it would be a good plan to promote "transaction" to an object with getters/setters so that you only need to change one of those properties. 

The above code works for me with the latest dredd version, so if it doesn't work for you you might try upgrading? And you're right - those docs are incorrect. The property has always been `uri`.

jolleon:
Yep this works, thanks! I also found the "uriTemplate" on the transaction object which seems to be a relatively clean way to handle my use case:

    transaction.fullPath = transaction.origin.uriTemplate.replace('{id}', stash['userId']);
    transaction.request.uri = transaction.fullPath

netmilk:
Hi,

[...]

PR #470, opened 2016-05-04
Docs update of hook example code for replacing URL params

lazlojuly:
Updates doc with suggested method for replacing url params described in #186 .


honzajavorek:
Thanks! :balloon:

Issue #478, opened 2016-05-10
Cannot read property 'replace' of null


LinusU:
I'm getting this error when trying out Dredd version `1.1.0-pre.1`, it worked at version `1.0.10`.

```text
/.../node_modules/dredd/lib/transaction-runner.js:339
      fullPath = '/' + [this.parsedUrl['path'].replace(/^\/|\/$/g, ""), request['uri'].replace(/^\/|\/$/g, "")].join("/");
                                                                                      ^

TypeError: Cannot read property 'replace' of null
  at TransactionRunner.configureTransaction (/.../node_modules/dredd/lib/transaction-runner.js:339:87)
  at TransactionRunner.configureTransaction (/.../node_modules/dredd/lib/transaction-runner.js:3:57)
  at /.../node_modules/async/lib/async.js:246:17
  at iterate (/.../node_modules/async/lib/async.js:146:13)
  at /.../node_modules/async/lib/async.js:157:25
  at /.../node_modules/async/lib/async.js:248:21
  at TransactionRunner.configureTransaction (/.../node_modules/dredd/lib/transaction-runner.js:381:12)
  at TransactionRunner.configureTransaction (/.../node_modules/dredd/lib/transaction-runner.js:3:57)
```

Ramblurr:
I'm a fresh dredd user (literally npm install -ged a few minutes ago) running node v.5.5.0, and I'm getting this error when my `endpoint` config has a path after the `host:port`.

How can I work around this for now?

Ramblurr:
Looks like I had  installed `v1.1.0-pre.1`. Downgrading to `v1.0.10` fixed it (v1.0.11 isn't on npm yet).

honzajavorek:
Thanks for reporting this! That's definitely a regression we need to fix. Yes, downgrading is a preferred option right now.

smizell:
@LinusU I was wondering if you include any additional information to help debug this? I have seen this issue come up when people have a URI template but did not define the parameters for it.

Example:

```apib
FORMAT: 1A

# My API

# Item [/item{?foo}]
Parameters should be defined here, but they are not.

## GET
+ Response 200
```

Do you have anything like this in your blueprint?

smizell:
@LinusU I have added some more information to an issue that may be related. Maybe this is what you're seeing as well?

https://github.com/apiaryio/dredd/issues/469

honzajavorek:
[...]

Issue #469, opened 2016-05-04
Undocumented parameters raise an exception


demianh:
NPM updated my dredd to the latest release v1.1.0-pre.0, and now my tests are failing.

I think it has to do with parameters that aren't documented. See ?sort in "Document List", it is in the URL, but not listed in the documentation. 
In earlier versions of dredd, they just were omitted. Now i get an Exception. Not sure if this new behaviour is intended, but the old behaviour was definitely more convenient :)

I'll probably go and document all the undocumented parameters in each section anyway (currently they are documented as plain text in a common section) but i wanted to report this regression.

```apib
FORMAT: 1A

# My API

# Group Document

## Document List [/document{?sort}]

### Create Document [POST]
Create a document

+ Request
    [Document][]

+ Response 201

        18010

## Document [/document/{id}]
Get a document

+ Parameters
    + id (required, number, `1804`) ... Document ID.

+ Model (application/json)

    + Body

            {
                "title": "factsheet.pdf",
            }

```

I am calling dredd like this:

`/vagrant/node_modules/dredd/bin/dredd api/documentation/webling2.md http://demo.webling.dev/api/1`

Output:

```
warn:  Compilation warning in file 'api/documentation/webling2.md': 
Ambiguous URI parameter in template: /document{?sort} 
Parameter not defined in blueprint:'sort' on Document > Document List > Create Document 
info: Beginning Dredd testing...

/vagrant/node_modules/dredd/lib/transaction-runner.js:339
' + [this.parsedUrl['path'].replace(/^\/|\/$/g, ""), request['uri'].replace(/^
                                                                    ^
TypeError: Cannot call method 'replace' of null
  at TransactionRunner.configureTransaction (/vagrant/node_modules/dredd/lib/transaction-runner.js:339:88)
  at TransactionRunner.configureTransaction (/vagrant/node_modules/dredd/lib/transaction-runner.js:3:57)
  at /vagrant/node_modules/dredd/node_modules/async/lib/async.js:246:17
  at iterate (/vagrant/node_modules/dredd/node_modules/async/lib/async.js:146:13)
  at async.eachSeries (/vagrant/node_modules/dredd/node_modules/async/lib/async.js:162:9)
  at _asyncMap (/vagrant/node_modules/dredd/node_modules/async/lib/async.js:245:13)
  at Object.mapSeries (/vagrant/node_modules/dredd/node_modules/async/lib/async.js:228:23)
[...]

PR #484, opened 2016-05-13
Ambiguous parameters raise exception (fix)

honzajavorek:
1. Fixes apiaryio/dredd#469 and fixes apiaryio/dredd#478.
2. Some "on the way" clean up work attached (linter, whitespace, redundant files).

I suggest to review commit-by-commit. It should be easy reading. The 1. takes _single line_ and the rest is just cleanup.


package.json
```diff
@@ -31,7 +31,7 @@
     "chai": "^3.4.1",
     "clone": "^1.0.0",
     "coffee-script": "^1.9.1",
-    "dredd-transactions": "0.0.4",
```
honzajavorek: This is the main change in this PR, actually.

honzajavorek:
Coverage drop is probably due to removal of files and corresponding tests. There's nothing to cover in this PR.

PR #485, opened 2016-05-16
Preventing NaN in printed annotations

honzajavorek:
Preventing

```
(warning code 1) expected API name, e.g. '# <API Name>' on lines NaN-NaN, lines NaN-NaN
```

### DIFF
--- a/src/blueprint-utils.coffee
+++ b/src/blueprint-utils.coffee
@@ -1,25 +1,25 @@
-newlineRegExp = /\n/g
 
-blueprintUtils = {}
-
-blueprintUtils.characterIndexToPosition = (charIndex = 0, text = '') ->
-  pieceOfCode = text.substring 0, charIndex
-  return {
-    row: pieceOfCode.match(newlineRegExp)?.length + 1
-  }
+NEWLINE_RE = /\n/g
 
 
-blueprintUtils.sortNumbersAscending = (a, b) ->
+characterIndexToPosition = (charIndex = 0, code = '') ->
+  codeFragment = code.substring(0, charIndex)
+  row = (codeFragment.match(NEWLINE_RE)?.length or 0) + 1
+  {row}
+
+
+sortNumbersAscending = (a, b) ->
   return a - b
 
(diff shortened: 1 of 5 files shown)


honzajavorek:
The coverage dropped, but actually probably just because the overall percentage of lines has changed. I'm not going to add any tests, there's nothing left to cover in parts of the code I touched.

src/blueprint-utils.coffee
```diff
@@ -58,4 +58,9 @@ blueprintUtils.rangesToLinesText = (ranges) ->
   return pos
 
 
-module.exports = blueprintUtils
```
freaz: __To think about:__ The original approach is imo better because it allows you to stub/spy functions inside module.

test/unit/handle-runtime-problems-test.coffee
```diff
@@ -0,0 +1,186 @@
...
+      FORMAT: 1A
+      # Beehive API
+      \t\t
+    '''
+    filename = 'dummy-filename.apib'
+
+    beforeEach((done) ->
+      prepareData(apiDescriptionDocument, filename, (err, data) ->
+        error = handleRuntimeProblems(data)
+        done()
```
freaz: `done(err)`

src/blueprint-utils.coffee
```diff
@@ -1,25 +1,25 @@
...
 
-blueprintUtils.characterIndexToPosition = (charIndex = 0, text = '') ->
-  pieceOfCode = text.substring 0, charIndex
-  return {
-    row: pieceOfCode.match(newlineRegExp)?.length + 1
-  }
 
+characterIndexToPosition = (charIndex = 0, code = '') ->
+  codeFragment = code.substring(0, charIndex)
+  row = (codeFragment.match(NEWLINE_RE)?.length or 0) + 1
```
freaz: So this `or 0` is only real change in this PR right?

honzajavorek: Yes.

PR #487, opened 2016-05-17
Forgot to handle errors in test

honzajavorek:


test/unit/handle-runtime-problems-test.coffee
```diff
@@ -62,7 +62,7 @@ describe('handleRuntimeProblems()', ->
     beforeEach((done) ->
       prepareData(apiDescriptionDocument, filename, (err, data) ->
         error = handleRuntimeProblems(data)
```
freaz: WIll `handleRuntimeProblem` survive if `data === undefined`?

honzajavorek: I should stop doing before thinking. Happens to me a lot lately :neutral_face: (this is genuine neutral face, not Dredd's smile)

honzajavorek: Addressed (commit amended).

honzajavorek:
@freaz said (in person) it's okay now.

Issue #467, opened 2016-04-30
Recent change in docs made some old links 404


JackuB:
I've noticed docs link that got changed

```diff
- https://dredd.readthedocs.org/en/latest/usage/
+ https://dredd.readthedocs.org/en/latest/usage-cli/
```

Can a 301 redirect be added? For the old one?

honzajavorek:
I doubt it, it's mkdocs & RTD. But I can try.

honzajavorek:
Tried this

<img width="622" alt="screen shot 2016-05-02 at 16 26 07" src="https://cloud.githubusercontent.com/assets/283441/14956917/a271d672-1082-11e6-84aa-1aa026d28f35.png">

but still experiencing 404s. Not sure whether I should wait or whether it just doesn't work.

JackuB:
Nope, still 404. Not sure how RTD redirect works, but maybe it needs also the `/en/latest/` part http://docs.readthedocs.io/en/latest/automatic-redirects.html

honzajavorek:
The redirect should cout with that:

<img width="611" alt="screen shot 2016-05-05 at 12 08 50" src="https://cloud.githubusercontent.com/assets/283441/15041163/29f410ea-12ba-11e6-911b-cfe7c7696745.png">

I added one more, but without success to change anything:

<img width="817" alt="screen shot 2016-05-05 at 12 06 38" src="https://cloud.githubusercontent.com/assets/283441/15041166/392b7634-12ba-11e6-9880-e3f2d7574808.png">

I'm afraid we're being hit by https://github.com/rtfd/readthedocs.org/issues/1826 here.

smizell:
@honzajavorek looks like one hack is to do what is mentioned here:

https://github.com/rtfd/readthedocs.org/issues/1826#issuecomment-214753550

honzajavorek:
@JackuB @smizell Tried to address this by https://github.com/apiaryio/dredd/pull/489. Please review.

honzajavorek:
It works! Also, I think https://github.com/apiaryio/dredd/pull/491 will fix the rest.

PR #489, opened 2016-05-18
Redirect /usage to /usage-cli

honzajavorek:
Attempt to address #467 by @borsboom's workaround from https://github.com/rtfd/readthedocs.org/issues/1826#issuecomment-214753550. I think it will work for `/usage.html`, but I don't know how to do the same for `/usage/`.


JackuB:
LGTM 👍

PR #495, opened 2016-05-24
v1.1.0-pre.3

honzajavorek:
Using Dredd Transactions which are built on top of Fury.js.


pksunkara:
Tests failing. Otherwise 👍

PR #499, opened 2016-05-27
Making sure transaction.request.body is used only if really present

honzajavorek:
Making this change per @ddelnano's findings in https://github.com/snikch/goodman/pull/5#issuecomment-222054192.

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -514,7 +514,7 @@ class TransactionRunner
         @configuration.emitter.emit 'test error', error, test, () ->
         return callback()
 
-      req.write transaction.request['body'] if transaction.request['body'] != ''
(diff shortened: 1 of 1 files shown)


ddelnano:
@honzajavorek also linking to this to discover my findings after making it a less strict comparison [link](https://github.com/snikch/goodman/pull/5#issuecomment-222188389)

honzajavorek:
@ddelnano The change in this PR probably doesn't have a large impact on the issues you have, but in my opinion it's wrong and won't work in some cases, so it should be fixed anyway.

ddelnano:
@honzajavorek Yea I have dredd-hooks-go working properly as of last night.  Will be pushing my changes.  I was still not able to figure out what the root of the problem was for the issue I experienced but the work around to prevent dredd crashing was the to make sure the request body had some data in it.

PR #505, opened 2016-06-01
Update docs and hooks worker client for go hooks

ddelnano:
This PR depends on [this](https://github.com/snikch/goodman/pull/5).  Once that is merged, the go hooks should be considered compliant with the hooks for other languages: php, ruby, python, perl, etc.

I tested the changes with this [repo](https://github.com/ddelnano/dredd-hooks-go-example).  I could not install dredd from a github repo since npm does not run prepublish and the necessary commands to do so I made these commits on my fork and made sure the examples worked.

screen shot 2016-05-31 at 10 58 52 pm

### DIFF
--- a/docs/hooks-go.md
+++ b/docs/hooks-go.md
@@ -10,21 +10,49 @@ Go hooks are using [Dredd's hooks handler socket interface](hooks-new-language.m
 
 ```
 $ go get github.com/snikch/goodman
+$ cd $GOPATH/src/github.com/snikch/goodman
+$ go build -o $GOPATH/bin/goodman github.com/snikch/goodman/bin
 ```
 
 ## Usage
 
-Using Dredd with Go is slightly different to other languages, as a binary needs to be compiled for execution. Instead of providing a variety of hookfiles to run, your own binary is built using the Go hooks library `goodman` to run your specific hooks. The path to this binary is passed to dredd via the `--language` flag. You may also need to supply a `--hookfiles` flag to get the tests to run, but the value of this is of no matter (the dredd runner requires it).
+Using Dredd with Go is slightly different to other languages, as a binary needs to be compiled for execution. The --hookfiles flags will be compiled hook binaries.  See below for an example hooks.go file to get an idea of what this would look like.
 
 ```
-$ dredd apiary.apib http://localhost:3000 --language ./my-dredd-hook-server --hookfiles *.go
+$ dredd apiary.apib http://localhost:3000 --server ./go-lang-web-server-to-test --language go --hookfiles ./hook-file-binary
 ```
 
 ## API Reference
 
-You’ll need to know about three things to get your own hook handler up and running.
+In order to get a general idea of how the Go Hooks work, the main executable from the package $GOPATH/bin/goodman is an HTTP Server that Dredd communicates with and an RPC client.  Each hookfile then acts as a corresponding RPC server.  So when Dredd notifies the Hooks server what transaction event is occuring the hooks server will execute all registered hooks on each of the hookfiles RPC servers.
 
-1. A `Runner` is a type that you can define event callbacks on, such as `beforeEach`, `afterAll` etc.
+You’ll need to know a few things about the `Server` type in the hooks package.
+
+1. The `hooks.Server` type is how you can define event callbacks such as `beforeEach`, `afterAll`, etc.
+
+2. To get a `hooks.Server` struct you must do the following
+
+```go
+package main
+
+import (
+    "github.com/snikch/goodman/hooks"
+    trans "github.com/snikch/goodman/transaction"
+)
+
+func main() {
+    h := hooks.NewHooks()
+    server := hooks.NewServer(h)
+
+    // Define all your event callbacks here
+
+    // server.Serve() will block and allow the goodman server to run your defined
+    // event callbacks
+    server.Serve()
+    // You must close the listener at end of main()
+    defer server.Listener.Close()
+}
+```
 
 2. Callbacks receive a `Transaction` instance, or an array of them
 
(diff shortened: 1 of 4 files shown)


ddelnano:
@honzajavorek The PR this depended on is finally [merged](https://github.com/snikch/goodman/pull/5)

honzajavorek:
@ddelnano Great! I asked our Go lover @pksunkara to look at it once he has a free moment, so expect him to land & comment here soon.

ddelnano:
@honzajavorek I am having problems with the Go hooks dealing with requests that have an empty body.  Like the ones in the hooks template.  If I add a request body all the aruba/cucumber tests pass.  If I don't I get this dredd error.
screen shot 2016-06-03 at 9 00 04 am

Seems to me to be kind of like the error in #499 where it shouldn't allow whatever value that is to be passed to Buffer.byteLength, but I think there is some other issue where maybe the go hooks send back bad data or something that causes dredd to have these errors.

ddelnano:
@honzajavorek so the CI build for the go hooks pass if I add this to every .feature file

https://github.com/snikch/goodman/blob/master/features/execution_order.feature#L19-L20

ddelnano:
@honzajavorek deleted my previous comments, cause I thought this was on an issue and not this PR.  Sounds good to me 👍

docs/hooks-go.md
```diff
@@ -10,21 +10,49 @@ Go hooks are using [Dredd's hooks handler socket interface](hooks-new-language.m
...
 ```
 $ go get github.com/snikch/goodman
+$ cd $GOPATH/src/github.com/snikch/goodman
+$ go build -o $GOPATH/bin/goodman github.com/snikch/goodman/bin
 ```
 
 ## Usage
 
-Using Dredd with Go is slightly different to other languages, as a binary needs to be compiled for execution. Instead of providing a variety of hookfiles to run, your own binary is built using the Go hooks library `goodman` to run your specific hooks. The path to this binary is passed to dredd via the `--language` flag. You may also need to supply a `--hookfiles` flag to get the tests to run, but the value of this is of no matter (the dredd runner requires it).
+Using Dredd with Go is slightly different to other languages, as a binary needs to be compiled for execution. The --hookfiles flags will be compiled hook binaries.  See below for an example hooks.go file to get an idea of what this would look like.
```
pksunkara: > hookfiles flags should point to compiled hook binaries.

> get an idea of what the source file behind the go binary would look like.

docs/hooks-go.md
```diff
@@ -10,21 +10,49 @@ Go hooks are using [Dredd's hooks handler socket interface](hooks-new-language.m
...
 
 ```
-$ dredd apiary.apib http://localhost:3000 --language ./my-dredd-hook-server --hookfiles *.go
+$ dredd apiary.apib http://localhost:3000 --server ./go-lang-web-server-to-test --language go --hookfiles ./hook-file-binary
```
pksunkara: @honzajavorek What is this server option used for? I thought localhost:3000 is the server.

pksunkara: @ddelnano If you are talking about hook server, maybe rename it to `./golang-hook-server`

ddelnano: @pksunkara the --server option is the command to get the server that will be under test started.  So the command assumes that the `./go-lang-web-server` is a go web service that will start a webserver that listens on localhost port 3000.

ddelnano: @pksunkara That is not referring to the hook server.

pksunkara: Ah, thanks. I didn't know that.

ddelnano: @pksunkara the hooks server command gets set [here](https://github.com/apiaryio/dredd/pull/505/files#diff-fab7c8fd00434dbb9259f70a29405350R138).

docs/hooks-go.md
```diff
@@ -130,21 +156,22 @@ package main
...
-      t.Fail = true
+    h := hooks.NewHooks()
+    server := hooks.NewServer(h)
+    h.Before("Machines > Machines collection > Get Machines", func(t *trans.Transaction) {
+        t.Fail = true
     })
-    runner.Before("Machines > Machines collection > Post Machines", func(t *goodman.Transaction) {
-      t.Fail = "POST is broken"
+    h.Before("Machines > Machines collection > Post  Machines", func(t *trans.Transaction) {
+        t.Fail = "Post is broken"
```
pksunkara: `POST`

pksunkara:
Just 2 minor comments regarding docs. Will merge it after they are fixed.

docs/hooks-go.md
ddelnano: @pksunkara updated.

docs/hooks-go.md
ddelnano: @pksunkara can you be more specific?  Is this just not clear?

pksunkara: I just made the sentences a bit more clear. My comment has the updated sentences which you need to copy.

ddelnano: @pksunkara awesome thanks for clarifying.  I updated it.

ddelnano:
@pksunkara @honzajavorek I updated the PR addressing the feedback **before** this get merged, however, I am thinking there might be something that might need addressed first.  Dredd crashes with the Go hooks if the API Blueprint transaction has an empty request body.  See the difference between the Go [tests](https://github.com/snikch/goodman/blob/master/features/execution_order.feature#L19-L20) and the template [tests](https://github.com/apiaryio/dredd-hooks-template/blob/master/features/execution_order.feature#L16-L20)

ddelnano:
Here is an example of the dredd crash, I am going to try and prepare a fix.

screen shot 2016-06-03 at 1 27 54 pm

ddelnano:
@honzajavorek not sure exactly what the deal is with the findings in my previous comments.  I wrote out a full example using a sample go app, goodman, and dredd compiled from this branch and blueprint files that have empty request bodies run fine.  You can see the example [here](https://github.com/snikch/goodman/pull/7).  Either of you could run the example if you wanted to following the instructions in the Wiki to verify my claims that it works as expected, however, you would need to update the path in the [Makefile](https://github.com/snikch/goodman/pull/7/files#diff-e8849b97cf69febe9876e8ab47d44460R10) since I needed to use dredd compiled from this branch (since it uses these changes).

Let me know what you think.

pksunkara:
Does the above crash happen with other language hook files or only golang? Depending on that, we need to fix it in a separate PR. I am merging this for now.

ddelnano:
@pksunkara when I developed the php hooks I never ran into this.  I also tried to reproduce it with a real example instead of the cucumber tests (where I was seeing it) and I was unable to do so.  I opened a PR for another crash that I found that indicates a hook server failure ( #511 ).  This would just provide better information to the hook server dev instead of seeing the error `cannot read property 'length' of undefined`.

ddelnano:
@pksunkara @honzajavorek when do the pre releases go out?  I would like to merge my example in the goodman repo but since npm's support for installing repos via commit hashes does not run any building (so the coffeescript would not be compiled) I can't merge until this change is in a new release.

honzajavorek:
@ddelnano I should release another one within a day.

ddelnano:
awesome thanks!

Issue #504, opened 2016-05-31
Tests are failing when using multiple blueprints


aledeg:
I have an API with a lot of component. Each component has its own blueprint file and its own dredd config file. It's working perfectly.

My idea is to run all those separated tests at one by including all blueprints in one file. Unfortunately it does not work. Most of the tests are failing because of my 

I have a bunch of hook files which are loading data in the database for each test. My errors are all related to the database not being populated.

The following configurations are not working:

``` yml
...
blueprint: src/doc/bp*.apib
...
```

``` yml
...
blueprint:
    - src/doc/bp*.apib
...
```

``` yml
...
blueprint:
    - src/doc/bp1.apib
    - src/doc/bp2.apib
...
```

The following configuration are working:

``` yml
...
blueprint:
    - src/doc/bp1.apib
...
```

``` yml
...
blueprint: src/doc/bp1.apib
...
```

Am I missing something?
Is this related to dredd or to https://github.com/ddelnano/dredd-hooks-php?

ddelnano:
How are you running dredd?

ddelnano:
@aledeg I am thinking this is probably a dredd issue.  I am not exactly sure how dredd treats multiple blueprint files but in order for that to work dredd would need to execute all transactions for a given blueprint file together (and not merge them) and also shutdown and restart a new hook server for each blueprint file.

ddelnano:
Took a look into this some more and I don't think dredd supports wildcards for the `blueprint` config field.  Someone that contributes to dredd would know more but from what I saw doesn't seem like it is supported.

aledeg:
I am not sure what happens under the hood. But when I provide multiple blueprint files, with wildcards or not, I have the same result.
The blueprints files are loaded since all my actions are tested. The problem is that it looks like the code in the hooks are not executed since I do not have data in my database.

honzajavorek:
I think there are some tests to address this, but not very thorough in means of corner cases:

- https://github.com/apiaryio/dredd/blob/0b1721a889eb420b39b7ec68dbaf46b808f5b574/test/integration/cli/cli-test.coffee#L836
- https://github.com/apiaryio/dredd/blob/0b1721a889eb420b39b7ec68dbaf46b808f5b574/test/unit/dredd-test.coffee#L110
- https://github.com/apiaryio/dredd/blob/22c029382332520f07ee6c274bf117bc95d889bd/test/integration/cli/api-description-cli-test.coffee#L60

[...]

PR #509, opened 2016-06-01
Add failing test for blueprint with glob pattern

aledeg:
See #504

I made sure that blueprints and hooks are working properly when used alone.

There is no problem when using glob pattern for the hook files as long as there is only one blueprint file. Even if the blueprint file is matched with a wildcard. The problem comes only when there is more than one blueprint file in the list.
On the first place, I found out that error while using a configuration file with dredd. I guess it's the same issue.


michalholasek:
As @honzajavorek mentioned in his [comment](https://github.com/apiaryio/dredd/issues/504#issuecomment-351061688), this issue was fixed in https://github.com/apiaryio/dredd/pull/803, which has its own set of tests. Thank you @aledeg for the time you have put into this.

Closes #504.

PR #511, opened 2016-06-03
Add helpful error message if dredd hooks fail to modify transaction

ddelnano:
This PR is to address when developing a hooks server from [dredd-hooks-template](https://github.com/apiaryio/dredd-hooks-template/blob/master/features/execution_order.feature#L78-L93).  If the hooks fail to modify the transaction sent back to dredd, it will crash dredd.  This is to present a more helpful error to the hook developer.  I just went through this for the second time (PHP and Go) and it took me a little digging before I realized I wrote the hookfile wrong.

I tried to add a test but it would have been difficult to be able to stub everything to the point where I could trigger the error.  I didn't get to spend enough time wrapping my head around some of the really long tests.

### DIFF
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -299,7 +299,9 @@ class HooksWorkerClient
       # https://github.com/apiaryio/dredd-hooks-template/blob/master/features/execution_order.feature
       if process.env['TEST_DREDD_HOOKS_HANDLER_ORDER'] == "true"
         console.error 'FOR TESTING ONLY'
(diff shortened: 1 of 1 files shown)


honzajavorek:
Thanks!

honzajavorek:
This seems to break all hook tests:

screen shot 2016-06-09 at 13 16 54

All of them seem to fail at exactly this test. It's probably just something silly, about to investigate reasons. Example build: https://travis-ci.org/apiaryio/dredd/jobs/136387109

PR #523, opened 2016-06-14
Skipping non 2xx responses by default for Swagger

honzajavorek:
Skipping non-2xx Swagger transactions by default, user can unskip them in hooks. This is because non-2xx responses for still the same request are not useful in most flows when testing a working server. Unlike API Blueprint, Swagger doesn't have an ability to specify specific request-response pairs, which address the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "clone": "^1.0.0",
     "coffee-script": "^1.10.0",
     "colors": "^1.1.2",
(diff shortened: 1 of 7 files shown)


pksunkara:
@honzajavorek Conflicts. Please rebase.

honzajavorek:
@pksunkara rebased

PR #542, opened 2016-06-29
Add the ability to define environment variables

castarco:
This can be very useful to avoid having to define another setting in the dredd.yml file and also to avoid having to wrap our commands to be able to define environment variables.

### DIFF
--- a/src/dredd-command.coffee
+++ b/src/dredd-command.coffee
@@ -202,14 +202,12 @@ class DreddCommand
       parsedArgs = spawnArgs(@argv['server'])
       commandExtraOptions = {}
       command = parsedArgs.shift()
-      
+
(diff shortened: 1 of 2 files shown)


castarco:
Any idea on how to cover this patch with tests? I'm not a coffeescript expert and I'm missing something.

honzajavorek:
@castarco Hi, I'm sorry I wasn't very responsive in the past. I think I cannot accept this patch. First, I think https://github.com/apiaryio/dredd/issues/554#issuecomment-237683530 is the right way to go if you need this, second, IMHO it would be very hard to get this working under Windows. Anyway, I really appreciate you took the effort and sent a PR. It's a shame we didn't pick it up sooner and decided sooner.

PR #547, opened 2016-07-01
Added Go to hook handlers integration tests

honzajavorek:
Let's see whether everything works together!

### DIFF
--- a/scripts/test-hooks-handlers.coffee
+++ b/scripts/test-hooks-handlers.coffee
@@ -82,6 +82,9 @@ JOBS = [
   ,
     name: 'perl-hooks-handler'
     repo: 'https://github.com/ungrim97/Dredd-Hooks.git'
+  ,
+    name: 'go-hooks-handler'
+    repo: 'https://github.com/snikch/goodman.git'
 ]
 
 TRIGGER_KEYWORD = 'tests hook handlers' # inspired by https://help.github.com/articles/closing-issues-via-commit-messages/


honzajavorek:
Hmm, I think it won't pass, because there is no `npm install dredd -g` which could be replaced, in the https://github.com/snikch/goodman/blob/master/.travis.yml file.

ddelnano:
@honzajavorek does it need to be installed globally?

honzajavorek:
It's just that the testing script doesn't count with such setup, that's
all. It looks for "npm install dredd" in the Travis' file and attempts to
replace it with "npm install ./__dredd__/...", which is a folder where the
tested version of Dredd lives during the test.

In this case, we would probably have to also check for package.json and
somehow inject the tested Dredd installation there.

ddelnano:
@honzajavorek I keep getting build errors on the `npm install` when it is trying to build protagonist.  Here is one of the failed [jobs](https://travis-ci.org/snikch/goodman/builds/141841600).  I originally was using whatever node version travis pre installs (something 0.10.x) but then tried using 5.11.x and still received the same failures.

ddelnano:
@honzajavorek so I had another PR open on the goodman repo that I tried re running the build (it was failing for a different reason) and now it can't `npm install` either.  I looked at protagonist and saw they released a new version yesterday and that the current dredd package.json would have installed that version.

honzajavorek:
@ddelnano Protagonist should install all right. There were some issues with the latest version on Mac, but they released a fix (1.4.1) immediately - https://github.com/apiaryio/protagonist/releases

From the build you linked it seems to me you need to turn Travis into a `sudo` mode, otherwise Protagonist will have hard time during installation, since there are probably some essential libs / build tools missing. See `dredd-example`'s minimal config: https://github.com/apiaryio/dredd-example/blob/master/.travis.yml

I didn't realise this before, but since Protagonist is Dredd's prominent dependency, we should probably mention this somewhere and test out again (and update) our Travis CI configs we have all over the docs :grimacing:

We soon plan to depend on js-only version of the parser by default though.

ddelnano:
@honzajavorek I will try using sudo in the travis build but what is weird is how my other build worked previously and now it is failing on the install.

ddelnano:
@honzajavorek I updated the build with `sudo: required` and it is still failing on installation, when 2 weeks ago when I was working on the PR it was working fine https://travis-ci.org/snikch/goodman/jobs/142013584 .

Here is the PR https://github.com/snikch/goodman/pull/8

honzajavorek:
See https://github.com/snikch/goodman/pull/10. Just tested out on our hooks repos and it seems to work now.

honzajavorek:
@ddelnano Thanks! That should do the trick for our testing script without any changes. Let's see how it goes.

honzajavorek:
🎉 

https://travis-ci.org/apiaryio/dredd/builds/145563920

screen shot 2016-07-18 at 18 08 17

ddelnano:
👏

Issue #339, opened 2015-12-30
Idea: Ability to specify config file


romansklenar:
Hello, this would be a nice feature for dredd to have. I'll try to describe it.

__Use case:__
On CI server I need to run dredd with different configuration than locally.

__Solution:__
Adding an option for dredd to choose config file via optional parameter would solve this issue.
Consider following:
```bash
dredd # defaults to dredd.yml
dredd --config dredd.ci.yml # or dredd -f dredd.ci.yml
```

netmilk:
HI @romansklenar, 

This makes perfect sense to me. I think it will be very easy to implement and it can be coupled with some work on CLI arguments in very near future.

#328, #235, #323, #321, #270, #251

romansklenar:
Thanks! :heart: Looking forward!

mike-marshall0164:
I just went searching to see if this was possible, glad I'm not the only one!

honzajavorek:
Implemented by #548!

PR #548, opened 2016-07-04
--config option

honzajavorek:
Implemented #339. Needed for https://github.com/apiaryio/dredd-example/ so it is able to pick both API Blueprint and Swagger during one build.

------

**Please ignore AppVeyor, it's not properly set up and actively used yet.**


pksunkara:
Merge away when tests pass. 👍

PR #549, opened 2016-07-04
v1.2.0

honzajavorek:
See changelog to read about what has been added.


honzajavorek:
I had to cancel some builds since Travis CI was very slow this afternoon in spinning up builds. However, I checked everything important twice and made sure both Dredd test suite and hooks' test suites pass.

PR #573, opened 2016-07-22
Dredd is now installable also without C++ compiler

honzajavorek:
- Dredd is now installable also without C++ compiler
- Contains also bug fix for https://github.com/apiaryio/dredd/pull/570


honzajavorek:
According to Travis CI there seem to be two issues with this related to 0.10 and 0.12 node versions:

1. Trouble to get `test/unit/hooks-worker-client-test.coffee` passing.
2. `./scripts/cov` seems to hang and builds get terminated although otherwise they pass.

honzajavorek:
The issues are probably not caused by drafter.js *per se*, but rather by running Dredd's test suite on the lightweight Travis CI infrastructure for the first time.

PR #585, opened 2016-08-01
Update hooks-nodejs.md

BrydenR:
requestStash is not a variable in this program. I think it may be a typo, where requestStash was supposed to be responseStash.


honzajavorek:
@BrydenR You are correct! Thank you!

Issue #578, opened 2016-07-27
Vulnerability in minimatch


honzajavorek:
There's a [vulnerability in `minimatch` library](https://snyk.io/vuln/npm:minimatch:20160620), which propagates to Dredd through other dependencies. I've

- Sent a PR to `html` dependency to solve this: https://github.com/maxogden/commonjs-html-prettyprinter/pull/11

- After some time nobody reacted to the PR, I created a fork of the project in [apiaryio/commonjs-html-prettyprinter](https://github.com/apiaryio/commonjs-html-prettyprinter) and Dredd will depend on that (I re-created the PR to upstream to go from our fork: https://github.com/maxogden/commonjs-html-prettyprinter/pull/12)

galenandrew:
@honzajavorek looks like maxogden/commonjs-html-prettyprinter#14 has been merged, so this issue can be resolved by updating to `html@1.0.0` (this would close #589).

honzajavorek:
@galenandrew Ah, we just merged our own crafted fix. Good! I'll put back upstream dependency.

PR #589, opened 2016-08-05
Address security issue in the HTML package

honzajavorek:
Closes https://github.com/apiaryio/dredd/issues/578. Uses https://github.com/apiaryio/commonjs-html-prettyprinter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "file": "^0.2.2",
     "gavel": "^0.5.3",
     "glob": "^7.0.5",
-    "html": "0.0.10",
+    "html": "git://github.com/apiaryio/commonjs-html-prettyprinter.git#honzajavorek/upgrade-glob",
     "htmlencode": "0.0.4",
     "inquirer": "^1.1.0",
     "js-yaml": "^3.6.1",


package.json
```diff
@@ -37,7 +37,7 @@
     "file": "^0.2.2",
     "gavel": "^0.5.3",
     "glob": "^7.0.5",
-    "html": "0.0.10",
+    "apiaryio/commonjs-html-prettyprinter": "",
```
abtris: This can be valid version or publish npm package
`"apiaryio/commonjs-html-prettyprinter": "https://github.com/apiaryio/commonjs-html-prettyprinter",`

honzajavorek: The thing is, for `npm install <something>` it works: https://docs.npmjs.com/cli/install

> Note: if you do not include the @-symbol on your scope name, npm will interpret this as a GitHub repository instead, see below. Scopes names must also be followed by a slash.

```
npm install githubname/reponame
```

But it doesn't work for `package.json`: https://docs.npmjs.com/files/package.json#github-urls

abtris:
👍

Issue #625, opened 2016-09-12
Dredd should simplify reporting of bugs


honzajavorek:
### How?

- GitHub Issue Template - could it help? - https://github.com/blog/2111-issue-and-pull-request-templates
- Dredd's `version` command could contain debugging information user should post when creating issue or asking for help.
- If Dredd fails, it could print out "new issue link" with all necessary debugging info pre-filled: https://github.com/apiaryio/dredd/issues/new?body=hello

### What?

- version
- drafter.js or protagonist?
- OS type and version
- some environment variables or just whether some environment variables are filled or not
- what command was exactly executed and what is in the `dredd.yml`, or basically with which options Dredd was ran

PR #626, opened 2016-09-12
Report OS type and version in dredd --version

honzajavorek:
Addressing the simplest item from https://github.com/apiaryio/dredd/issues/625.


miiila:
👍 will merge after tests get passed

Issue #629, opened 2016-09-20
Install Dredd without having Coffee globally installed 


alikh31:
Hey guys,

Is there anyway you can modify the package to run without having coffee-script globally installed, coffee-script turned to a hidden dependencies for the package.

when I try to install the package as I get the error: 

```
> dredd@2.1.0 postinstall /opt/bitnami/apps/jenkins/jenkins_home/workspace/device-model-api/alikh%2Fdredd/node_modules/dredd
> coffee scripts/print-installation-guidelines.coffee

sh: 1: coffee: not found
npm ERR! Linux 3.13.0-85-generic
npm ERR! argv "/home/tomcat/.nvm/versions/node/v4.4.4/bin/node" "/home/tomcat/.nvm/versions/node/v4.4.4/bin/npm" "install"
npm ERR! node v4.4.4
npm ERR! npm  v2.15.1
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn

npm ERR! dredd@2.1.0 postinstall: `coffee scripts/print-installation-guidelines.coffee`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the dredd@2.1.0 postinstall script 'coffee scripts/print-installation-guidelines.coffee'.
npm ERR! This is most likely a problem with the dredd package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     coffee scripts/print-installation-guidelines.coffee
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs dredd
npm ERR! Or if that isn't available, you can get their info via:
npm ERR! 
npm ERR!     npm owner ls dredd
npm ERR! There is likely additional logging output above.
```
`postinstall` doesn't seems to do much stuff, maybe using `require('coffee-script/register')` would solve the issue if you insist using coffee-script in all stand alone scripts.

honzajavorek:
@alikh31 Hi, Dredd should not require global CoffeeScript to be installed. At the time of `postinstall` local `coffee-script` package should be already available in the `node_modules`. The `package.json` scripts should look for binaries in the `node_modules/.bin` folder by default, so when given `"my-script": "coffee ...",`, then `npm run my-script` should evaluate `coffee` as `node_modules/.bin/coffee`.

I see you have npm v2.15.1, not sure whether it makes any difference. I just tried it out with npm 3.8.6 and it works.

[...]

PR #630, opened 2016-09-22
rewrite print-installation-guidelines in javascript

alikh31:
follow up issue #629 

there won't be need to install coffee-script globally in case of installing Dredd.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "dredd": "bin/dredd"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


honzajavorek:
Thanks Ali!

It seems like there's a glitch in our hooks tests when code is pushed by a non-Apiary user 😬 . I will investigate that. Meanwhile, I prepared a separate PR with your commits: #632

honzajavorek:
Hm, looks like the builds failed and even failed to report back under the PR, but for no obvious reason. Glitch in the universe?

PR #632, opened 2016-10-03
Rewrite print installation guidelines in JS

honzajavorek:
For no obvious reason, npm scripts written in CoffeeScript can cause problems. This change avoids any problems all together, since it introduces rewrite of the script, which is being run every time after installation, into a plain ES5 JavaScript.

This fixes https://github.com/apiaryio/dredd/issues/629 and supersedes https://github.com/apiaryio/dredd/pull/630.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "dredd": "bin/dredd"
   },
   "scripts": {
-    "postinstall": "coffee scripts/print-installation-guidelines.coffee",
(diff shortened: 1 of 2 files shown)


w-vi:
Possibly squash it to single commit.

honzajavorek:
@w-vi I didn't want to interfere much with the contributed commits. Leaving it as it is. Thanks!

PR #636, opened 2016-10-04
Add extensive 'verbose', 'debug', and 'silly' logging

### LINKED ISSUE #635: Add extensive logging for better debugging of Dredd
Currently, `--level=verbose` or `--level=silly` have basically zero effect on Dredd and do not help to debug any Dredd-related problems at all. The lowest level of messages Dredd now produces is more or less `info`, which contains informative messages for Dredd user, not debugging messages helpful for resolving issues with Dredd itself.

This requirement is driven by my own personal need when developing Dredd as well as by resolving issues of actual Dredd users (@philsturgeon, or @zmorris in #633). Related also to #625.

honzajavorek:
### What is this change about?

- Primarily, this PR closes #635. Please see description in the issue.
- It also prints some debugging info (Dredd version, system version, node version, ...) in the beginning, which is a prototype of https://github.com/apiaryio/dredd/issues/625. We agreed upon this with @netmilk - better ugly and now than beautiful and later.
- During works on this PR I also found out my computer is prone to https://github.com/nodejs/node/issues/6456. The Node.js issue mentions a workaround. I added this workaround to `bin/dredd` to avoid the issue for myself and any other users who might have been affected.

### Why test coverage dropped?

I believe it's because I added a lot of new lines, also to places which were uncovered previously. That made the uncovered parts statistically larger than they were.


src/add-hooks.coffee
```diff
@@ -44,11 +44,12 @@ addHooks = (runner, transactions, callback) ->
...
 
     catch error
-      logger.warn 'Skipping hook loading...'
-      logger.warn 'Error reading hook file "' + filePath + '"'
-      logger.warn 'This probably means one or more of your hookfiles is invalid.'
-      logger.warn 'Message: ' + error.message if error.message?
-      logger.warn 'Stack: ' + error.stack if error.stack?
+      logger.warn("""\
+        Skipping hook loading. Error reading hook file '#{filePath}'. \
+        This probably means one or more of your hook files is invalid.
```
miiila: NIT: `s/is/are`

src/add-hooks.coffee
```diff
@@ -87,9 +88,9 @@ addHooks = (runner, transactions, callback) ->
         loadSandboxHooksFromStrings(callback)
       else
         # not sandboxed code can't be loaded from the string
-        msg = """
-        Not sandboxed hooks loading from strings is not implemented,
-        Sandbox mode must be enabled when loading hooks from strings."
+        msg = """\
```
miiila: NIT: Unnecessary `"""`.

src/dredd-command.coffee
```diff
@@ -9,13 +9,15 @@ spawnArgs = require 'spawn-args'
 Dredd = require './dredd'
 interactiveConfig = require './interactive-config'
 configUtils = require './config-utils'
-logger = require './logger'
+logger = require('./logger')
 
 packageData = require('../package.json')
 
+
```
miiila: Is the new line necessary? Same for line 20.

src/dredd-command.coffee
```diff
@@ -54,17 +56,19 @@ class DreddCommand
...
+      return callback()
 
     term = =>
-      logger.info 'Sending SIGTERM to the backend server'
+      logger.info('Sending SIGTERM to backend server process.')
       @serverProcess.kill 'SIGTERM'
 
     kill = =>
-      logger.info 'Killing backend server'
+      logger.info('Killing backend server process.')
```
miiila: I'd keep `'Sending SIGKILL...` wording, but YMMV.

src/hooks-worker-client.coffee
```diff
@@ -29,14 +29,19 @@ class HooksWorkerClient
     @emitter = new EventEmitter
 
   start: (callback) ->
+    logger.verbose('Looking up hooks handler implementation:', @language)
     @setCommandAndCheckForExecutables (executablesError) =>
```
miiila: Unnecessary fat arrow.  Same applies to similar statements on lines below.

src/reporters/apiary-reporter.coffee
```diff
@@ -177,38 +181,34 @@ class ApiaryReporter
...
 
-      res.on 'end', =>
-        if @verbose
-          logger.log 'Rest Reporter Response ended'
+      res.on 'end', ->
+        logger.verbose('Apiary reporter Response received.')
 
         try
           parsedBody = JSON.parse buffer
         catch e
```
miiila: Is caught error ignored intentionally? Wouldn't it help for faster orientation in JSON parse error?

src/reporters/apiary-reporter.coffee
```diff
@@ -222,18 +222,14 @@ class ApiaryReporter
...
 
     unless @configuration['apiToken'] == null
       options.headers['Authentication'] = 'Token ' + @configuration['apiToken']
 
-    if @verbose
-      info =
-        options: options
-        body: body
-      logger.log 'Rest Reporter Request:', JSON.stringify(info, null, 2)
+    logger.verbose('Apiary reporter request:', JSON.stringify({options: options, body}, null, 2))
```
miiila: Looks weird with one extra key only.

src/dredd-command.coffee
honzajavorek: I usually use two lines to separate top-level module content and one line to separate everything else. The CS style guide surprised me by suggesting that after `class` there should be three new lines or something like that, so if that popped up on me, I obeyed, although I don't think 3 lines are in any way better than 2.

src/dredd-command.coffee
honzajavorek: I'm afraid many developers and users of Dredd do not know anything about signals, so I'm trying to move away from these terms, at least for user-focused `info` logging (I'm fine with such expressions in `debug` etc.). Also, I'm not sure SIGKILL will be appropriate term to use when we'll start supporting Windows, would need to check.

src/hooks-worker-client.coffee
honzajavorek: Yes. But I didn't want to snowball even more than I already did 😞 It would be probably worth to make a dedicated PR to fix some of the most outstanding linter issues in Dredd codebase.

src/reporters/apiary-reporter.coffee
honzajavorek: I don't understand your comment. But when I look at the change, `{options: options, body}` could be written as `{options, body}`.

src/reporters/apiary-reporter.coffee
honzajavorek: Good point! This could be handled in a better way. I'll address this in this PR when you finish the initial review.

src/reporters/apiary-reporter.coffee
```diff
@@ -242,21 +238,18 @@ class ApiaryReporter
       else
         return callback error, req, null
 
-    if @configuration.apiUrl?.indexOf('https') is 0
-      if @verbose
-        logger.log 'Starting REST Reporter HTTPS Request'
+    logger.verbose('Starting Apiary reporter HTTP request')
+    if @configuration.apiUrl.match(/^https/)
```
miiila: When compared with previous version `?` is missing. Please consider, whether it should be present or not.

src/reporters/apiary-reporter.coffee
```diff
@@ -242,21 +238,18 @@ class ApiaryReporter
       else
         return callback error, req, null
 
-    if @configuration.apiUrl?.indexOf('https') is 0
-      if @verbose
-        logger.log 'Starting REST Reporter HTTPS Request'
+    logger.verbose('Starting Apiary reporter HTTP request')
```
miiila: Sentence `Starting Apiary reporter HTTP request` doesn't make sense for me, but I can be missing some context.

src/reporters/apiary-reporter.coffee
```diff
@@ -242,21 +238,18 @@ class ApiaryReporter
...
-
     else
-      if @verbose
-        logger.log 'Starting REST Reporter HTTP Response'
+      logger.debug('Using unsecured HTTP for Apiary reporter request')
       req = http.request options, handleResponse
 
-      req.on 'error', handleReqError
-
+    req.on 'error', handleReqError
```
miiila: `req.on('error', handleReqError)`

src/apply-configuration.coffee
```diff
@@ -13,8 +14,23 @@ coerceToArray = (value) ->
   else
     return value
 
-applyConfiguration = (config) ->
 
+applyLoggingOptions = (options) ->
+  # coerce color to bool
+  if options.color == 'false'
+    options.color = false
+  else if options.color == 'true'
```
miiila: Only strings have special handling here. I would change comment in a way to emphasise it (e.g. `coerce string color to bool, leave others`).

src/dredd-command.coffee
```diff
@@ -8,6 +8,7 @@ spawnArgs = require 'spawn-args'
 
 Dredd = require './dredd'
 interactiveConfig = require './interactive-config'
+{applyLoggingOptions} = require './apply-configuration'
```
miiila: I am not sure with the filename `apply-configuration`, when it expose more then one function. Maybe `configuration` only would be enough?

src/dredd-command.coffee
```diff
@@ -266,7 +269,6 @@ class DreddCommand
       , waitMilis
 
   run: ->
-    logger.verbose('Preparing for start of testing.')
```
miiila: I am not sure with logging policy, so just pointing this removal for reconsideration, whether it should be really removed or not.

bin/dredd
```diff
@@ -2,6 +2,14 @@
 
 var fs = require('fs');
 
+
+// Ignore this block, it's pure magic, temporary fix
```
miiila: 🎆

src/dredd-command.coffee
```diff
@@ -281,7 +296,8 @@ class DreddCommand
       return if @finished
 
     configurationForDredd = @initConfig()
-    @dreddInstance = @initDredd configurationForDredd
+    @logDebuggingInfo(configurationForDredd)
```
miiila: Is it intentional to output the debugging info always? It seems to me with every Dredd run, debug info will be printed.

src/apply-configuration.coffee
```diff
@@ -91,8 +91,8 @@ applyConfiguration = (config) ->
   if config.blueprintPath
     configuration.options.path.push config.blueprintPath
 
-  configuration.options.method.map((method) ->
-    method.toUpperCase()
+  configuration.options.method = (
```
miiila: Shouldn't it be `configuration.options.methods`? I think name should reflect the fact it's array.

src/dredd-command.coffee
honzajavorek: This logging happens before logging level is set, so it can never make it to the output.

src/dredd-command.coffee
honzajavorek: Yes. It's poor man's implementation of #625 - with @netmilk we decided we'll just print this kind of info in the beginning of every Dredd run for now.

src/apply-configuration.coffee
honzajavorek: The thing is, the Dredd option is `--method`, but accepts also multiple values, so in configuration handling, it's coerced to array while the original singular name sticks. Not ideal and it's on multiple places in Dredd currently. Confused me too, at the first sight.

src/hooks-worker-client.coffee
miiila: Ah, I missed the indentation, sorry for that.

miiila:
@honzajavorek Could you please just briefly comment https://github.com/apiaryio/dredd/pull/636#discussion-diff-82753444R22 and https://github.com/apiaryio/dredd/pull/636#discussion-diff-82754205R11? It's neither addressed in last commit nor commented, so I don't know, if you missed it or didn't consider as important. Thanks.

honzajavorek:
@miiila Addressed everything (including those two you mentioned) and tests passed.

miiila:
@honzajavorek so LGTM!

PR #641, opened 2016-10-11
Development dependency promoted, fixes last release

honzajavorek:
Dredd version 2.2.0 cannot be installed due to this 😞 Thanks very much @XVincentX for discovering this right away! 💕 👍 🇮🇹

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "request": "^2.74.0",
     "spawn-args": "^0.2.0",
     "spawn-sync": "^1.0.15",
+    "sync-exec": "^0.6.2",
     "uri-template": "^1.0.1",
     "winston": "^2.2.0"
   },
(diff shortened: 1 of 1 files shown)


w-vi:
This has ruined my CI run , switching to stable :)

honzajavorek:
@w-vi That won't help, stable is only avoiding pre-releases 😐 I'm really not sure how to catch these kinds of errors in CI before releasing the package.

w-vi:
@honzajavorek maybe after publishing you can have a special CI run which would install dredd and run it, so at least you can catch it fast.

honzajavorek:
@w-vi I have @XVincentX for that. I need something before release. Filing an issue: https://github.com/apiaryio/dredd/issues/642

w-vi:
@honzajavorek what about build local package and then trying installing that? `rm -rf`ing node_modules so it starts clean?

PR #258, opened 2015-07-30
Keep forward slashes in URL if they're present

coagulant:
Fixes #93
Adds micro-dependency (url-join)

Issue #644, opened 2016-10-20
Integration tests of hook handlers have problems under contributors' PRs


honzajavorek:
See https://travis-ci.org/apiaryio/dredd/jobs/169183118 for https://github.com/apiaryio/dredd/pull/643 from fork https://github.com/vdaubry/dredd/:

```
$ npm run test:hooks-handlers
> dredd@0.0.0-semantically-released test:hooks-handlers /home/travis/build/apiaryio/dredd
> coffee scripts/test-hooks-handlers.coffee
Tested commit (9ab4f30edc3e91e34b00b77d32c8deb11f0a55fc) is part of the '#643' PR
Setting Git user name to 'GitHub'
Setting Git e-mail to 'noreply@github.com'
Preparing branch dependent-build/pr643/169183114/ruby-hooks-handler
Pushing dependent-build/pr643/169183114/ruby-hooks-handler to GitHub
Preparing branch dependent-build/pr643/169183114/python-hooks-handler
Pushing dependent-build/pr643/169183114/python-hooks-handler to GitHub
Preparing branch dependent-build/pr643/169183114/php-hooks-handler
Pushing dependent-build/pr643/169183114/php-hooks-handler to GitHub
Preparing branch dependent-build/pr643/169183114/perl-hooks-handler
Pushing dependent-build/pr643/169183114/perl-hooks-handler to GitHub
Preparing branch dependent-build/pr643/169183114/go-hooks-handler
Pushing dependent-build/pr643/169183114/go-hooks-handler to GitHub
```

Strange git name, git email, branches never made it to the Dredd repo. This needs to be tested and fixed.

honzajavorek:
...and this is why: https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions

honzajavorek:
Closed by https://github.com/apiaryio/dredd/pull/782, in favor of https://github.com/apiaryio/dredd/issues/784.

PR #662, opened 2016-11-15
Keep forward slashes in URL if they're present

### LINKED ISSUE #93: Forward slashes are not added
Having blueprint:
## Address Detail [/account/addresses/{id}/]
- Parameters
  - id (required, string, `123`) ... String `id` of a customer address.

But dredd is making following request (verified by server side):
/account/addresses/123
instead of
/account/addresses/123/

On log trace it shows:
fail: GET /account/addresses/123/ duration: 176ms

So that is 404 as url doesn't match, and as a solution if I enable redirect to forward slashes, then it claims for 301 and still fails.

Thank you

browniebroke:

I've got this issue as well and found an un-mergeable pull request open. This is my attempt at rebasing it.
Bare with me, it's my first PR in coffee script!

Not sure if this requires writing some doc, let me know if so.

Fixes #93
Replaces #258

- [x] To write docs
- [x] To write tests

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -354,8 +354,7 @@ class TransactionRunner
     segments = (segment.replace(/^\/|\/$/g, '') for segment in segments)
     # Keep trailing slash at the end if specified in requestPath
     # and if requestPath isn't only '/'
-    # https://github.com/apiaryio/dredd/issues/93
-    trailingSlash = (requestPath != '/' and requestPath.slice(-1) == '/' and '/' or '')
+    trailingSlash = if requestPath isnt '/' and requestPath.slice(-1) is '/' then '/' else ''
     return '/' + segments.join('/') + trailingSlash
 
   # Factory for 'transaction.test' object creation
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -352,7 +352,11 @@ class TransactionRunner
     # See also https://github.com/joyent/node/issues/2216
     segments = [serverPath, requestPath]
     segments = (segment.replace(/^\/|\/$/g, '') for segment in segments)
-    return '/' + segments.join('/')
+    # Keep trailing slash at the end if specified in requestPath
+    # and if requestPath isn't only '/'
+    # https://github.com/apiaryio/dredd/issues/93
+    trailingSlash = (requestPath != '/' and requestPath.slice(-1) == '/' and '/' or '')
+    return '/' + segments.join('/') + trailingSlash
 
   # Factory for 'transaction.test' object creation
   createTest: (transaction) ->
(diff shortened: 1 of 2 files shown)


honzajavorek:
@browniebroke Thank you very much for sending your first Pull Request here, moreover in Coffee Script! We have a public holiday here in the Czech Republic, so my answers aren't very timely, but I'll take a look next week for sure! 🎈

src/transaction-runner.coffee
```diff
@@ -352,7 +352,11 @@ class TransactionRunner
     # See also https://github.com/joyent/node/issues/2216
     segments = [serverPath, requestPath]
     segments = (segment.replace(/^\/|\/$/g, '') for segment in segments)
-    return '/' + segments.join('/')
+    # Keep trailing slash at the end if specified in requestPath
+    # and if requestPath isn't only '/'
+    # https://github.com/apiaryio/dredd/issues/93
+    trailingSlash = (requestPath != '/' and requestPath.slice(-1) == '/' and '/' or '')
```
honzajavorek: It's preferred to use `isnt` instead of `!=` and `is` instead of `==`, but it's just nitpick, it will work and it's okay.

honzajavorek: What does ` and '/' or ''` mean? I think this will just evaluate as follows:

```
requestPath != '/'
AND
requestPath.slice(-1) == '/'
AND
'/'  # always true
OR
''  # always false
```

src/transaction-runner.coffee
```diff
@@ -352,7 +352,11 @@ class TransactionRunner
     # See also https://github.com/joyent/node/issues/2216
     segments = [serverPath, requestPath]
     segments = (segment.replace(/^\/|\/$/g, '') for segment in segments)
-    return '/' + segments.join('/')
+    # Keep trailing slash at the end if specified in requestPath
+    # and if requestPath isn't only '/'
+    # https://github.com/apiaryio/dredd/issues/93
```
honzajavorek: I think it's not needed to link the issue, the comment is explanatory enough and the issue will be fixed by the PR.

honzajavorek:
@browniebroke Very good! Thanks very much for the PR. Please ignore the pending hooks tests, they are broken for contributed PRs (https://github.com/apiaryio/dredd/issues/644).

src/transaction-runner.coffee
browniebroke: For the `and '/' or ''`, that was my attempt at doing a one-liner for:
```coffeescript
if requestPath isnt '/' and requestPath.splice(-1) is '/'
	trailingSlash = '/'
else
	trailingSlash = ''
```

After digging a bit I found out that it's usually done with:
```coffeescript
trailingSlash = if requestPath isnt '/' and requestPath.slice(-1) is '/' then '/' else ''
```

Thanks for your comment.

PR #664, opened 2016-11-23
Replace deprecated uuid module

AlexTes:
Author of node-uuid decided to deprecate for reasons I do not know and
replace it with uuid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "inquirer": "^1.1.0",
     "js-yaml": "^3.6.1",
     "markdown-it": "^8.0.1",
-    "node-uuid": "^1.4.7",
     "optimist": "^0.6.1",
     "pitboss-ng": "^0.3.2",
     "proxyquire": "^1.7.10",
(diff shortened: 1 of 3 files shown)


AlexTes:
I think I missed some steps from CONTRIBUTING. I'll take a look later today.

honzajavorek:
@AlexTes Thanks for the Pull Request! I checked both projects:

### node-uuid
- https://www.npmjs.com/package/node-uuid
- https://github.com/broofa/node-uuid
- I see no deprecation notice
- Last change 5 days ago
- Is fork of https://github.com/kelektiv/node-uuid according to GitHub

### uuid
- https://www.npmjs.com/package/uuid
- https://github.com/kelektiv/node-uuid
- I see no deprecation notice
- Last change 2 days ago
- Is upstream according to GitHub

----------------

I'm a bit confused about which one of those two is preferred and actively developed as of now 😕

AlexTes:
You can read up on it here https://github.com/kelektiv/node-uuid/issues/142

Broofa the original maintainer flagged he did not have the time to maintain. The repo was handed off to the current owner Kelektiv. Maintainer of the fork is in agreement.

I can list a couple of practical reasons but if you want more re-assuring I think the guys involved will do a better job than I!

Keep in mind since the maintainer of `node-uuid` has flagged it deprecated every user of this module will get an npm warning explaining they should no longer use `node-uuid` but use `uuid` instead.

honzajavorek:
@AlexTes Thanks for linking the issue and the explanation! Both helped me to get an idea on what's going on there.

Issue #204, opened 2015-06-01
Run Dredd's test suite on Windows


netmilk:
Support for Windows with [AppVeyor](http://www.appveyor.com/) CI as mentioned in #126

netmilk:
Interesting feedback is here: https://github.com/apiaryio/dredd/issues/126#issuecomment-125994574

honzajavorek:
I created [Windows label](https://github.com/apiaryio/dredd/issues?utf8=%E2%9C%93&q=label%3AWindows+) which lists all related issues.

honzajavorek:
Status:

### Open PRs

- [ ] https://github.com/apiaryio/dredd/pull/732 needs to be merged
- [ ] https://github.com/apiaryio/dredd/pull/730 needs to be merged

### Open, but failing PRs

- [ ] https://github.com/apiaryio/dredd/pull/734 needs to be polished and merged
- [ ] https://github.com/apiaryio/dredd/pull/733 needs to be polished and merged

### Subsequent works

- New branch `honzajavorek/consolidate-cli-tests-and-helpers` should be created with
    - 97ce68b
    - fea595d
    - 46c81b6
    - 150a4cc
    - bf7cbbd
    - 474d0a4
    - possibly also c91f70c8b76949105e3aacb9c1796443043f01c5

- New branch for scripts:
    - 80d371591676c6a4b84b9e9d932fd018b2b85597
    - 10af1dbd713ad54b01a43b4dd69ee108eedfe4a9

- New branch to introduce AppVeyor:
    - 2253019
    - b433f04fc162289e7a5f0a61e4e6f57ff207732d
    - e102c1b2c6d71f09dd1b73d7570ec2efdc55d98a
    - merge 03428ac527e993a41d9cbb7bea8b120587e9fdfe and 03428ac527e993a41d9cbb7bea8b120587e9fdfe somehow (mostly parts skipping some tests on Windows should be missing) and incorporate the missing changes

- Delete branches `honzajavorek/appveyor` and `honzajavorek/appveyor-badge`

PR #667, opened 2016-11-25
Make spawning server command and 'which' Windows-compatible

honzajavorek:

I want to add an experimental AppVeyor (Windows-based) CI builds to [dredd-example](https://github.com/apiaryio/dredd-example/). This PR is a small step towards better Windows support.

- **The `--server` option should now work also on Windows.** I manually verified that this change is an improvement and enables running the server command on Windows.
- **Removed direct subprocess call to `which` command, which isn't present on Windows.** This is an obvious blocker for non-JS hooks being supported on Windows and I just wanted to have it out of our way. I didn't verify if it actually enables anything right away.

- #204

### DIFF
--- a/test/unit/hooks-worker-client-test.coffee
+++ b/test/unit/hooks-worker-client-test.coffee
@@ -5,9 +5,9 @@ net = require 'net'
 {assert} = require 'chai'
 clone = require 'clone'
 
-childProcessStub = require 'child_process'
-loggerStub = require '../../src/logger'
-whichStub =  require '../../src/which'
+crossSpawnStub = require('cross-spawn')
+whichStub = require('../../src/which')
+loggerStub = require('../../src/logger')
 
 Hooks = require '../../src/hooks'
 commandLineOptions = require '../../src/options'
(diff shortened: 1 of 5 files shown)


test/unit/hooks-worker-client-test.coffee
```diff
@@ -172,21 +183,9 @@ describe 'Hooks worker client', ->
 
           hooksWorkerClient.stop (err) ->
             assert.isUndefined err
-            assert.equal childProcessStub.spawn.getCall(0).args[1][0], 'somefile.rb'
+            assert.equal crossSpawnStub.spawn.getCall(0).args[1][0], 'somefile.rb'
             done()
 
-    describe 'when --language nodejs option is given', ->
```
honzajavorek: I've just moved this block above so it's not between two Ruby hooks tests.

netmilk:
@honzajavorek If we're talking Windows support - did you consider hooking up Appveyor to make sure it's actually working in some time?

honzajavorek:
@netmilk I'd love to, but Dredd's test suite isn't ready to run there. I want to hook up AppVeyor to `dredd-example` with big fat "experimental" warning so we have at least little proof Dredd installs on Windows and is able to test simple Blueprint/Swagger with JavaScript hooks.

I want to add related docs (Windows and installation in general) in subsequent PR, later today.

To run the whole Dredd test suite on AppVeyor would be awesome, but needs more time.

w-vi:
LGTM

Issue #660, opened 2016-11-08
Drop support for 0.10 and 0.12 at the end of 2016


honzajavorek:
According to [this table](https://github.com/nodejs/LTS#lts-schedule), Node.js 0.10 and 0.12 will both be officially dead on 2016-12-31. We should celebrate New Year's Eve by removing those two Node versions from our build matrix 🎊 🍸 

### Checklist

- [x] Node 0.10 and 0.12 are removed from build matrix
- [x] `npm run lint` isn't limited to Node 6 anymore ( ca689f9 )
- [x] Update `nock` devDependency in Dredd (they dropped support for 0.10, 0.12 and 4)
- [x] Node 0.10 and 0.12 are removed also from dependencies, such as...
  - [x] Gavel.js
  - [x] Gavel.js ➡️ 0.10 doesn't need to be excluded [here](https://github.com/apiaryio/gavel.js/blob/master/.travis.yml#L19)
  - [x] Dredd Transactions
- [x] Change `engines` field in `package.json` in Dredd and all related dependencies
  - [x] Gavel.js
  - [x] Dredd Transactions
- [x] Make sure Dredd hooks implementations will install Dredd on newer Node in the CI
  - [x] Python
  - [x] Ruby
  - [x] PHP
  - [x] Go
  - [x] Perl
  - [x] Dredd Hooks Template's Travis CI

PR #671, opened 2016-12-02
Updating dependencies

honzajavorek:
Usual chore.

- See [this](https://github.com/semantic-release/semantic-release/releases/tag/v6.3.2), `travis_after_all.py` should not be needed anymore.
- I've documented supported engines (so I could change it in the future when https://github.com/apiaryio/dredd/issues/660 will happen)
- I **did not** update `nock`, because it dropped support for 0.10/0.12/4. I'll do it later and I noted it in the https://github.com/apiaryio/dredd/issues/660.

The `continuous-integration/travis-ci/pr` tests are failing because of https://github.com/apiaryio/dredd/issues/672. I can't do much about it at the moment.


miiila:
🎱

Issue #528, opened 2016-06-14
Dredd's CI tests should check if the package is installable


honzajavorek:
Use `npm pack` to check whether the resulting package is installable. Dredd is transpiled package, uses `.npmignore` and we don't want to repeat `v1.1.0-pre.6`/`v1.1.0-pre.7` failure.

See also https://github.com/apiaryio/dredd/issues/642 for other ideas.

honzajavorek:
Proposal on solution. Script, which would:

1. Get contents of `master` branch of [dredd-example](https://github.com/apiaryio/dredd-example) with current dredd code in special folder called `./__dredd__`, the same way [`test-hooks-handlers.coffee`](https://github.com/apiaryio/dredd/blob/master/scripts/test-hooks-handlers.coffee) does it.
2. Remove any files in `./__dredd__/` which are ignored by git. (Or just make somehow sure that only git-visible files are copied to `./__dredd__`).
3. Use [`npm prune --production`](https://docs.npmjs.com/cli/prune) in `./__dredd__` folder.
4. Use [`npm pack`](https://docs.npmjs.com/cli/pack) to pack the `./__dredd__` folder as tarball.
5. Edit dredd-example's `package.json` so it uses Dredd from the tarball (probably [this](https://docs.npmjs.com/files/package.json#local-paths) or [`npm install <tarball file>`](https://docs.npmjs.com/cli/install).
6. Install & run test suite of dredd-example. Check for result.

The script would run either as the last part of any build or as a separate "service", similar to hook handler integration tests.

PR #673, opened 2016-12-05
Update README.md

henricavalcante:

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,12 @@ Dredd supports writing [hooks](https://dredd.readthedocs.io/en/latest/hooks/)
 $ npm install -g dredd
 ```
 
+or 
+
+```
+$ yarn global add dredd
+```
+ 
 ## Quick Start
 
 1.  Create an [API Blueprint][] file called `api-description.apib`.


honzajavorek:
@henricavalcante Hi, thanks for the suggestion, but this won't be so easy, I believe.

1. **Verifiability.** Did you test Dredd is actually installable by yarn? I did not. Promoting the option means the project actually declares it's installable by yarn and if that's not the case, it's a bug. There are already sometimes problems to keep this promise for npm (see https://github.com/apiaryio/dredd/issues/528). I'm far from adding yarn before we can even rely Dredd is always installable by npm (thus before we have https://github.com/apiaryio/dredd/issues/528 implemented).
2. **Changes not sufficient.** There is more related documentation and you didn't make the change everywhere, especially in the Installation Guide in the docs.
3. **Docs usability.** `npm` is a current standard for installing Node.js packages. So far `yarn`'s audience is early adopters or people who seek for optimization - let's call them power users of the JavaScript ecosystem. Dredd aims to be installable by anyone, even people not very skilled in the JS ecosystem. I believe anyone using `yarn` already knows they can install any npm package they wish and seeing `npm install -g dredd` in docs is for them easy to translate into respective `yarn` command. On the other hand, Node.js newbie gets easily confused by the fact there are multiple options. For that matter I would not suggest adding `yarn` like this.

If you attach some reasoning around this or in some way address my comments, I'm fine with further discussing this and re-opening, but for now, I'm closing this, I'm sorry.

PR #688, opened 2017-01-10
Note on security - Apiary Reporter

honzajavorek:

We should be transparent about this.

- [x] To write docs


honzajavorek:
@netmilk What about now?

Issue #574, opened 2016-07-22
Dredd's test suite won't pass on Travis' non-sudo architecture


honzajavorek:
Now with JS parser (#573), we could remove the requirement for sudo in our CI builds and get Travis' more performant container-based architecture.

However, I tried to do it when implementing #573 and it seems like its own different can of worms, so I sticked with the old option.

It would be nice to resolve any issues (exit codes, shell, ...) and switch to non-sudo architecture. The same thing could also be beneficial in case we're ever wanting to run the test suite in Docker containers, since I believe the issues would be similar.

Issue #422, opened 2016-03-18
Docker for development


honzajavorek:
Dredd should provide basic Docker support for development.
- It might remove dependency on specific CI (currently Travis CI) and Dredd's test suite could be easily ran anywhere.
  - https://docs.travis-ci.com/user/docker/
  - https://circleci.com/docs/docker
- It would allow anyone to easily run tests and contribute. We could remove [Virtual Development Environment](https://github.com/apiaryio/dredd/blob/master/VirtualDevelopmentEnvironment.md) based on Vagrant.

honzajavorek:
More ideas:

- Dev-focused image should come with pre-linked libraries, such as dredd-transactions, fury adapters, etc.
- The dev-focused image should use [`.gitmessages`](https://robots.thoughtbot.com/better-commit-messages-with-a-gitmessage-template) and `commitizen` to help people with Conventional Changelog as much as possible

honzajavorek:
Since we're starting to support Windows now, I think it would be nice to have a Windows Docker container or a way how to quickly provision a Windows machine to develop and test Dredd.

honzajavorek:
https://github.com/apiaryio/dredd-docker/pull/1 will bring user-oriented Docker image.

I'm not sure whether Dredd _developers_ struggle to install Dredd that much and if so, I'd rather pursue the idea of having a monorepo #704, where everything is handled within one repository instead of reinventing the wheel by `npm link`-ing separate projects inside a Docker image. Also, Dredd supports Windows and it seems it's not so easy to have this all handled by Docker https://github.com/apiaryio/dredd/issues/1093#issuecomment-427020850.

While I'm adding user-oriented Docker image now, the developer-oriented is no priority for me as of now. However, I'm happy to accept contributions for such Docker image.

honzajavorek:
Monorepo with Lerna in place is the way to go. Closing.

PR #689, opened 2017-01-10
Use Travis CI without sudo

honzajavorek:

Travis CI offers two kinds of runtimes. Simplified,

- one is distro-based allows you to use `sudo` and install all kind of stuff on your own,
- second is container-based, has more limitations, but should be newer and faster.

Dredd did need `sudo` previously, because of C++11 compilation of the [drafter](https://github.com/apiaryio/drafter). Since drafter is now also JS-only, Dredd does not need the distro-based architecture anymore. This PR removes `sudo` and all related options from the Travis CI configuration. The PR also adds `--no-optional` to `npm install`, which forces npm to ignore C++11 compilation of drafter and to use the JS version directly. That should improve build time.

I did try to migrate Dredd's test suite to non-sudo builds previously already, but the test suite did not pass and grappled with the new/simpler/different environment. A lot of stuff was failing without obvious reasons. I filed https://github.com/apiaryio/dredd/issues/574 and postponed the works. Now I'm back to find out that Travis CI probably made changes to the new container-based architecture and there are no changes needed in the test suite at all. It passed at first try, no failures. Curious!

- Closes https://github.com/apiaryio/dredd/issues/574
- Possibly related to #422 and/or #204

- [x] To write docs - there were no docs related to this change
- [x] To write tests - :trollface: 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-sudo: "required"
-dist: "trusty"
 language: "node_js"
 node_js:
   - "0.10"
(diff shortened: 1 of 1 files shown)


miiila:
(➕row + ➖ rows < 0️⃣ ) == True, therefore 👍 :trollface:

Issue #705, opened 2017-01-25
Decaffeinate the codebase


honzajavorek:
[Decaffeinate](https://github.com/decaffeinate/decaffeinate) dredd and [dredd-transactions](https://github.com/apiaryio/dredd-transactions). CoffeeScript did great things for the JavaScript ecosystem and it will be in our hearts forever, together with good memories.

However, the times has changed. ES2015 rules the world now. [Apiary](https://github.com/apiaryio) decided to adopt latest ECMAScript versions for new projects. I've written our stands [here](https://github.com/coffeescript6/discuss/issues/32#issuecomment-246619091). There is no official push to have Dredd as JavaScript, but I think it's better for its future.

For a project to be viable Open Source in the JavaScript world of today it's crucial not to repel contributors by using a language, which is provably fading away. Therefore we should try to eventually migrate the codebases towards JavaScript.

honzajavorek:
This should be linked from contributing docs.

honzajavorek:
Thanks to @michalholasek we are JavaScript now!

<img width="321" alt="image" src="https://user-images.githubusercontent.com/283441/35810410-ed56a604-0a8b-11e8-98a6-dd15e46f5ab2.png">

PR #712, opened 2017-02-08
Dumb implementation for support --port and --host for hooks server

ddelnano:

To allow compliant hooks servers to listen on different ports and ip addresses

https://github.com/apiaryio/dredd-hooks-template/issues/11
https://github.com/apiaryio/dredd-hooks-template/pull/17

Still very much WIP since I believe there will need to be a discussion on what flags the hooks servers should look for (php already uses `--port` and `--host` but this can be up for discussion), how to stage these changes amongst the projects, etc.  I just wanted to get the simplest thing done so that I could run the dredd-hooks-php cucumber tests with [this](https://github.com/apiaryio/dredd-hooks-template/pull/17) new test to validate it works properly.

### DIFF
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -163,10 +163,10 @@ class HooksWorkerClient
         callback()
 
   spawnHandler: (callback) ->
-    pathGlobs = [].concat ["--port", "#{@handlerPort}", "--host", "#{@handlerHost}"], @runner.hooks?.configuration?.options?.hookfiles
+    args = [].concat ["--port", "#{@handlerPort}", "--host", "#{@handlerHost}"], @runner.hooks?.configuration?.options?.hookfiles
 
-    logger.info("Spawning `#{@language}` hooks handler process with command `#{@handlerCommand} #{pathGlobs.join ","}`.")
-    @handler = crossSpawn.spawn @handlerCommand, pathGlobs
+    logger.info("Spawning `#{@language}` hooks handler process with command `#{@handlerCommand} #{args.join ","}`.")
+    @handler = crossSpawn.spawn @handlerCommand, args
 
     @handler.stdout.on 'data', (data) ->
       logger.info("Hooks handler stdout:", data.toString())
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -164,7 +164,6 @@ class HooksWorkerClient
 
   spawnHandler: (callback) ->
     pathGlobs = [].concat ["--port", "#{@handlerPort}", "--host", "#{@handlerHost}"], @runner.hooks?.configuration?.options?.hookfiles
-    console.info pathGlobs
 
     logger.info("Spawning `#{@language}` hooks handler process with command `#{@handlerCommand} #{pathGlobs.join ","}`.")
     @handler = crossSpawn.spawn @handlerCommand, pathGlobs
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -163,7 +163,7 @@ class HooksWorkerClient
         callback()
 
   spawnHandler: (callback) ->
-    args = [].concat ["--port", "#{@handlerPort}", "--host", "#{@handlerHost}"], @runner.hooks?.configuration?.options?.hookfiles
+    args = [].concat ["--port=#{@handlerPort}","--host=#{@handlerHost}"], @runner.hooks?.configuration?.options?.hookfiles
 
     logger.info("Spawning `#{@language}` hooks handler process with command `#{@handlerCommand} #{args.join ","}`.")
     @handler = crossSpawn.spawn @handlerCommand, args
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -163,9 +163,10 @@ class HooksWorkerClient
         callback()
 
   spawnHandler: (callback) ->
-    pathGlobs = [].concat @runner.hooks?.configuration?.options?.hookfiles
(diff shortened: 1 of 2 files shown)


src/hooks-worker-client.coffee
```diff
@@ -163,9 +163,10 @@ class HooksWorkerClient
         callback()
 
   spawnHandler: (callback) ->
-    pathGlobs = [].concat @runner.hooks?.configuration?.options?.hookfiles
+    pathGlobs = [].concat ["--port", "#{@handlerPort}", "--host", "#{@handlerHost}"], @runner.hooks?.configuration?.options?.hookfiles
```
honzajavorek: Also, lately I tend to prefer to have all Dredd options in the `--option=value` format, consistently. In the case of Dredd itself, both formats should work, I just wanted to choose one way to go and stick with it.

src/hooks-worker-client.coffee
```diff
@@ -163,9 +163,10 @@ class HooksWorkerClient
...
+    console.info pathGlobs
```
ddelnano: Yep, removed it.

src/hooks-worker-client.coffee
ddelnano: Updated both of these.

ddelnano:
## Todo
- [ ] Find out impact for existing hook servers
- [ ] Finish discussion on if this is what we want.  I like the idea of Dredd being able to discover if hooks support these flags at runtime but this is a much bigger change.

ungrim97:
Will require an update to support the port/host arguments. Passing them in to the current version shouldn't break anything. (it will look for hookfiles files called --port=1234 and --host=127.0.0.1 but obviously won't find one and will skip them)

Will make an update to accept the new params

ungrim97:
https://github.com/ungrim97/Dredd-Hooks/pull/7 should cover whats needed to make the Perl version work for this

w-vi:
I'll check Python hooks and see if I can implement this feature quickly.

ddelnano:
@honzajavorek @w-vi @ungrim97 so looks like all existing hooks servers will not be affected by this change except the Go ones.  How do we want to handle that?  I can get those updated soon but wanted to understand better what the plan was.

w-vi:
Python hooks won't do anything bad, will skip the args so no issue here.

ddelnano:
But what I mean is do we want to continue down this path?  It's going to cause existing go hook servers problems.  All other hooks seem fine but still.

honzajavorek:
I'm sorry for not being prompt in answering (got sick). @ddelnano thanks very much for checking all the hooks! This is amazing work.

I see three options how to proceed:

- **Let the older Go hooks installations crash.** I don't like it very much, but I can't decide, since I have very little insight into how many people use the Go hooks, how prompt they are in upgrading to newer versions, etc. Probably up to @ddelnano and @snikch to consider.
- **The `features` event mentioned by @w-vi in https://github.com/apiaryio/dredd-hooks-template/pull/17#issuecomment-279988469** I like the idea, but implementing it would involve a lot of coding - both on hooks and Dredd sides. I'd prefer to have this as a future improvement instead of an instant solution.
- **Hardcoded check.** Is there a way we could hard code a simple check for Go hooks version? It's the least visually appealing solution, but doesn't involve much coding, doesn't make anything crash, and gets the job done.

What do you think?

ddelnano:
@honzajavorek I definitely want to avoid option 1 :).  And since the `features` event seems like alot more work than what we need for now some type of hard coded check will be necessary.  I am going to brainstorm some ideas the rest of today and post back some potential ideas.

ddelnano:
@honzajavorek sorry for the long delay, had an unexpected hospital trip but I'm doing much better.

Another way we could check the hooks server that I thought of that would be very dumb but get the job done.  We could have dredd try and run the hooks server with the `--host` and `--port` arguments and if it fails try it without the `--host` and `--port` flags.  The drawback is that unrelated failures will probably fail twice.  For the hardcoded version check on the Go hooks how would we accomplish that?  There is no way to check the version using the go hook binary.

honzajavorek:
@ddelnano Whoa! 🚑 Whatever it was, I wish you to get better quickly!

The fact there is no way we can hardcode a check for go hooks version is unfortunate. I'll think about the trial & error approach. At first sight, I don't like it very much, but at the same time, it can be the only way we're left with to save the situation 🙂

ddelnano:
@honzajavorek thanks, things are back to normal now :).

One thing I didn't think of for the hardcoded check.  In the next version of the go hooks we could add a command to output the version and then if it doesn't support the version command we can assume that it won't support the `--host` and `--port`.

honzajavorek:
@ddelnano What is the current behavior of go hooks if they're provided with `--version`? Do they start and wait for connection, do they crash, ...?

ddelnano:
It's going to crash...

honzajavorek:
Then I think we're okay with what you proposed in https://github.com/apiaryio/dredd/pull/712#issuecomment-287864558. Dredd could try to run the hook handler with options and if it crashes, it tries to run it without the options. If it doesn't crash, it respects the test results.

ddelnano:
Ok. Yea obviously not ideal but there aren't many options unfortunately. This weekend I will have some time I'll give that a shot. And work on getting the go hooks updated to use the new flags.

honzajavorek:
I'm very sorry we were not able to get back to this. I need to close this as we're going to work on #705 first, which requires having no pending Pull Requests. We'll get back to this proposal as soon as we're addressing #917. Finishing this PR is listed on #917 as one of the necessary steps, so it shouldn't get forgotten. Thank you for the contribution! I hope we'll have the chance to tidy up the template soon.

PR #713, opened 2017-02-10
Prefer --option=value (GNU-style) format of long options

honzajavorek:

I prefer to stick to UNIX / GNU conventions. This style of options is also less error-prone when trying to test Dredd on Windows.

- #204 

- [x] To write docs


honzajavorek:

PR #714, opened 2017-02-10
Use home directory

honzajavorek:

The test checks a situation when Dredd is trying to read an existing path, but the path isn't a file, it is a directory. The test expects an exception will occur.

This change ensures the target is existing directory and that it is on a file system, which is native to the underlying operation system. E.g. when trying to read a directory as a file on NFS on Windows (can happen easily in VirtualBox, for instance), the operation won't throw an exception, but will read an empty string.

- #204 

- [x] To write tests


honzajavorek:

PR #715, opened 2017-02-10
Use cross-spawn instead of exec

honzajavorek:

This seems to be the only sane way to spawn subprocess on both Windows and UNIX in Node.js.

- #204 


honzajavorek:

PR #716, opened 2017-02-10
Drop support for Node 0.10 and 0.12

honzajavorek:

Node 0.10 and 0.12 are not supported anymore, so we're removing them from CI and we're removing any related workarounds.

- See https://github.com/apiaryio/dredd/issues/660 for the list of workarounds which this PR fixes.
- See https://github.com/nodejs/LTS#lts-schedule for details on why Node.js 0.10 and 0.12 can be both considered as officially dead since 2016-12-31.
- This is a breaking change, since we're dropping support of certain runtimes. Merging this PR should release **Dredd 3.0.0**.

Waits for https://github.com/apiaryio/dredd-transactions/pull/69 and https://github.com/apiaryio/gavel.js/pull/86, closes https://github.com/apiaryio/dredd/issues/660.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "dredd": "bin/dredd"
   },
   "engines": {
-    "node": ">= 0.10"
+    "node": ">= 4"
   },
   "scripts": {
     "lint": "conventional-changelog-lint --from=master && coffeelint src",
@@ -34,9 +34,9 @@
     "coffee-script": "^1.10.0",
     "colors": "^1.1.2",
     "cross-spawn": "^5.0.1",
-    "dredd-transactions": "^3.0.1",
+    "dredd-transactions": "^4.0.0",
     "file": "^0.2.2",
-    "gavel": "^0.5.3",
+    "gavel": "^1.0.0",
     "glob": "^7.0.5",
     "html": "^1.0.0",
     "htmlencode": "0.0.4",
@@ -69,7 +69,7 @@
     "lcov-result-merger": "^1.2.0",
     "mocha": "^3.0.0",
     "mocha-lcov-reporter": "^1.2.0",
-    "nock": "^8.0.0",
+    "nock": "^9.0.4",
     "semantic-release": "^6.3.2",
     "sinon": "^1.17.4"
   },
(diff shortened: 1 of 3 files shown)


ddelnano:
@honzajavorek we should make sure the hook servers us node > 0.12 in their travis jobs.  I just had both the goodman and php hooks repo break on the way it was installing dredd previously.

honzajavorek:
@ddelnano That's very good point! Are you sure Travis' default Node version is less than 4? I found this: https://github.com/travis-ci/travis-ci/issues/5218

ddelnano:
@honzajavorek doesn't look like it on the latest dredd-hooks-php build that I had to fix a few days ago.

screen shot 2017-02-10 at 10 31 56 am

ddelnano:
Here is the link for reference https://travis-ci.org/ddelnano/dredd-hooks-php/jobs/199693011

honzajavorek:
I found [this](http://austinpray.com/ops/2015/09/20/change-travis-node-version.html), but to me it seems to be overly complicated. Maybe

```yaml
before_install:
  - "nvm install node && nvm use node && node --version"
```

would be sufficient. `node` is an alias for the latest version of Node.

package.json
```diff
@@ -69,7 +69,7 @@
     "lcov-result-merger": "^1.2.0",
     "mocha": "^3.0.0",
     "mocha-lcov-reporter": "^1.2.0",
-    "nock": "^8.0.0",
+    "nock": "^9.0.4",
```
honzajavorek: Nock could not be updated until now, because they dropped 0.10 and 0.12 support in their new versions. See https://github.com/apiaryio/dredd/issues/660.

provisioning.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
```
honzajavorek: This file isn't actively used anywhere and should be deprecated by #422 in the future, but since it's still in the repo, I've updated it.

ddelnano:
@honzajavorek anything else that needs updated before this can be merged?

honzajavorek:
I think the only missing part is https://github.com/apiaryio/dredd-hooks-ruby/pull/32 now.

honzajavorek:
@miiila @w-vi I requested a review from both of you, but one review from either of you will do the job. So who comes first, wins 🥇

w-vi:
@hj @miiila done

 :medal_sports:  :trophy:

Issue #721, opened 2017-02-24
Error connecting to server under test when server on localhost


karpilin:
#### Describe your problem
Each test is failing with the following error:
```
error: Error connecting to server under test!
````

Despite the api endpoint is set to `http://localhost:8000` , the api calls are sent to `http://127.0.0.1:8000`.

Running a web server on `127.0.0.1` instead of `localhost` solves the problem.

#### What command line options do you use?

```
$ dredd
```

#### What is in your `dredd.yml`?

```yaml
reporter: apiary
custom:
  apiaryApiKey: *******
  apiaryApiName: ********
dry-run: null
hookfiles: './hooks.php'
language: 'bin/dredd-hooks-php'
sandbox: false
server:
server-wait: 3
init: false
names: false
only: []
output: []
header: []
sorted: false
user: null
inline-errors: false
details: false
method: []
color: true
level: info
timestamp: false
silent: false
path: []
hooks-worker-timeout: 5000
hooks-worker-connect-timeout: 1500
hooks-worker-connect-retry: 500
hooks-worker-after-connect-wait: 100
hooks-worker-term-timeout: 5000
hooks-worker-term-retry: 500
hooks-worker-handler-host: localhost
hooks-worker-handler-port: 61321
config: ./dredd.yml
blueprint: api.apib
endpoint: 'http://localhost:8000/api/v1'
```

#### What's your `dredd --version` output?

```
dredd v2.2.5 (Darwin 15.6.0; x64)
```

#### Does `dredd --level=debug` uncover something?

```
debug: Evaluating results of transaction execution #1: Authorisation > Logout > Logout
debug: No emission of test data to reporters { Error: connect ECONNREFUSED 127.0.0.1:8000
    at Object.exports._errnoException (util.js:1028:11)
    at exports._exceptionWithHostPort (util.js:1051:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)

  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 8000 }
```

honzajavorek:
Thanks for filing this! I've bumped into a similar problem when [working on Windows support](https://github.com/apiaryio/dredd/tree/honzajavorek/appveyor), but it's clear from your logs that you're using macOS. This needs tests and further investigation. Are you sure your computer correctly resolves `localhost` to an IP address?

honzajavorek:
[...]

PR #732, opened 2017-03-03
Use 127.0.0.1 instead of localhost

honzajavorek:

Using 'localhost' can provably cause problems on Windows in some cases. Replacing 'localhost' with '127.0.0.1' avoids these problems. The change is made everywhere to be consistent and, especially in docs, to present just one way to do things to Dredd's users.

Also, `localhost` can be missing in people's `hosts` files or it can be directed to a different IP address for some reason, so it might be better not rely on the alias to always work. For that reason the PR changes also default values for some Dredd options.

- Related to https://github.com/apiaryio/dredd/issues/721
- Part of #204

- [ ] To write docs - N/A
- [ ] To write tests - N/A

### DIFF
--- a/docs/data-structures.md
+++ b/docs/data-structures.md
@@ -15,7 +15,7 @@ Transaction object is passed as a first argument to [hook functions](hooks.md) a
     - resourceName: `Hello, world!` (string)
     - actionName: `Retrieve Message` (string)
     - exampleName: `Example 2` (string)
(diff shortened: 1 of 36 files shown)


honzajavorek:
@miiila @w-vi I marked you both as reviewers, but it's a race again - one review is sufficient from whoever comes first 🙂 🏁

PR #733, opened 2017-03-03
Parse hook handler command string, lookup only the command part

honzajavorek:

If hook handler command string is 'node hello.js --option=1', we want to find out only whether the 'node' executable is present in the system. For that reason, we first parse the command and lookup (using 'which') only the actual command part, without parameters. Looking up the whole string may or may not work on UNIX/Linux, but it does not work at all on Windows.

The PR also contains several minor improvements:

- Improved logging (wording, information) in `DreddCommand` class
- Using spawn instead of exec for dynamically determining the package version (debugging info in `DreddCommand` class)

#### 😱 The coverage has dropped! Why?!

The coverage has dropped because there is one more relevant line of code in part of Dredd, which is historically uncovered by tests. It's the part which takes care of killing server/hook handler processes and I plan to separate it and rewrite it in the future, so I don't consider the coverage change to be important in this case.

- #204

- [ ] To write docs - N/A
- [x] To write tests

### DIFF
--- a/src/dredd-command.coffee
+++ b/src/dredd-command.coffee
@@ -1,11 +1,10 @@
 path = require 'path'
 optimist = require 'optimist'
-console = require 'console'
 fs = require 'fs'
 os = require 'os'
 spawnArgs = require 'spawn-args'
-{spawn} = require('cross-spawn')
-execSync = require('sync-exec')
+spawn = require('cross-spawn')
+console = require('console') # stubbed in tests by proxyquire
 
 Dredd = require './dredd'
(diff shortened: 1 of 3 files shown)


honzajavorek:
@miiila @w-vi I marked you both as reviewers, but it's a race again - one review is sufficient from whoever comes first 🙂 🏁 The drop of coverage is explained in the PR description. If the PR is okay and you agree with my reasoning about the coverage, feel free to use your admin rights to force the merge even though a required status didn't pass.

Issue #591, opened 2016-08-08
Is there  a way to use proxy for dredd?


alex-glv:
I want to be able to configure proxy for dredd to be able to reach my endpoint.
Is there currently a way to do it or should I set an nginx service to proxy requests?

honzajavorek:
I don't think there's any way, currently, but maybe @netmilk knows about something.

alex-glv:
Thanks, @honzajavorek 

Apparently, the difference between proxied and non-proxied requist is the destination host and HTTP request.

Say, your proxy is at ```http://proxy``` and your API endpoint is at ```http://api```.

To proxy the request, http request should look s follows:
```
GET http://api/v1/endpoint
Host: http://proxy
[...]
```

The requests will go through the proxy server.

So, finally to use proxy from dredd one has to insert a hook into e.g hooks.js file:
```
hooks.beforeEach((transaction, done) => {
    transaction.fullPath = "http://api" + transaction.fullPath;
    [...]
});
```

And issue Dredd command with the  proxy as the destination:
```
dredd swagger-spec.json http://proxy  -f hooks.js
```

honzajavorek:
Hmm, this could go to docs 📖

honzajavorek:
@alex-glv Closing this as a duplicate of https://github.com/apiaryio/dredd/issues/51. I won't put your example to docs, because I'm going to work on a proxy support in Dredd 🙂 I'd appreciate if you watch out for new versions of Dredd and once the feature is released, if you could let me know it solves your problem.

honzajavorek:
Support for HTTP(S) proxy has been implemented in Dredd [v3.3.0](https://github.com/apiaryio/dredd/releases/tag/v3.3.0). See [documentation](https://dredd.readthedocs.io/en/latest/how-it-works/#using-https-proxy) on how to use it.

Issue #51, opened 2014-03-04
Support for http(s)_proxy


ChrisWeiss:
it doesn't appear that dredd leverages http_proxy or https_proxy when doing it's web calls to the API endpoint.

honzajavorek:
Related: https://github.com/apiaryio/dredd/issues/591

weejames:
It would be great to see this made available. The lack of this makes Dredd difficult to use behind our company proxy.

honzajavorek:
This is going to be implemented 🎉 I'd appreciate all of you commenting here (and adding @alex-glv from #591) - do you think [this kind of support](https://github.com/request/request#proxies) would be sufficient to fulfill your dreams? 🙂

honzajavorek:
Support for HTTP(S) proxy has been implemented in Dredd [v3.3.0](https://github.com/apiaryio/dredd/releases/tag/v3.3.0). See [documentation](https://dredd.readthedocs.io/en/latest/how-it-works/#using-https-proxy) on how to use it.

PR #734, opened 2017-03-03
Use the request library instead of custom implementation

honzajavorek:

I'm not sure why Dredd uses custom implementation for requesting the server under test. I experienced some problems with it on Windows. This PR introduces the `request` library as a way to request the server under test. It should also help in the future to support proxies, because the library supports it out of the box 😎 

- #204 
- #51 & #591

- [ ] To write docs - N/A
- [x] To write tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "optimist": "^0.6.1",
     "pitboss-ng": "^0.3.2",
     "proxyquire": "^1.7.10",
-    "request": "^2.74.0",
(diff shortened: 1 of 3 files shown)


honzajavorek:
@miiila @w-vi I marked you both as reviewers, but it's a race again - one review is sufficient from whoever comes first 🙂 🏁

honzajavorek:
@w-vi You're super fast! 🏅

Issue #743, opened 2017-03-07
Flaky test: Comparing timestamps when testing apiary reporter


honzajavorek:
This test seems to be flaky on Windows:

```
  1) CLI - Reporters When apiary reporter is used When hooks file uses hooks.log function for logging and hooks are in sandbox mode should send test step with startedAt larger than 'before' hook log timestamp:
     AssertionError: expected 1488902643384 to be >= 1488902643385
```

PR #737, opened 2017-03-03
Basic Windows support (Dredd tested on AppVeyor)

honzajavorek:
#### :rocket: Why this change?

Testing Dredd on Windows-based CI, such as [AppVeyor](http://appveyor.com/), is the first step towards proper Windows support.

#### 👷‍♀️ To Do

- [ ] When merged, remove "experimental" label from the AppVeyor badge on [dredd-example](https://github.com/apiaryio/dredd-example/)

#### :memo: Related issues and Pull Requests

- Closes https://github.com/apiaryio/dredd/issues/204
- Closes https://github.com/apiaryio/dredd/issues/55
- Closes https://github.com/apiaryio/dredd/issues/429

#### :white_check_mark: What didn't I forget?

- [x] To write docs
- [x] To write tests
- [x] To put [Conventional Changelog](https://dredd.readthedocs.io/en/latest/contributing/#sem-rel) prefixes in front of all my commits and run `npm run lint`

PR #752, opened 2017-03-17
Synchronize Pitboss time with real Node.js time

honzajavorek:

Windows tests are extremely flaky. After deeper investigation it turns out [Pitboss](https://www.npmjs.com/package/pitboss-ng) sandbox has different time than the rest of Dredd, so whatever runs within the sandbox can under certain circumstances seem to happen in the future.

- Follow-up to #737 
- Closes https://github.com/apiaryio/dredd/issues/743

- [x] To write docs
- [x] To write tests


honzajavorek:
@w-vi @miiila Who comes first wins 🏁 This one is tricky, feel free to ask if something isn't clear.

src/transaction-runner.coffee
```diff
@@ -174,6 +175,15 @@ class TransactionRunner
...
     for log in results.logs or []
       @logs.push log
+
+    # For unknown reasons, the sandboxed code in Pitboss runs in "future"
+    # on Windows (and just sometimes - it's flaky). The extreme case is
+    # that sandboxed 'before' hooks can have timestamps with a millisecond
+    # later time then the HTTP transaction itself. Following line
+    # synchronizes the time. It waits until the time of the normal Node.js
+    # runtime happens to be later than time inside the Pitboss sandbox.
+    while Date.now() - results.now < 0 then # ...then do nothing...
```
w-vi: Dirty hack but should work, would not a sleep as in `setTimeout` work better. Depends on how out of sync the time is though.

honzajavorek: It's one millisecond. I have the same concerns, but wanted your opinion on whether you think one millisecond is too much to have active waiting. You're the as close to HW as possible person here 😉

honzajavorek: Discussed in person - 1 millisecond is OK (too small time)

PR #759, opened 2017-03-29
attempt to use gobin when provided and fall back to gopath

hcliff:

- [`yup` ] To write docs
- [ `nope`] To write tests
- [`npm run lint` doesn't work] To put [Conventional Changelog](https://dredd.readthedocs.io/en/latest/contributing/#sem-rel) prefixes in front of all my commits and run `npm run lint`

### DIFF
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -148,12 +148,20 @@ class HooksWorkerClient
       return callback(new Error(msg))
 
     else if @language == 'go'
-      gopath = process.env.GOPATH
-      @handlerCommand = "#{gopath}/bin/goodman"
+      # https://golang.org/cmd/go/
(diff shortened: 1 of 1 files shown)


honzajavorek:
@ddelnano I see your 👍 Does that mean you approve? 😄 I'm not really well educated in the Go ecosystem. But the changes look good to me, the code makes sense.

@hcliff Thanks very much for the PR! 🥇 I think I'll merge it! Just out of curiosity - I see `npm run lint` doesn't work for you. What was the issue?

hcliff:
It's actually running fine now, don't mind me!

ddelnano:
@honzajavorek i'm still pretty new to go which is why I didn't have this implemented but this is definitely the right move so thanks @hcliff!

honzajavorek:
@hcliff Thank you very much! 🎉 It will get auto-released shortly.

Issue #670, opened 2016-11-30
Dredd doesn't correctly kill subprocess on *NIX systems


honzajavorek:
It seems like Dredd doesn't properly kill the server process on *NIX systems.

While https://github.com/apiaryio/dredd/issues/669 is a request for improvement, because Windows are still not officially supported and tested, this is a regression and needs to be fixed.

honzajavorek:
The newest version of Dredd [v3.4.0](https://github.com/apiaryio/dredd/releases/tag/v3.4.0) was supposed to introduce consistent killing of subprocesses for *NIX and Windows, but the outcome is it correctly kills processes on macOS and Windows, but doesn't kill them on Linux - it hangs. This is a serious issue and it's being worked on. Pin your version of Dredd to  if you're experiencing problems.

PR #764, opened 2017-04-05
Listen on 'exit' event instead of 'close'

honzajavorek:

When working with spawned subprocesses, it seems like under some circumstances the 'close' event isn't emitted on Linux when the process terminates. The 'exit' event seems to be called consistently accross operating systems.

- Close https://github.com/apiaryio/dredd/issues/670#issuecomment-291900957

- [ ] To write docs - N/A
- [ ] To write tests - N/A (I'm not really sure why in tests the `close` event gets called correctly while manually it doesn't work 😢 )


w-vi:
`close` event is fired when subprocess stdin/stdout are closed, this is not always the case if you the subprocess shares them with others so it is not reliable, good move to use `exit`.

Issue #729, opened 2017-03-01
Use native RTD redirects in docs


honzajavorek:
In https://github.com/rtfd/readthedocs.org/issues/1826#issuecomment-247995569 there is now a comment mentioning how to get the native RTD redirects working. It might be worth move to native redirects instead of custom static files (https://github.com/apiaryio/dredd/issues/467).

Also related to #728. Maybe it's worth the opposite - to keep the redirects in the repository and still generate the static files to keep it under our control. Or we could find a way how to upload a list of redirects to https://readthedocs.org/dashboard/dredd/redirects/ programatically during the build process.

**Note:** Support for redirects to `#links` needs to be verified as it's heavily used at the moment. Also, some pages were split into multiple pages, so interaction with user is actually needed. The page could be rendered in RTD theme colors and offer places to go and search.

Issue #728, opened 2017-03-01
Add build step to documentation


honzajavorek:
Inspect http://docs.readthedocs.io/en/latest/builds.html and add a custom build step. It could help with:

- Generating redirects (docs [1](http://docs.readthedocs.io/en/latest/automatic-redirects.html), [2](http://docs.readthedocs.io/en/latest/user-defined-redirects.html), https://github.com/rtfd/readthedocs.org/issues/1826)
- Generating CLI docs
- Integrating docs with issues on GitHub
- ...

honzajavorek:
Not as easy as it looked like: https://github.com/mkdocs/mkdocs/issues/206#issuecomment-293965195

honzajavorek:
Not as easy as it looked like: https://github.com/mkdocs/mkdocs/issues/206#issuecomment-293965195

Issue #613, opened 2016-08-20
Broken link to chai assertions


honzajavorek:
> https://dredd.readthedocs.io/en/latest/how-it-works/#custom-expectations says to: "check out Chai.js assertions".. but following link shows no mention of Chai. I think it should link to here: https://dredd.readthedocs.io/en/latest/hooks-nodejs/#using-chai-assertions

Issue #562, opened 2016-07-18
Docs search is broken


honzajavorek:
For some reason, the search feature in Dredd docs doesn't really work.

https://readthedocs.org/search/?q=continuous&type=file&project=dredd&version=latest

vs

https://github.com/apiaryio/dredd/search?l=markdown&q=continuous&type=Code&utf8=%E2%9C%93

honzajavorek:
Looks like there are various issues for this:

- https://github.com/rtfd/readthedocs.org/issues/1088
- https://github.com/rtfd/readthedocs.org/issues/1633

RTD-MkDocs integration to blame. With Dredd documentation growing, I'm really tempted to migrate to Sphinx and ReST.

honzajavorek:
https://github.com/rtfd/readthedocs.org/issues/1088 contains some crazy workarounds, implemented e.g. in https://github.com/docksal/docksal/pull/120/.

PR #775, opened 2017-04-26
Migrate docs to Sphinx

honzajavorek:

- We need more extensible docs build system to implement [some of the ideas in our issues](https://github.com/apiaryio/dredd/labels/Epic%3A%20docs), but MkDocs have no hooks/extensions and no official Python API
- We need to fix several bugs on our docs deployed on ReadTheDocs, which have support for MkDocs, but the support is very limited (and they do not pay much attention to fix their bugs related to MkDocs builds)

- Closes https://github.com/apiaryio/dredd/issues/728 (Add build step to documentation)
- Closes https://github.com/apiaryio/dredd/issues/562 (Docs search is broken)
- Helps https://github.com/apiaryio/dredd/issues/551 to get resolved (When docs don't build on RTD, it won't show anywhere)
- Closes https://github.com/apiaryio/dredd/issues/729 (Use native RTD redirects in docs)
- Closes https://github.com/apiaryio/dredd/issues/613 (Broken link to chai assertions)
- Closes https://github.com/apiaryio/dredd/pull/821 (Avoid redirect in link to Jenkins)

#### ✨ Invisible work

- I've created a simple Python script, which compares the old docs and the new docs and informs me about any link tagets which were present previously and would be broken with the new version of docs
- I've created a simple Python script to programmatically upload a list of redirects to the admin interface of RTD (I'm considering to package it and publish to PyPI, but I don't have time for that now) - **Update:** https://pypi.python.org/pypi/rtd-redirects/

#### 🏁 What not to forget after merge

- [ ] Setup correct redirects in the admin of the `dredd` RTD project

#### 👇 Why coverage dropped?

Honestly, I have no idea. The report says `src/reporters/nyan-reporter.coffee` has lower coverage, but I did not touch that file 🤔 

- [x] To write docs
- [x] To write tests


.github/CONTRIBUTING.md
```diff
@@ -150,18 +159,73 @@ about how it works.
...
 
-The main documentation is written in [Markdown][] using [MkDocs][]. Dredd uses
-[ReadTheDocs][] to build and publish the documentation:
+Dredd's documentation is written in [Markdown][] using [Sphinx][]. [ReadTheDocs][] is used to build and publish the documentation:
 
 - [https://dredd.readthedocs.io](https://dredd.readthedocs.io) - preferred long URL
 - [http://dredd.rtfd.org](http://dredd.rtfd.org) - preferred short URL
 
-Source of the documentation can be found in the [docs][] directory. To contribute to Dredd's documentation, you will need to follow the [MkDocs installation instructions](http://www.mkdocs.org/#installation). Once installed, you may use following commands:
+Source of the documentation can be found in the [docs][] directory. To render Dredd's documentation on your computer, you need Python 3 and Node.js installed.
```
kylef: Any Python 3? Perhaps "Python > 3.3" or such?

.github/CONTRIBUTING.md
```diff
@@ -150,18 +159,73 @@ about how it works.
...
+
+#### Installation and Development
+
+1.  Make sure `node` is an executable and `npm install` has been done for the Dredd directory. Extensions to the docs are written in Node.js and Sphinx needs to have a way to execute them.
+2.  [Get Python 3](https://www.python.org/downloads/). On macOS, run `brew install python3`.
+3.  Create a [virtual environment](https://docs.python.org/3/library/venv.html) and activate it:
+
+    ```sh
+    python3 -m venv ./env
```
kylef: I think it's more common to call it `venv` over `env`. Prevents any confusion with environment variables files (often in `.env` files).

.github/CONTRIBUTING.md
```diff
@@ -150,18 +159,73 @@ about how it works.
...
+#### Node.js Extensions
+
+There are some extensions hooked into the build process of [Sphinx][], modifying how the documents are processed. They're written in Node.js, because:
+
+- It's better to have them in the same language as Dredd.
+- This way they're able to import source files (e.g. `src/options.coffee`).
+
+<!-- To display the Hercule syntax, escaping is needed, because this document itself gets processed by Hercule. But escaping by backslashes doesn't work inside Markdown code blocks, so manual <code> element is added. -->
+
+By default, [Hercule][] is attached as an extension, which means you can use the <code class="docutils literal"><span class="pre">:\[Title](link.md)</span></code> syntax for including other Markdown files. All other extensions are custom and are automatically loaded from the `docs/_extensions` directory.
```
kylef: Couldn't you use markdowns backpacks here for the code block instead of using HTML `<code`?

docs/conf.py
```diff
@@ -0,0 +1,352 @@
...
+# You can specify multiple suffix as a list of string:
+source_suffix = '.md'
+source_parsers = {'.md': CommonMarkParser}
+
+# The master document.
+master_doc = 'index'
+
+# General information about the project.
+project = 'Dredd'
+copyright = 'Apiary Czech Republic, s.r.o.'
```
kylef: Is this correct anymore?

docs/conf.py
```diff
@@ -0,0 +1,352 @@
...
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+def get_release():
+    try:
+        # Is internet available? this is to be able to generate docs
+        # e.g. in train without internet connection
+        urllib.request.urlopen('https://www.npmjs.com/package/dredd', timeout=3)
+    except urllib.request.URLError:
+        # Offline, use dummy release number
```
kylef: Maybe for the case on ReadTheDocs (when `READTHEDOCS` is set) this should error to prevent the documentation from being stored and served with the wrong version.

docs/install-node.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# In the ReadTheDocs build, this file is ran by Sphinx' conf.py to install
+# support for Node.js.
+
+set -e
+
+
+if [ "$READTHEDOCS" = 'True' ]; then
```
kylef: Perhaps this check is redundant. You are already checking for the env variable before executing this file:

```python
if os.environ.get('READTHEDOCS') == 'True':
    installation_output = subprocess.getoutput('bash ' + os.path.join(docs_dir, 'install-node.sh'))
    ...
```

.github/CONTRIBUTING.md
honzajavorek: I haven't try it out, but I believe it should be okay with any Python 3. RTD has 3.5, so it might make sense to pin the version to that (at least by mentioning it in docs).

.github/CONTRIBUTING.md
honzajavorek: No :( https://github.com/jamesramsay/hercule/issues/378

docs/conf.py
honzajavorek: No idea 😱 cc @abtris @nadade ... I guess any changes apply to all our `package.json` files as well.

docs/install-node.sh
honzajavorek: Yup, I just wanted to be sure somebody won't execute this locally and cause some mess on their computer. Leaving out the `if` in the config file would make the config file's logic less clear, I think. So I better made this redundant.

.github/CONTRIBUTING.md
Almad: +1

.github/CONTRIBUTING.md
```diff
@@ -150,18 +159,73 @@ about how it works.
...
+[AutoStructify]: https://recommonmark.readthedocs.io/en/latest/auto_structify.html
+
+#### Node.js Extensions
+
+There are some extensions hooked into the build process of [Sphinx][], modifying how the documents are processed. They're written in Node.js, because:
+
+- It's better to have them in the same language as Dredd.
+- This way they're able to import source files (e.g. `src/options.coffee`).
+
+<!-- To display the Hercule syntax, escaping is needed, because this document itself gets processed by Hercule. But escaping by backslashes doesn't work inside Markdown code blocks, so manual <code> element is added. -->
```
Almad: Leftover?

docs/conf.py
Almad: I'd leave it as it is currently as it is in sync with rest of the software. I guess copyright transfer is something we'll be doing later down the line...

honzajavorek:
I appreciate the reviews very much! Thanks! I'll look into this as soon as I have a free moment 👍

.github/CONTRIBUTING.md
honzajavorek: Nope, it refers to the upcoming paragraph:

```
<!-- To display the Hercule syntax, escaping is needed, because this document itself
gets processed by Hercule. But escaping by backslashes doesn't work inside
Markdown code blocks, so manual <code> element is added. -->

By default, [Hercule][] is attached as an extension, which means you can
use the <code class="docutils literal"><span class="pre">:\[Title](link.md)</span></code>
syntax for including other Markdown files. All other extensions are custom and
are automatically loaded from the `docs/_extensions` directory.
```

honzajavorek: It is related to https://github.com/jamesramsay/hercule/issues/378

honzajavorek:
@kylef Addressed your comments. Won't merge before I get green from you as well 💚

.github/CONTRIBUTING.md
kylef: Yeah, the reason I asked is that lots of libraries (even those that support Py 2.7) dropped Python 3 < 3.3 support because some 2.7 behaviours were reintroduced (`%` etc).

Looking at Sphinx release notes, they did drop Python 3.3 too (https://www.sphinx-doc.org/en/stable/changes.html#release-1-5-released-dec-5-2016)

> Drop python 2.6 and 3.3 support

honzajavorek: Yup - http://lucumr.pocoo.org/2013/5/21/porting-to-python-3-redux/#drop-2-5-3-1-and-3-2

honzajavorek: In the steps below this sentence I specified it should be 3.5 and higher, particularly because the build happens on 3.5.

honzajavorek:
book

PR #779, opened 2017-05-09
Syntax highlight API Blueprint in documentation

kylef:
This PR uses [pygments-apiblueprint](https://github.com/kylef/pygments-apiblueprint) to add syntax highlighting for API Blueprint to the rendered documentation on readthedocs. While pygments-apiblueprint is not perfect, it is certainly better than no syntax highlighting. If you spot any rendering issues with API Blueprints feel free to open issues over there and I'll try fixing (I did find https://github.com/kylef/pygments-apiblueprint/issues/1 which I'll fix but no need to block this PR).

Note, this also version locks all of the mkdocs/pygments and related dependencies so that they are not changed and break the docs. If you'd rather not lock the versions I can change requirements.txt to not specify versions, but I think it would be a good idea to lock them so documentation doesn't silently break when a version of mkdocs or dependency has a breaking change.

For readthedocs to install the dependencies declared in `docs/requirements.txt` the path must be specified on read the docs admin. Please enter it there before merging:

screen shot 2017-05-09 at 20 04 22

##### Before

screen shot 2017-05-09 at 20 01 08

##### After

screen shot 2017-05-09 at 20 01 05

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -49,7 +49,7 @@ In order to retrieve transaction names please run Dredd with the `--names` argum
 
 For example, given an API Blueprint file `api-description.apib` as following:
 
-```markdown
+```apiblueprint
 FORMAT: 1A
 
 # Machines API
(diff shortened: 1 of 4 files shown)


honzajavorek:
Thanks! Let's see how the result of the build will look like 👍

honzajavorek:
Looks great!

image

PR #780, opened 2017-05-11
Remove trailing slash from Apiary API URL

honzajavorek:

If Apiary API URL is configured with a trailing slash, it causes double slashes and may cause further problems.

- [ ] To write docs - N/A
- [x] To write tests

### DIFF
--- a/src/reporters/apiary-reporter.coffee
+++ b/src/reporters/apiary-reporter.coffee
@@ -26,7 +26,7 @@ class ApiaryReporter
     @errors = []
     @serverError = false
     @configuration =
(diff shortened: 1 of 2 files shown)


src/reporters/apiary-reporter.coffee
```diff
@@ -202,7 +202,7 @@ class ApiaryReporter
       logger.debug('Apiary reporter response:', JSON.stringify(info, null, 2))
       return callback(null, res, parsedBody)
 
-    body = JSON.stringify(reqBody)
+    body = if reqBody then JSON.stringify(reqBody) else ''
```
kuba-kubula: `reqBody` can't be `false` here?

Issue #744, opened 2017-03-07
Hook handlers should be tested also on Windows


honzajavorek:
We should test whether hook handlers work on Windows - at least the Python and Ruby ones. I played manually with it (hence #742), it seems to work, but it would be nice to be _sure_ at any point in the future that it works.

honzajavorek:
Closed by https://github.com/apiaryio/dredd/pull/782, in favor of https://github.com/apiaryio/dredd/issues/784.

Issue #702, opened 2017-01-25
Change how integration test suite for hook handlers work


honzajavorek:
There are multiple issues with current integration test suite for hook handlers: https://github.com/apiaryio/dredd/issues/672, https://github.com/apiaryio/dredd/issues/647, https://github.com/apiaryio/dredd/issues/644.

My idea how to approach this would be to redesign the test suite and use Travis CI build matrix for running

- Dredd test suite for multiple node versions
- test suite of each of the hook handler implementations, with the version of Dredd in given build

It would enable us to overcome the limitation of sharing credentials with contributors etc. We wouldn't need to push to GitHub or report to GitHub status API. We would just design the matrix using some env variables and then prepare the testing environment accordingly for each build - e.g. for testing Ruby hooks, it would pull (no credentials needed) the Ruby hooks codebase to the Dredd build and run just its Gherkin test suite.

### Challenges

- How to prepare the environment and run the hook handler's test suite? Definition of how to run the test suite is in `.travis.yml` of the pulled repository, but we'll get that file in the middle of running Travis CI build. Is there a way how to interpret the file and execute it in running Travis CI container?
- How to effectively combine node.js build matrix (node 4, node 6) with hook handlers' build matrix?
- Should we use code from `master` of the hook handler repo or should we install its latest version as a package? Which one makes more sense?
- Could we cache / parallelize something?
- This would effectively make the Dredd build a multi-language build - https://github.com/travis-ci/travis-ci/issues/538 Maybe we could use `addons` to provision at least the languages - https://docs.travis-ci.com/user/installing-dependencies/

### Other solutions

- Find existing tool (sth like https://github.com/stephanmg/travis-dependent-builds?)
- Change CI (dependent builds are not on Travis CI roadmap anymore: https://github.com/travis-ci/travis-ci/issues/249)

honzajavorek:
Closed by https://github.com/apiaryio/dredd/pull/782, in favor of https://github.com/apiaryio/dredd/issues/784.

Issue #672, opened 2016-12-02
Integration tests for hook handlers have problems with finding .travis.yml


honzajavorek:
https://travis-ci.org/apiaryio/dredd/jobs/180697371

Apparently, when `.travis.yml` is one of the files changed in your PR, it breaks the script taking care of running the integration tests of hook handlers. Not exactly sure why.

-----

**Update:** It looks like it doesn't matter if `.travis.yml` is changed or not, it just fails.

honzajavorek:
Closed by https://github.com/apiaryio/dredd/pull/782, in favor of https://github.com/apiaryio/dredd/issues/784.

Issue #647, opened 2016-10-25
Integration tests of hook handlers start after PR build finishes


honzajavorek:
...which makes the whole build time very long.

honzajavorek:
Closed by https://github.com/apiaryio/dredd/pull/782, in favor of https://github.com/apiaryio/dredd/issues/784.

PR #782, opened 2017-05-16
Remove hooks handlers testing

honzajavorek:

The integration tests of hook handlers were disabled in the past already due to various issues (#644, #647, #672, #702, #744). Now there's a security issue in one of the libraries the code of the tests uses: https://nodesecurity.io/advisories/sync-exec_tmp-files-readable-by-other-users

image

Dredd already dropped support for Node 0.10 and 0.12, so native `child_process.execSync()` could be used instead of the sync-exec library, but since the code is essentially dead code now, it's like perform fixes in commented-out code. I decided it would be the best to remove the code completely for now and to fix it once someone is re-introducing the test suite.

- #644 
- #647 
- #672 
- #702 
- #744

- [x] To write docs
- [ ] To write tests - N/A


honzajavorek:
I discussed this with @netmilk in person - let's remove this for now and let's revisit soon what can we do to re-introduce the tests.

Issue #795, opened 2017-05-29
Dredd HTTP client should not follow redirects


argos83:
#### Describe your problem

I'm unable to verify one of my API endpoints which produces a 303 redirect to an external service

My endpoint API (swagger) looks like this

```
  /signed_url:
    get:
      description: |
        Redirect to S3 object download url
      consumes: []
      parameters:
         - name: id
           in: query
           description: the S3 object key
           required: true
           type: string
           x-example: 123
      responses:
        '303':
          description: Redirect response to S3
          headers:
            Location:
              description: The S3 signed url to the requested object
              type: string
              format: uri
```
I would like Dredd to verify the expectations on my service response, instead it follows the redirect and checks the expectations on the external service response (which of course won't match my API specification and fails the test).
I can't think of a use case in which following redirects would be a desired behavior, but if so it should be an opt-in setting.

#### What command line options do you use?

```
$ dredd api.swagger.yaml "${BASE_URL}" --hookfiles=./test/dredd-hooks/*.js
```

#### What is in your `dredd.yml`?

no dredd.yml

#### What's your `dredd --version` output?

```
dredd v3.5.0 (Darwin 16.6.0; x64)
```

#### Can you send us failing test in a Pull Request?

If you agree on this being a bug I can create a PR with a fix.

honzajavorek:
I need to check with @netmilk what is the intended behavior, but I think you're right, this looks like a bug. I can imagine Dredd would follow redirects, but probably only if asked, e.g. by a specific command-line option. My hunch is that this behavior was introduced with replacing the native HTTP(S) request implementation with the `request` library in https://github.com/apiaryio/dredd/pull/749 The next steps should be:

- Install version of Dredd before https://github.com/apiaryio/dredd/pull/749 and try whether it follows redirects. If yes, then this is clearly a bug. If not, this is a feature request 😄 
- Write a failing test for the correct behavior
- Fix the behavior, make the test passing

honzajavorek:
I checked with @netmilk and he says previously Dredd **did not follow** redirects. This is a bug, regression.

argos83:
FYI, responses for POST containing a redirect were not being followed. Only GET responses with redirection. Anyways, I've added tests for both cases.

honzajavorek:
Released as v3.5.1 🎉

yogevglucome:
@honzajavorek As you said before, is there any way to use some flag to allow Dredd to follow redirect?
[...]

PR #801, opened 2017-06-15
#795: HTTP client should not follow redirects

argos83:
Dredd is following 3xx redirections (as a result of a GET request, not for POST requests) because that is the default behavior implemented in the `request` package. This makes it impossible to validate API specs that produce a redirection.

#795 

- [ ] To write docs
- [x] To write tests


honzajavorek:
Released as v3.5.1 🎉

PR #803, opened 2017-06-15
Resolve glob patterns before passing them to hook handlers

### LINKED ISSUE #680: Dredd could resolve glob patterns before passing them to hook handlers
### Background

Dredd can be extended by hooks, which can be written in various languages. Languages other than JS are provided by so-called "hook handlers". Those are applications, which start with Dredd and communicate with Dredd via a TCP socket. When hook handler is being started by Dredd, it receives some initial data by command line arguments, namely a list of hook files to load:

> ...it loads files specified as CLI server arguments
> -- [Dredd docs](https://dredd.readthedocs.io/en/latest/hooks-new-language/#what-to-implement)

### The problem

Users can provide path to hook files also in a form of a glob pattern. Dredd should resolve the glob, order the files alphabetically and pass such resolved paths to hook handlers. Reasons:

- Currently, every hook handler implements their own resolution of glob patterns. Because of various glob implementations in individual languages, the behavior might not be the same as in Dredd itself (other CLI options can also be provided with globs and there could be inconsistency).
- Every hook handler implements their own ordering of the resulting paths. Since hook files are imperative scripts, their order is significant and needs to be deterministic.

This issue comes from real user needs expressed by Pull Request in Python hook handler: https://github.com/apiaryio/dredd-hooks-python/pull/26

### What needs to be done

- Test which verifies hook handlers get list of hook files alphabetically sorted.
- Test which verifies Dredd resolves all glob paths to full paths to hook files before it passes them to hook handlers.
- Implementation.
- Docs, especially https://dredd.readthedocs.io/en/latest/hooks-new-language/

### Related issues

- This should close https://github.com/apiaryio/dredd/issues/742 as well.
- Enables https://github.com/apiaryio/dredd-hooks-template/issues/22.

### Expected time frame

3 days

### LINKED ISSUE #742: Dredd should pass hook files as absolute paths to a hook handler
Currently, it seems Dredd just passes the hookfile file names to the hook handler process. However, on Windows this does not seem to work well for relative paths. Dredd should evaluate everything, sort the files alphabetically, and pass a list of absolute paths to the hook handler. That would also solve also https://github.com/apiaryio/dredd/issues/680.

michalholasek:

Currently, Dredd provides hook file paths / globs to hook handlers in the same form (untouched) as they are passed via the CLI. Since every language has its own implementation of glob pattern resolution, it can potentially lead to problems with wrong execution order of hook files.

With the changes introduced in this pull request, Dredd will now resolve glob patterns, matched file paths resolves to their absolute form, sorts them alphabetically and then passes to hook handlers.

Resolves #680 and resolves #742.

- [x] To write docs
- [x] To write tests

### DIFF
--- a/docs/usage-cli.md
+++ b/docs/usage-cli.md
@@ -102,8 +102,7 @@ Extra header to include in every request. This option can be used multiple times
 Show usage information.<br>
 
 ### --hookfiles, -f
-Specifies a pattern to match files with before/after hooks for running tests. Files are executed in
-alphabetical order.<br>
+Specifies a pattern to match files with before/after hooks for running tests<br>
 
 ### --hooks-worker-after-connect-wait
 How long to wait between connecting to hooks worker and start of testing. [ms]<br>
--- a/docs/usage-cli.md
+++ b/docs/usage-cli.md
@@ -102,7 +102,8 @@ Extra header to include in every request. This option can be used multiple times
 Show usage information.<br>
 
 ### --hookfiles, -f
-Specifies a pattern to match files with before/after hooks for running tests<br>
+Specifies a pattern to match files with before/after hooks for running tests. Files are executed in
+alphabetical order.<br>
 
 ### --hooks-worker-after-connect-wait
 How long to wait between connecting to hooks worker and start of testing. [ms]<br>
--- a/src/add-hooks.coffee
+++ b/src/add-hooks.coffee
@@ -12,6 +12,8 @@ sandboxHooksCode = require './sandbox-hooks-code'
 mergeSandboxedHooks = require './merge-sandboxed-hooks'
 HooksWorkerClient = require './hooks-worker-client'
 
+# Ensure platform agnostic path.basename function
+basename = if process.platform is 'win32' then path.win32.basename else path.basename
 
 addHooks = (runner, transactions, callback) ->
   # Note: runner.configuration.options must be defined
@@ -31,10 +33,7 @@ addHooks = (runner, transactions, callback) ->
 
           delete allHooks[hookType][transactionName]
 
-  loadHookFile = (filename, basePath) ->
-    basePath ?= customConfigCwd or process.cwd()
-    filePath = path.resolve(basePath, filename)
-
+  loadHookFile = (filePath) ->
     try
       proxyquire filePath, {
         'hooks': runner.hooks
@@ -55,15 +54,15 @@ addHooks = (runner, transactions, callback) ->
     if typeof(runner.configuration.hooksData) != 'object' or Array.isArray(runner.configuration.hooksData) != false
       return callback(new Error("hooksData option must be an object e.g. {'filename.js':'console.log(\"Hey!\")'}"))
 
-    # run code in sandbox
+    # Run code in sandbox
     async.eachSeries Object.keys(runner.configuration.hooksData), (key, nextHook) ->
       data = runner.configuration.hooksData[key]
 
-      # run code in sandbox
+      # Run code in sandbox
       sandboxHooksCode data, (sandboxError, result) ->
         return nextHook(sandboxError) if sandboxError
 
-        # merge stringified hooks
+        # Merge stringified hooks
         runner.hooks = mergeSandboxedHooks(runner.hooks, result)
 
         # Fixing #168 issue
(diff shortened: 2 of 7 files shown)


docs/usage-cli.md
```diff
@@ -102,7 +102,8 @@ Extra header to include in every request. This option can be used multiple times
 Show usage information.<br>
 
 ### --hookfiles, -f
-Specifies a pattern to match files with before/after hooks for running tests<br>
+Specifies a pattern to match files with before/after hooks for running tests. Files are executed in
```
honzajavorek: This is actually generated from `options.coffee` (should be solved by #775). Edit `options.coffee` and then run `npm run build`.

src/add-hooks.coffee
```diff
@@ -12,6 +12,8 @@ sandboxHooksCode = require './sandbox-hooks-code'
 mergeSandboxedHooks = require './merge-sandboxed-hooks'
 HooksWorkerClient = require './hooks-worker-client'
 
+# Ensure platform agnostic path.basename function
+basename = if process.platform is 'win32' then path.win32.basename else path.basename
```
honzajavorek: AppVeyor failed for this PR, so I guess this doesn't work. When I've seen AppVeyor failing, I was afraid we will need to do some hairy debugging, but reading the code, the problem needs to be on this line, so luckily this is pretty isolated and should be easy to track down :)

src/add-hooks.coffee
```diff
@@ -95,23 +94,43 @@ addHooks = (runner, transactions, callback) ->
...
 
-      # no data found, doing nothing
+      # No data found, doing nothing
       return callback()
 
   # Loading hookfiles from fs
   else
+    # Expand hookfiles - sort files alphabetically and resolve their paths
+    hookfiles = [].concat runner.configuration?.options?.hookfiles
+    files = hookfiles.reduce((result, item) ->
```
honzajavorek: I'd prefer better variable name then `item` - e.g. `path` or `file` or something specific like that.

honzajavorek: The same goes for `input`. Is it `resolvedPaths`? Is it something else?

src/add-hooks.coffee
```diff
@@ -95,23 +94,43 @@ addHooks = (runner, transactions, callback) ->
...
+        # [
+        #   { basename: 'filename1.coffee', path: './path/to/filename1.coffee' }
+        #   { basename: 'filename2.coffee', path: './path/to/filename2.coffee' }
+        # ]
+        .map((filepath) -> basename: basename(filepath), path: filepath)
+        # Sort 'em up
+        .sort((a, b) -> a.basename > b.basename)
+        # Resolve paths to absolute form. Take into account user defined current
+        # working directory, fallback to process.cwd() otherwise
+        .map((item) -> path.resolve(customConfigCwd or process.cwd(), item.path))
```
honzajavorek: 👍 honestly, I didn't even know there's `customConfigCwd` 😄

test/unit/add-hooks-test.coffee
```diff
@@ -125,6 +125,77 @@ describe 'addHooks(runner, transactions, callback)', () ->
...
+        expected = [
+          'multifile_hooks.coffee',
+          'test2_hooks.js',
+          'test_hooks.coffee'
+        ]
+
+        actual = runner.hooks.configuration.options.hookfiles
+
+        actual.forEach (item, index) ->
+          assert.include item, expected[index]
```
honzajavorek: This assert won't happen if the implementation returns empty `actual` array. `assert.deepEqual(actual, expected)` works for me, usually, in these simple cases.

src/add-hooks.coffee
michalholasek: This works as intended, the problem was in `cli-test.coffee` - fixed by https://github.com/apiaryio/dredd/pull/803/commits/5273b321aee06c08c7fa8a00dad1f3f815c89f61.

src/add-hooks.coffee
michalholasek: Addressed in https://github.com/apiaryio/dredd/pull/803/commits/60d2a4fdce00d10dccc66cfbca5d7c0bbe11be8e.

test/unit/add-hooks-test.coffee
michalholasek: Addressed in https://github.com/apiaryio/dredd/pull/803/commits/c09efdb98fe2fe5ba294baefdd481b52e4aae413.

docs/usage-cli.md
michalholasek: Addressed in https://github.com/apiaryio/dredd/pull/803/commits/e867e4ad0e11b95ff53cb867c5ea3c5943d7e6e3.

michalholasek:
@honzajavorek Addressed all your comments, hopefully we'll have a green light in all checks.

test/integration/cli/cli-test.coffee
```diff
@@ -1,5 +1,7 @@
-{assert} = require('chai')
 net = require('net')
+path = require('path')
+
+{assert} = require('chai')
```
honzajavorek: Why this require was moved?

michalholasek: Just a habit - I like to differentiate between default and destructured imports (requires).

PR #809, opened 2017-06-22
Update hooks-go.md

m00sey:
Examples fail to compile.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/docs/hooks-go.md
+++ b/docs/hooks-go.md
@@ -88,7 +88,7 @@ import (
 
 func main() {
     h := hooks.NewHooks()
-	server := hooks.NewServer(hooks.NewHooksRunner(h))
+    server := hooks.NewServer(hooks.NewHooksRunner(h))
     h.BeforeAll(func(t []*trans.Transaction) {
         fmt.Println("before all modification")
     })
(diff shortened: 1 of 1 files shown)


IngmarStein:
The goodman documentation was updated for https://github.com/snikch/goodman/releases/tag/0.2.1 but not here, sorry about that.

m00sey:
Ah I see, that documentation seems a bit clearer too.

You can close this, or link to the other docs, up to you.

Thanks,

ddelnano:
Its probably good to have it in both since the other languages have similar docs but I agree it needs to be thought of when any of the hook servers are updated.

honzajavorek:
@m00sey Thanks for the change! Could you please just edit the commit messages so they're prefixed with `docs: `? Dredd uses Semantic Release, which needs these tags to know whether it should automatically release a new version or not when PR gets merged.

honzajavorek:
@m00sey To me the commits still don't seem to be completely right, but I sense some fighting with git complexity here and I'm here to help - https://github.com/apiaryio/dredd/pull/824 It just took me a bit longer then I anticipated, I'm sorry for that. Thanks for the change!

@ddelnano Would you have any ideas on how we could easily sync the docs? Now that #728 is fixed by #775, almost anything is possible 😎

PR #821, opened 2017-07-18
Fix jenkins link

gmile:

Jenkins link was opening too slowly. The change will avoid redirect and thus open faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,4 +83,4 @@ $ npm install -g dredd
 
 [Travis CI]: https://travis-ci.org/
 [CircleCI]: https://circleci.com/
-[Jenkins]: http://jenkins-ci.org/
+[Jenkins]: https://jenkins.io/


honzajavorek:
Hi @gmile, thanks! The CI fails, because we need the commit to be formatted as `docs: Fix jenkins link`, so [Semantic Release could automatically release correct Dredd versions according to commits](https://dredd.readthedocs.io/en/latest/contributing/#semantic-release-and-conventional-changelog). I appreciate the contribution, but the commit message needs to be fixed first.

Also, there's https://github.com/apiaryio/dredd/pull/775, which already contains the change and will be merged soon. It also contains checker for broken links and redirects, so this shouldn't happen anymore.

It's great to see someone fixing Dredd docs! Small change, but I'm actually quite excited :) I leave it up to you if you want to get this passing & merged or if https://github.com/apiaryio/dredd/pull/775 is good enough for you. I'd be happy to have your name in contributors.

gmile:
Hi @honzajavorek, thank for looking into the failure! I've force-pushed a commit with a message formatted per your requirement. Hopefully the build passes now.

Issue #825, opened 2017-07-25
Internal links in docs are broken


honzajavorek:
Nobody noticed 😢 Since #775 internal links are broken, because they omit the `/en/latest/` part. All internal links lead to 404.

honzajavorek:
Just to be clear, this is 100 % my responsibility. I merged the new docs and in the excitement I did not click through the new ones sufficiently.

Issue #551, opened 2016-07-04
When docs don't build on RTD, it won't show anywhere


honzajavorek:
- At least badge on README?
- Status for Pull Requests touching docs?
- Adding building docs to CI as part of tests?

honzajavorek:
Idea: Get build status from RTD API, wait for it in the CI - https://readthedocs.org/api/v2/build/5578661/

honzajavorek:
RTD have a badge! 😏

PR #826, opened 2017-07-25
Fix internal links in docs

honzajavorek:

Local links do not work on current version of the docs. This fix is tested on https://dredd.readthedocs.io/en/honzajavorek-fix-local-links/ and it looks like it resolves the problem.

Fixes https://github.com/apiaryio/dredd/issues/825.

- [ ] To write docs - N/A
- [ ] To write tests - N/A


honzajavorek:
Not able to get the builds passing. This needs to be rebased on top of https://github.com/apiaryio/dredd/pull/829. However in that case, the CI will stop building the docs, so we won't be sure whether docs are building correctly 😕 (related to #551) But that's as good as it gets for now.

Issue #832, opened 2017-08-01
Make GOPATH environment variable optional (now mandatory)


vladosaurus:
From Go 1.8 onward it is optional to specify the GOPATH environment variable:
https://golang.org/cmd/go/

as seen here:
https://github.com/apiaryio/dredd/blob/master/src/hooks-worker-client.coffee#L131

honzajavorek:
@ddelnano @snikch What do you think about this?

ddelnano:
@honzajavorek the default should be `$HOME/go` if it is not set.  So we just need to provide that as a default.  I can try and submit a patch for that sometime this week.

snikch:
Makes sense to me 👌

honzajavorek:
This is being addressed by @vladosaurus' own PR: https://github.com/apiaryio/dredd/pull/835 👍

honzajavorek:
I believe this has been already tackled by https://github.com/apiaryio/dredd/pull/928.

PR #835, opened 2017-08-02
fixing GOPATH issue: #832

vladosaurus:

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/src/hooks-worker-client.coffee
+++ b/src/hooks-worker-client.coffee
@@ -126,21 +126,28 @@ class HooksWorkerClient
       # If the GOBIN environment variable is set, commands are installed to the
       # directory it names instead of DIR/bin. GOBIN must be an absolute path."
       # Use the gobin if provided, otherwise fall back to gopath
-      gobin = process.env.GOBIN
-      if !gobin
-        gobin = "#{process.env.GOPATH}/bin"
-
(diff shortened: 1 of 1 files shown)


src/hooks-worker-client.coffee
```diff
@@ -126,21 +126,28 @@ class HooksWorkerClient
...
-      else
-        callback()
+      {exec} = require 'child_process'
+      getGobin = (callback) ->
+        gobin = process.env.GOBIN
+        if !gobin
+          if process.env.GOPATH
+            gobin = "#{process.env.GOPATH}/bin"
+          else
+            exec "go env GOPATH", (error, stdout, stderr) -> callback "#{stdout.trim()}/bin"
```
honzajavorek: I don't like the subprocess call that much. Also, I'm not sure whether this would work correctly on Windows. If Go spec clearly states the default value is `$HOME/go`, I think we can do something like

```coffee
os = require('os')

...

else
  gobin = "#{os.homedir()}/go/bin"
```

What do you think? This would also be synchronous and we wouldn't need to wrap it in the `getGobin` function.

Also note that typically, `require` statements appear on the top of each CoffeeScript (or JavaScript) file. It's not very usual to do conditional requiring.

Issue #617, opened 2016-08-29
Find a way how to work with binary files in hooks


aledeg:
I don't know if I should open that ticket here or [here](https://github.com/ddelnano/dredd-hooks-php).

I am trying to change the request body with the content of a file with a hook, but it does not work.
I have the following error:

```
SyntaxError: Unexpected end of input
  at Object.parse (native)
  at Socket.<anonymous> (/usr/lib/node_modules/dredd/lib/hooks-worker-client.js:268:34)
  at emitOne (events.js:90:13)
  at Socket.emit (events.js:182:7)
  at readableAddChunk (_stream_readable.js:153:18)
  at Socket.Readable.push (_stream_readable.js:111:10)
  at TCP.onread (net.js:534:20)
```

This is what I have in my hook:

``` php
$transaction->request->body = file_get_contents($filename);
```

Is this the good way of doing that?

Ping @ddelnano

ddelnano:
@aledeg can you give more information about the http request you are trying to send??  What the type of file this is, content type of the req should be, etc.

aledeg:
@ddelnano here is the information you wanted.

- method used: POST
- type of file: jpeg image
- content type: image/*

Here is the actual image used for the tests:
![default](https://cloud.githubusercontent.com/assets/3056148/18082093/8f46b300-6e9d-11e6-8c14-44673e0cb144.jpg)

I know that my api is working because I've tested it with behat and manually. It seems limited to dredd so far.

aledeg:
@honzajavorek Any insight about that issue?

honzajavorek:
@aledeg We're still missing a good canonical example for sending/testing binary data in Dredd using hooks. It's on my roadmap to prepare related examples, write tests for them so we're sure Dredd really works the way we recommend and put them to documentation. Related issues: #52 #87 and maybe even more.

From what you shared I can't tell exactly whether it's PHP hook handler issue, Dredd issue or issue in your hooks. Please try to investigate further what exactly might be happening so we can take a closer look.

- What command line options do you use? `$ dredd ( write your CLI options here )`
- What is in your `dredd.yml`?
- What's your `dredd --version` output?
- Does `dredd --level=debug` uncover something?

aledeg:
@honzajavorek I've made some more tests. Here is what I've got now.

* error message:
```
SyntaxError: Unexpected end of JSON input
  at Object.parse (native)
[...]

Issue #87, opened 2014-08-04
Dredd testing of APIary calls returning an image / PNG


imp-partner-test:
Dredd tests of APIary calls returning an image / PNG fail. Is it possible to check not the content of the returned image but just that there is an image not null as response?
Here is the output of my test:

[31mfail [39m: GET /imp-endpoint/api/V1/Resources?left_lower_lat=52.482780222078205&left_lower_lng=13.3154296875&right_upper_lng=13.359375&right_upper_lat=52.50953477032729&width=250&heigth=250 duration: 1176ms
 [31mfail [39m: body: Real and expected data does not match.
 [32mrequest [39m: 
body: 
headers: 
    accept: image/png
    User-Agent: Dredd/0.3.7 (Linux 3.5.0-52-generic; ia32)
    Authorization: Basic ZG1wOmRtcEAxMjM0
uri: /imp-endpoint/api/V1/Resources?left_lower_lat=52.482780222078205&left_lower_lng=13.3154296875&right_upper_lng=13.359375&right_upper_lat=52.50953477032729&width=250&heigth=250
method: GET

 [31mexpected [39m: 
headers: 
    content-type: image/png
body: 
�PNG
statusCode: 200

 [31mactual [39m: 
statusCode: 200
headers: 
    x-powered-by: Servlet/3.0 JSP/2.2 (GlassFish Server Open Source Edition 3.1.2.2 Java/Oracle Corporation/1.7)
    server: GlassFish Server Open Source Edition 3.1.2.2
    access-control-allow-origin: *
    access-control-allow-methods: GET, POST, PUT, DELETE, OPTIONS
    content-type: image/png
    content-length: 26572
    date: Tue, 15 Jul 2014 09:00:26 GMT
body: 
�PNG
IHDR    � ?1g�IDATx^��

http://stackoverflow.com/questions/24824867/dredd-testing-of-apiary-calls-returning-an-image-png

jaydiablo:
Has there been any updates with regards to this?

The workaround mentioned in #52 doesn't seem to work any longer, but I haven't seen any mention of binary responses being supported in the API Blueprint format either.

Am I missing something, or is this still not supported?

Thanks!

jaydiablo:
Well, another workaround, which is working for me, is to modify the transaction in a preValidation hook:

```
hooks.beforeValidation("Machines > Machines collection > Get Machines", function (transaction) {
    transaction.real.body = '';
});
```

[...]

PR #836, opened 2017-08-03
Support setting Request encoding for response

bb010g:
Mostly useful for Node.js hooks that want to deal with binary data.

Request keeps on messing up higher bytes in some binary data I want to test in a hook.

[Epic: empty responses and binary files](https://github.com/apiaryio/dredd/issues?q=label%3A%22Epic%3A%20empty%20responses%20and%20binary%20files%22)
#617, #87

- [X] To write docs
- [ ] To write tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "pretest": "npm run build",
     "test": "mocha \"test/**/*-test.coffee\"",
     "test:coverage": "scripts/coverage.sh",
-    "prepublish": "npm run build",
+    "prepare": "npm run build",
+    "prepublish": "check-node-version --npm \">=4\" || npm run prepare",
     "coveralls": "scripts/coveralls.sh",
     "semantic-release": "scripts/semantic-release.sh"
   },
@@ -55,6 +56,7 @@
   },
   "devDependencies": {
     "body-parser": "^1.17.1",
+    "check-node-version": "^2.1.0",
     "coffee-coverage": "^2.0.1",
     "coffeelint": "^1.15.7",
     "conventional-changelog-lint": "^1.1.9",
--- a/docs/data-structures.md
+++ b/docs/data-structures.md
@@ -27,6 +27,7 @@ Transaction object is passed as a first argument to [hook functions](hooks.md) a
     - headers (object) - keys are HTTP header names, values are HTTP header contents
     - uri: `/message` (string) - request URI as it was written in API description
     - method: `POST` (string)
+    - encoding: `utf8` (object) - response encoding
 - expected (object) - the HTTP response Dredd expects to get from the tested server
     - statusCode: `200` (string)
     - headers (object) - keys are HTTP header names, values are HTTP header contents
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -456,6 +456,7 @@ class TransactionRunner
     options.method = transaction.request.method
     options.headers = transaction.request.headers
     options.body = transaction.request.body
+    options.encoding = transaction.request.encoding
     options.proxy = false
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -19,7 +19,8 @@
     "pretest": "npm run build",
     "test": "mocha \"test/**/*-test.coffee\"",
     "test:coverage": "scripts/coverage.sh",
-    "prepublish": "npm run build",
+    "prepare": "npm run build",
+    "prepublish": "check-node-version --npm \">=4\" || npm run prepare",
```
honzajavorek: Ah, I did not notice the commit message during review:

> See https://iamakulov.com/notes/npm-4-prepublish/ , npm/npm#16685 , and https://stackoverflow.com/a/44136814/1098906

Could you send this change in a separate PR, please? At the same time, I'm still not completely sure about it, because Dredd is being published on CI exclusively, so we could just make sure that `package.json` contents match the npm version which publishes the package. That basically means just pinning npm to a specific version in the Travis CI config and then having the `package.json` specific to that npm version.

bb010g: Sure, I can separate that out. I ended up sticking it in here because of the need to depend on my Git branch with Yarn, and yarnpkg/yarn#2875 / yarnpkg/yarn#3553 requiring an explicit `prepare` for that. The benefit of using this slightly more defensive syntax is that, if you need to do source compilation for whatever reason, you can still work with an older npm. If we don't care about npm 3 compatibility though, the prepublish script can be dropped entirely.

bb010g:
For non-JS hooks, you'd probably want to convert any Buffers you find to/from the native byte array type for that language before & after each hook. I'm not too familiar with how the hook architecture works, though, so I'd have to figure out where all that's handled. Since we're already wrapping the request options, having an extra `'bytes'` option that maps to null should work.

As far as I've tested, Buffers are safe to have any time in `request.body` before sending. `'binary'` encoded strings will also remain as-is until Request goes to send them, in which case they'll be converted to `'utf8'`. Converting to a Buffer here at the end of a `before*` hook works fine. As for dealing with binary data on the `real.body` end, Gavel wants a string, so you have to convert back to a string at the end of a `before*Validation` hook. Automatically performing this step should be fine.

package.json
honzajavorek: Hmm makes sense. Let's do this, then!

honzajavorek:
I'm very sorry we were not able to get back to this. It's a significant change, which needs some design work and a lot of tests to be completely sure it works as it should be and that we're not breaking something.

I need to close this as we're going to work on #705 first, which requires having no pending Pull Requests. We'll get back to this proposal as soon as we're addressing the rest of the "empty responses and binary files" epic (the empty responses were implemented in https://github.com/apiaryio/dredd/pull/906).

Thanks for all the work! I'll do my best for this contribution not to be forgotten!

honzajavorek:
@bb010g I'm trying to revisit this and I'm struggling to reproduce the problem you had:

> Request keeps on messing up higher bytes in some binary data I want to test in a hook.

Would you be able to provide an example of such data so I can make it part of Dredd's test suite? I thought I'd be able to reproduce this with any binary data, but I wasn't successful with an arbitrary PNG image.

bb010g:
Sorry, I'm not using Dredd currently and don't have access to the old files, but IIRC it was failing on bytestrings that weren't valid UTF-8. A single byte of 0xff should do it.

honzajavorek:
This is going to get resolved, tested, and documented in https://github.com/apiaryio/dredd/pull/1094. Thank you @bb010g for pointing the issue out, for trying to provide a solution, and for discussing the issue. It was really helpful!

PR #819, opened 2017-07-17
Hotfix: Return correct exit code when tests fail

michalholasek:
#### :rocket: Why this change?

Currently, Dredd returns wrong exit code (`O`) when test(s) fail (should return `1`). This change fixes the problem.

#### :memo: Related issues and Pull Requests

Resolves https://github.com/apiaryio/dredd/issues/818.

#### :white_check_mark: What didn't I forget?

- [ ] To write docs - N/A
- [ ] To write tests - N/A
- [x] To put [Conventional Changelog](https://dredd.readthedocs.io/en/latest/contributing/#sem-rel) prefixes in front of all my commits and run `npm run lint`

PR #838, opened 2017-08-03
Add test case for CLI exit code regression

michalholasek:
This PR adds a test case for a regression fixed in https://github.com/apiaryio/dredd/pull/819.

- [ ] To write docs - N/A
- [x] To write tests


michalholasek:
I'll investigate. Makes me wonder if the solution I came up with originally was more suited or not.

Issue #786, opened 2017-05-18
Multipart not passing formdata to endpoint


fidransky:
Hi everybody,

Im using Dredd to test our Jersey-powered API and absolutely love it. However I cannot figure out how to send valid and complete multipart form data to a file upload endpoint we have using Dredd. Im pretty sure the endpoint works correctly since the application (AngularJS) works perfectly.

In my API blueprint, I have these lines (copied directly from Chrome):

```
### Create a New Route using datafile upload [POST /areas/{areaId}/routes/datafile]

+ Request (multipart/form-data; boundary=---WebKitFormBoundaryvkZWodTjTABclggB)

        ------WebKitFormBoundaryvkZWodTjTABclggB
        Content-Disposition: form-data; name="dataFile"; filename="yoso.txt"
        Content-Type: text/plain


        ------WebKitFormBoundaryvkZWodTjTABclggB
        Content-Disposition: form-data; name="route"

        {"name":"Whatever"}
        ------WebKitFormBoundaryvkZWodTjTABclggB--
```

The server side looks as following (for Java fans):

```
@Path("areas/{areaId}/routes/datafile")
@POST
@Consumes(MediaType.MULTIPART_FORM_DATA)
Response createRoute(@PathParam("areaId") Long areaId,
	 @FormDataParam("dataFile") InputStream dataFile,
	 @FormDataParam("dataFile") FormDataContentDisposition dataFileHeader,
	 @FormDataParam("route") String routeJSON);
```

To me, it seems that Dredd does not pass the formdata correctly thus the server returns error 500.

Im using Dredd version v3.4.4 (Windows_NT 10.0.15063; x64).

Ive read lots of various how-tos and SO topics, including the [official guide](https://dredd.readthedocs.io/en/latest/how-to-guides/#sending-multipart-requests) but Im still struggling with it. In case you need more information, feel free to ask.

Thanks

honzajavorek:
@fidransky Thanks for reporting this! This is a bug. I was able to reproduce the problem. Steps to reproduce:

1. Create a public API project on apiary.io
2. Use following API Blueprint:

```apib
FORMAT: 1A

# Test Form Data API

# POST /data

+ Request (multipart/form-data;boundary=---BOUNDARY)

    + Body

            ---BOUNDARY
            Content-Disposition: form-data; name="text"
            Content-Type: text/plain

            Hello
            ---BOUNDARY
            Content-Disposition: form-data; name="json"; filename="filename.json"
[...]

PR #841, opened 2017-08-04
Modify multipart body and calculate Content-Length in the right order

honzajavorek:

Current implementation of multipart request bodies is broken. See https://github.com/apiaryio/dredd/issues/786.

In https://github.com/apiaryio/dredd/pull/734 Dredd's custom implementation of HTTP requesting was replaced with the `request` library. There is some special handling of multipart requests, solving https://github.com/apiaryio/api-blueprint/issues/401, which modifies the body payload and also modifies the `Content-Length` header accordingly. This wasn't updated with introduction of the `request` library. That resulted in the `Content-Length` being sent with incorrect number, causing the server under test to hang, waiting for more bytes.

My changes makes the code dealing with https://github.com/apiaryio/api-blueprint/issues/401 a lot simpler and more functional (no side effects). It removes any modifications to `Content-Length` and instead it changes the order in which the body gets modified and the header gets calculated.

#### :memo: Related issues and Pull Request

- https://github.com/apiaryio/dredd/pull/734
- https://github.com/apiaryio/api-blueprint/issues/401
- https://github.com/apiaryio/dredd/issues/74
- https://github.com/apiaryio/dredd/pull/118

Fixes https://github.com/apiaryio/dredd/issues/786.

- [ ] To write docs - N/A
- [x] To write tests

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -553,6 +553,9 @@ class TransactionRunner
   # An actual HTTP request, before validation hooks triggering
   # and the response validation is invoked here
   performRequestAndValidate: (test, transaction, hooks, callback) ->
+    if transaction.request.body and @isMultipart(transaction.request.headers)
+      transaction.request.body = @fixApiBlueprintMultipartBody(transaction.request.body)
+
     @setContentLength(transaction)
     requestOptions = @getRequestOptionsFromTransaction(transaction)
 
@@ -585,10 +588,6 @@ class TransactionRunner
           return callback(@hookHandlerError) if @hookHandlerError
 
           @validateTransaction test, transaction, callback
-
-
-    if transaction.request['body'] and @isMultipart requestOptions
-      @replaceLineFeedInBody transaction, requestOptions
 
     try
       @performRequest(requestOptions, handleRequest)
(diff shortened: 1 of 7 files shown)


src/transaction-runner.coffee
```diff
@@ -674,17 +673,17 @@ class TransactionRunner
...
-
-  replaceLineFeedInBody: (transaction, requestOptions) ->
-    if transaction.request['body'].indexOf('\r\n') == -1
-      transaction.request['body'] = transaction.request['body'].replace(/\n/g, '\r\n')
-      transaction.request['headers']['Content-Length'] = Buffer.byteLength(transaction.request['body'], 'utf8')
-      requestOptions.headers = transaction.request['headers']
+  isMultipart: (headers) ->
+    contentType = caseless(headers).get('Content-Type')
+    return false unless contentType
+    return contentType.indexOf('multipart') > -1
```
michalholasek: You can join these two `return`s into one - `return false if not contentType else contentType.indexOf('multipart') > -1`.

honzajavorek: Addressed in 3055aba

michalholasek: Honestly, I think that change made in https://github.com/apiaryio/dredd/commit/3055aba05315ca2e4a9f68a23c124c5ed1562890 is worse than the original two returns. I think that solution from above reads a little bit better, yours has a little bit of Yoda tongue in it 🙂

honzajavorek: Ha, I though that was typo of yours, I did not know that's valid CoffeeScript 😄 

```coffee
return false if not contentType else contentType.indexOf('multipart') > -1
```

won't compile. I'll change it to normal `if/else` 😏

honzajavorek: Addressed in a4da5dd

michalholasek: OK!

honzajavorek:
The drop in coverage is caused by "statistics". There's less lines in the transaction runner now, which means the uncovered part takes larger percentage of it. The drop is thus caused by Dredd consisting of less lines in overall, not by adding uncovered code:

image

honzajavorek:
_Got verbal approval from @michalholasek of the coverage drop explanation, merging._

Issue #818, opened 2017-07-13
Errors do not affect CLI exit code


clementgarnier:
#### Describe your problem

Dredd CLI returns 0 when tests are errored, as long as none are failing, e.g.
```
complete: 0 passing, 0 failing, 1 errors, 0 skipped, 1 total
```

#### What command line options do you use?

```
$ dredd
```

#### What is in your `dredd.yml`?

```yaml
dry-run: null
hookfiles: null
language: nodejs
sandbox: false
server: /foo/bar/baz
server-wait: 3
init: false
custom: {}
names: false
only: []
reporter: []
output: []
header: []
sorted: false
user: null
inline-errors: false
details: false
method: []
color: true
level: info
timestamp: false
silent: false
path: []
hooks-worker-timeout: 5000
hooks-worker-connect-timeout: 1500
hooks-worker-connect-retry: 500
hooks-worker-after-connect-wait: 100
hooks-worker-term-timeout: 5000
hooks-worker-term-retry: 500
hooks-worker-handler-host: 127.0.0.1
hooks-worker-handler-port: 61321
config: ./dredd.yml
blueprint: sample_blueprint.md
endpoint: 'http://127.0.0.1:3000'
```

#### What's your `dredd --version` output?

```
dredd v4.1.1 (Darwin 16.6.0; x64)
```

#### Can you send us failing test in a Pull Request?

To reproduce, run dredd with above configuration; note `/foo/bar/baz` is voluntarily inexistent, which will fail to start the server and mark all tests as error. A `sample_blueprint.md` can be created with any valid blueprint (I used the first example [here](https://raw.githubusercontent.com/apiaryio/api-blueprint/master/examples/01.%20Simplest%20API.md)) but not required to reproduce, as a missing blueprint file will also trigger the error state.

michalholasek:
Hello @clementgarnier,

Thank you for the issue! As @honzajavorek said in https://github.com/apiaryio/dredd/issues/765#issuecomment-314994049, this is a serious issue and should be fixed ASAP - I'll start to dig into it right away.

clementgarnier:
Thank you so much for the timely fix! When should we expect this to be released?

honzajavorek:
It gets automatically released once [this build](https://travis-ci.org/apiaryio/dredd/builds/254403285) passes. There are some flaky tests and we did not notice it failed and the release did not happen.

honzajavorek:
[...]

PR #851, opened 2017-08-09
Return correct exit code when server fails to start

michalholasek:
This PR fixes the issue with wrong CLI exit code when running Dredd with incorrect argument to `--server` command (reported in https://github.com/apiaryio/dredd/issues/818).

Closes #838

- [ ] To write docs - N/A
- [x] To write tests

### DIFF
--- a/src/dredd-command.coffee
+++ b/src/dredd-command.coffee
@@ -53,7 +53,7 @@ class DreddCommand
 
   # Gracefully terminate server
   stopServer: (callback) ->
-    unless @serverProcess?
+    if not @serverProcess or not @serverProcess.spawned
       logger.verbose('No backend server process to terminate.')
       return callback()
     if @serverProcess.terminated
@@ -219,9 +219,9 @@ class DreddCommand
       @serverProcess.on 'exit', =>
         logger.info('Backend server process exited')
 
-      @serverProcess.on 'error', (error) =>
-        logger.error('Command to start backend server process failed, exiting Dredd', error)
-        @_processExit(2)
+      @serverProcess.on 'error', (err) =>
+        logger.error('Command to start backend server process failed, exiting Dredd', err)
+        @_processExit(1)
 
       # Ensure server is not running when dredd exits prematurely somewhere
       process.on 'beforeExit', =>
@@ -332,11 +332,9 @@ class DreddCommand
   exitWithStatus: (error, stats) ->
     if error
       logger.error(error.message) if error.message
-      process.exitCode = 1
       return @_processExit(1)
 
     if (stats.failures + stats.errors) > 0
-      process.exitCode = 1
       @_processExit(1)
     else
       @_processExit(0)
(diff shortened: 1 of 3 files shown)


honzajavorek:
Failing in all builds:

```
  Babysitting Child Processes
    when gracefully terminated by childProcess.signalTerm()
      process without support for graceful termination
        ...

TypeError: Cannot read property 'indexOf' of undefined
```

michalholasek:
Shit

@honzajavorek Fixed.

src/dredd-command.coffee
```diff
@@ -219,9 +219,9 @@ class DreddCommand
       @serverProcess.on 'exit', =>
         logger.info('Backend server process exited')
 
-      @serverProcess.on 'error', (error) =>
-        logger.error('Command to start backend server process failed, exiting Dredd', error)
-        @_processExit(2)
+      @serverProcess.on 'spawnerror', (err) =>
```
honzajavorek: I just realized we probably want to do this in case of any kind of error happening to the backend server child process. If the process errors, Dredd cannot ensure the testing will make sense, and it should stop testing and terminate. So I propose we actually keep `@serverProcess.on 'error'` here. What do you think?

test/integration/cli/server-process-cli-test.coffee
```diff
@@ -90,6 +90,26 @@ describe 'CLI - Server Process', ->
       it 'should exit with status 0', ->
         assert.equal dreddCommandInfo.exitStatus, 0
 
+    describe 'When it fails to start', ->
```
honzajavorek: I just realized there's one more side to the problem: The process starts, but Dredd isn't able to connect to it using HTTP. Looking into coverage, this scenario seems to be untested:

image

_transaction-runner.coffee_

-----

I hope it would still work as expected (Dredd fails all tests and it ends with exit code 1), but it would be awesome if we had a test for that as well. What do you think?

It should be fairly easy to test, with the same approach, just spawning an existing script, which hangs the same way as server would do (endless `setTimeout` loop?), but which doesn't handle HTTP in any way. Maybe there's even already a script like that in `test/fixtures`, I haven't take a look.

src/dredd-command.coffee
michalholasek: I have no problem with that, we can differentiate between different type of runtime errors in single error handler hub.

test/integration/cli/server-process-cli-test.coffee
michalholasek: Agree, I'll have a look.

src/dredd-command.coffee
michalholasek: Reflected in https://github.com/apiaryio/dredd/pull/851/commits/ab90347f3db951441f0cfaa3b30ff9cae901c12d.

test/integration/cli/server-process-cli-test.coffee
honzajavorek: My attempt is in https://github.com/apiaryio/dredd/pull/862. If the section screenshoted  above won't get green, then the test is useless and my PR should get closed.

Issue #834, opened 2017-08-01
Hook socket message (documentation) inconsistencies


hobofan:
While implementing the hook handler for Rust I stumbled upon a few things that I feel are mis-/underdocumented. I'll write up a PR for it, but I'd like to get some input on it first.

The documentation under https://github.com/apiaryio/dredd/blob/6445caa0976eaf83b460bf7eade7ed73843845de/docs/hooks-new-language.md#tcp-socket-message-format shows a few event types that don't actually exist (or are not sent) like `before` and `after` and instead have to be triggered by some application logic in the hook handler. I think that should be mentioned by the documentation.

The `data` field can either be an array or an object. I have seen that it is only an array for `beforeAll` and `afterAll` events, but never have seen more than one element in it. Are there actually any circumstances under which there might be more than element in it?

honzajavorek:
@hobofan Thanks for all the work and the investigation. I need to take a look at the `before` and `after` events, because it sounds suspicious to me that they are not sent. I mean - it might be a bug in Dredd, not in docs 😱 I need to verify that part.

Regarding `beforeAll` and `afterAll`, they get an array of all transactions in the API description document. If you tested with just one endpoint, you get an array of a single transaction.

hobofan:
> I mean - it might be a bug in Dredd, not in docs 😱 I need to verify that part.

I took that bit of arcane knowledge from the golang implementation.

Fixing that bug would very certainly break the Go and Rust implementations. Considering that this would be another breaking change besides https://github.com/apiaryio/dredd/pull/712#issuecomment-281220810, I feel that there is a need for introducing versioning of the hook handling API to allow for a more robust API evolvement in the future, but thats a topic for its own issue... 😅

honzajavorek:
> I took that bit of arcane knowledge from the golang implementation.

I've just realized that in the time I was learning about Dredd, there were some issues from @snikch filed, which addressed certain things which came up during his work on the Go hooks. Digged them up and found https://github.com/apiaryio/dredd/issues/271, which sounds completely relevant to this issue.

The hook handling API should certainly get more attention. There are multiple issues to be addressed, some work in progress, ... Need to find time to tidy up there.

honzajavorek:
@hobofan If the second part of your issue about `beforeAll` & `afterAll` and array of transactions is resolved, would you mind if I merge this with https://github.com/apiaryio/dredd/issues/271?

[...]

PR #853, opened 2017-08-10
docs: fix inconsistencies in hook socket message docs

### LINKED ISSUE #271: hooks-worker-client should sent all events.
Making this a separate issue, split from #269.

> The before, beforeValidation and after hooks aren't submitted when using the hooks-worker-client . Since the internal hook handlers are just like the ones required in the end handler, you can't even add a hooks.before(fn()) callback, because it expects a name (whereas the names are defined in the compiled binary logic). To get around this, and send the events, I've duplicated the beforeEach, beforeEachValidation and afterEach callbacks, but changed the event name in each. This way it's up to the hook handler to decide whether the event is relevant (i.e. there is a named callback available). Another approach to this would be to advise the hook handlers to consume the beforeEach event twice, once for everything, and then consider it the same event for before - except the documentation clearly states it should expect a separate before event.

If you check out where the worker client adds the callbacks to the `hooks` value, you'll see that no callbacks are supplied for `before`, `beforeValidation` or `after`.

https://github.com/apiaryio/dredd/blob/master/src/hooks-worker-client.coffee#L117

This is because those callbacks need a name to match against, however those `events` still need to be sent to a hook handler to determine if it has its own matching callback for the specific event name.

In my code, I added a `before` event, by duplicating `beforeEach` and changing the event name. It then became the responsibility of my hook handler to determine if the event `name` matches the callback registered for the `before` event.

``` coffee
hooks.beforeEach (transaction, callback) ->
  # avoiding dependency on external module here.
  uuid = Date.now().toString() + '-' + Math. random().toString(36).substring(7)

  # send transaction to the handler
  message =
    event: 'before' # Note this is `before` not `beforeEach` now
    uuid: uuid
    data: transaction

  …

  emitter.on uuid, messageHandler
```

hobofan:

CLOSES #271 

Also adresses my question of #834, what the array of transaction objects contains.

#271, #834

- [X] To write docs
- [X] ~To write tests~ N/A

### DIFF
--- a/docs/hooks-new-language.md
+++ b/docs/hooks-new-language.md
@@ -59,17 +59,14 @@ On Linux or macOS, Dredd uses the `SIGTERM` signal to tell the hook handler proc
 - transaction (object)
     - uuid: `234567-asdfghjkl` (string) - Id used for event unique identification on both server and client sides
     - event: `event` (enum) - Event type
-        - beforeAll (string)
-        - beforeEach (string)
-        - before (string)
-        - beforeEachValidation (string)
-        - beforeValidation (string)
-        - after (string)
-        - afterEach (string)
-        - afterAll (string)
+        - beforeAll (string) - Signals the hook handler to run the `beforeAll` hooks
+        - beforeEach (string) - Signals the hook handler to run the `beforeEach` and `before` hooks
+        - beforeEachValidation (string) - Signals the hook handler to run the `beforeEachValidation` and `beforeValidation` hooks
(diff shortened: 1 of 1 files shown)


honzajavorek:
@hobofan This is wonderful. Thank you so much! I have to ping @w-vi as well, he'll be so happy 😄 Just yesterday we've been talking about how this and the fact the API between Dredd and hooks really needs some love.

w-vi:
:heart:

Issue #890, opened 2017-10-11
Hook files are not always executed in alphabetical order.


matthewl-biomatters:
As an example, my dir contains 14 files:
```
$ ls dredd-example
a  b  c  d  e  f  g  h  i  j  k  l  m  n
```

When I run Dredd, it logs this showing they are not in alphabetical order:
```
info: Found Hookfiles:
0=/Users/matthewl/projects/dredd-example/h,
1=/Users/matthewl/projects/dredd-example/a,
2=/Users/matthewl/projects/dredd-example/i, 
3=/Users/matthewl/projects/dredd-example/c, 
4=/Users/matthewl/projects/dredd-example/b, 
5=/Users/matthewl/projects/dredd-example/d, 
6=/Users/matthewl/projects/dredd-example/e, 
7=/Users/matthewl/projects/dredd-example/f, 
8=/Users/matthewl/projects/dredd-example/g, 
9=/Users/matthewl/projects/dredd-example/j, 
10=/Users/matthewl/projects/dredd-example/k, 
11=/Users/matthewl/projects/dredd-example/l, 
12=/Users/matthewl/projects/dredd-example/m, 
13=/Users/matthewl/projects/dredd-example/n
```

michalholasek:
Fixed by https://github.com/apiaryio/dredd/pull/891.

PR #891, opened 2017-10-11
fix: Hook files are not always executed in alphabetical order (#890)

matthewl-biomatters:
Just a simple mistake with the compare function.

I'm not sure if adding extra hook files for testing is actually helpful?  On my machine they are the smallest number of files needed to trigger the bug but now the bug is fixed they are kind of messy.

### DIFF
--- a/src/add-hooks.coffee
+++ b/src/add-hooks.coffee
@@ -115,7 +115,11 @@ addHooks = (runner, transactions, callback) ->
         # ]
         .map((filepath) -> basename: basename(filepath), path: filepath)
         # Sort 'em up
-        .sort((a, b) -> a.basename > b.basename)
+        .sort((a, b) -> switch
+          when a.basename < b.basename then -1
+          when a.basename > b.basename then 1
+          else 0
+        )
         # Resolve paths to absolute form. Take into account user defined current
         # working directory, fallback to process.cwd() otherwise
         .map((item) -> path.resolve(customConfigCwd or process.cwd(), item.path))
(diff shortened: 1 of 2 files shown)


matthewl-biomatters:
The test failure seems unrelated so created as #892

honzajavorek:
Thanks very much for the PR! 👍 🚀 I'm sorry for the delay in response - there has been some vacations on our end. @michalholasek has been implementing this recently, so he'll take care of the review.

michalholasek:
Hello @matthewl-biomatters, we solved the issue with failing Swagger test you filed in https://github.com/apiaryio/dredd/issues/892 - could please rebase on top of current master to see if the tests pass? Thank you.

michalholasek:
I looked into the failing test and it seems to be related to `./test/fixtures/sandboxed-hook*.js` glob pattern resolving to `sandboxed-hook.js` and `sandboxed-hooks-log.js` files, but I am not sure why it is causing the timeout. When I delete the glob pattern and replace previously used `sandbox-hook*.js`files with for example `/foo/bar/hook1.js` and `/foo/bar/hook2.js` (non existent files), it passes 🤔 

@matthewl-biomatters, would you be willing to dig deeper and resolve the issue?

matthewl-biomatters:
Thanks @michalholasek I figured it was something like that but ran out of time yesterday.  Sorry I don't really have time to dig into it further but wonder if we could do this instead to avoid the risk of unrelated files that are add in future also weirdly breaking the tests.

`hookfiles: ['./**/*_hooks.*', '/baz/x.js', '/foo/y.js', '/bar/z.js', '/foo/a.js', '/bar/b.js', '/baz/c.js', '/foo/o.js', '/bar/p.js']`

michalholasek:
@matthewl-biomatters Agree, please modify the test file and squash all commits into one. Thank you.

Issue #892, opened 2017-10-11
Test failing: CLI - Swagger Document When loaded from file When Swagger is loaded with warnings should print warning to stdout


matthewl-biomatters:
When I run `npm test` on `master` it fails with:

```
CLI - Swagger Document When loaded from file When Swagger is loaded with warnings should print warning to stdout:
     AssertionError: expected 'info: Beginning Dredd testing...\npass: GET (200) /machines duration: 44ms\ncomplete: 1 passing, 0 failing, 0 errors, 0 skipped, 1 total\ncomplete: Tests took 51ms\n' to include 'warn: Parser warning'
```

And it's failing in my unrelated PR (#891).

honzajavorek:
Thanks! The tests are indeed failing for a different reason than your PR. It's probably related to failing master in Dredd Transactions - https://github.com/apiaryio/dredd-transactions/pull/108#issuecomment-337252277 No fear 😎! Special Judicial Squad member @michalholasek is on his mission to figure it out.

michalholasek:
Fixed by https://github.com/apiaryio/dredd/pull/900.

PR #895, opened 2017-10-16
chore: update inquirer to version 3.3.0

### LINKED ISSUE #768: Upgrade inquirer
The `inquirer` library drives the `dredd init` interactive initialization wizard and AFAIK that part of Dredd is undertested, so it needs manual verifying if changed. Also it's major version number change, so breaking changes can be expected when upgrading.

It would be nice if tests for the `dredd init` are added before upgrading.

realityking:

The old `inquirer` version pulls down a lot of old dependencies and prevents deduplication.

Fixes #768

Tested `dread init` manually with a couple of variations, everything works fine.

- [X] To write docs - N/A
- [X] To write tests - N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "glob": "^7.0.5",
     "html": "^1.0.0",
     "htmlencode": "0.0.4",
-    "inquirer": "^1.1.0",
+    "inquirer": "^3.3.0",
     "js-yaml": "^3.8.3",
     "markdown-it": "^8.3.1",
     "optimist": "^0.6.1",


realityking:
As far as I can tell the failure is not directly related to `inquirer`. Any guess why it fails?

honzajavorek:
@realityking Thanks for the PR! 🚀 I'm sorry for the delay in response - there has been some vacations on our end.

The problem with `inquirer` is that we do not have much tests for the interactive config logic. If you [check the coverage](https://coveralls.io/github/apiaryio/dredd), you can see even if tests were passing, it gives us little confidence to approve the upgrade:

image

So either the logic needs to be tried out manually to be sure nothing got broken (especially given the upgrade is 2 major and 2 minor versions) or we need to test it. Any help with that would be much appreciated.

-----

The tests are indeed failing for a different reason. It's probably related to failing master in Dredd Transactions - https://github.com/apiaryio/dredd-transactions/pull/108#issuecomment-337252277 Special Judicial Squad member @michalholasek is on his mission to figure it out.

michalholasek:
Hello @realityking , we solved the issue with failing Swagger test https://github.com/apiaryio/dredd/issues/892 - could please rebase on top of current master to see if the tests pass? Thank you.

realityking:
Saw the issue being closed :) I’m traveling right now, I’ll rebase ASAP.

realityking:
@michalholasek @honzajavorek Rebased and tests pass.

I did run through the interactive mode a couple of times, choosing different options. I can't claim that this was comprehensive as I don't know all the option that there are. If you want some more through testing I'll have to take a look at the code and see what paths I might have missed.

honzajavorek:
@realityking I think it shouldn't be hard to go through it. It is isolated in the [interactive-config.coffee](https://github.com/apiaryio/dredd/blob/master/src/interactive-config.coffee) file, and I can see only a couple of things to pay attention to:

- Check the Inquirer's interface - `inquirer.prompt(questions).then(callback)` But since you tried it out and it worked, I guess that's okay.
- Look at the `when:` functions and see whether they need any changes.

I didn't check the Inquirer's roadmap so I don't know if the breaking changes were due to dropping old Node support or whether they were actually changing something important. If you confirm you gave it a scan and you think it's all right, I'll merge this 👍

realityking:
Good point on actually checking the changelog 😆 

Here's the changelog for 2.0

- Answer `name` with `.` are now expand in the answer object. e.g. `name.sub` would generate an answer object formatted as `{name: {sub: 'answer'}}`.
- `expand` prompt default value if not specified is now `help`.
- Drop support for node 0.10 and 0.12

and the one for 3.0

- Drop support for Node < 4
- Allow failing async validation by rejecting the returned Promise
- Fix bug with bottom bar

The node requirements are obviously harmless and `expand` is not used. I think we're safe (unless there are bugs 😉)

honzajavorek:
Thanks! All looks good! Except for

> Drop support for Node < 4

Dredd still supports Node 4. We plan to support it until its maintenance end on 2018-04-01. However, it seems they dropped the support only because of their own maintenance convenience, not because they actually write code incompatible with Node 4. I'm willing to risk it. If it stops working, we can always pin/downgrade the version.

honzajavorek:
And looking into the CI config, the Node 4 version is even [still there](https://github.com/SBoudrias/Inquirer.js/blob/master/.travis.yml#L4) 🙂

realityking:
I think you missed the `<` in 

> Drop support for Node < 4

But yeah, the changelog doesn't make much sense 😄

honzajavorek:
!!! You are right 😄

Issue #225, opened 2015-06-17
Running dredd with --name flag on large .apib files causes RangeError: Maximum call stack size exceeded


carrollgt91:
Steps to reproduce:
- Download the following gist as an .apib file: https://gist.github.com/carrollgt91/11c2447410002be8d5b3
- Run `$ dredd /path/to/480_endpoints.apib api.datafi.local --names`

I think this happens because each step in iterating over the endpoints in that file is actually synchronous, which blows up the call stack when you hit the `async.mapSeries` call in the `run` method of [transaction_runner.coffee](https://github.com/apiaryio/dredd/blob/master/src/transaction-runner.coffee). That being said, I'm not sure that's the case; that's just what the stack trace leads me to believe.

Here's the [stack trace](https://gist.github.com/carrollgt91/b292db16a80b56248190) if you're interested.

netmilk:
Hey @carrollgt91! Thank you very much for perfect report. I confirm I can replicate it and it's doing the same.

petebrowne:
I've run into similar performance issues on large apib files. If I use a glob, performance is fine

``` bash
$ dredd apibfiles/**/* test.local --names
# ...
complete: 0 passing, 0 failing, 0 errors, 0 skipped, 269 total
complete: Tests took 86ms

# compiled.apib is the same set of apib files, concatenated into a single large file
$ dredd compiled.apib test.local --names
# never finishes...
```

PR #901, opened 2017-10-24
Limit stack depth to avoid maximum call stack exceeded issues

jasonk000:

It might be helpful for people suffering from maximum call stack exceeded issues eg #225 

#225 could be fixed, I haven't tested but it at least solves my issues.

Tests break as many assume synchronous returns.
If someone else wants to pick it up and run with it they are welcome to.

### DIFF
--- a/src/transaction-runner.coffee
+++ b/src/transaction-runner.coffee
@@ -148,7 +148,7 @@ class TransactionRunner
           runHookCallback()
 
       async.timesSeries hooks.length, runHookWithData, ->
-        callback()
+        process.nextTick(() -> callback())
 
     else
       callback()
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "dredd",
-  "version": "0.0.0-semantically-released",
+  "name": "@iflix/dredd",
+  "version": "4.3.0-iflix1",
```
honzajavorek: I understand this was necessary for you to use your fork, but it cannot be accepted to upstream 🙂

honzajavorek:
Looking at the commits, could you please:

- remove 7a94ba071b4cb071dced3afc80c8abc6806618ca
- reword e0829bc69ccc1ee6f2de325ec4781f444d14347e so it's prefixed with `fix:`? (`fix: limit stack depth...`)

Thanks! That should fix the failing tests and allow us to get this auto-released when merged.

jasonk000:
Thanks @honzajavorek - I've made the requested changes.

Generally you can apply this pattern (`process.nextTick` to execute callbacks) to break up stack flows between async function returns if this issue crops up elsewhere as well.

Issue #899, opened 2017-10-22
Dredd fails if --output dir does not exists


alejandropg:
#### Describe your problem

When you specify the output file inside a directory that does not exist, Dredd fail in the boot time. You can not use a hook to create the dir because the fail is previous to the hooks execution.
The directory path should be created, or at least forced whit an option.

#### What command line options do you use?

```
$ dredd --reporter junit --output non-existing-dir/dredd.xml
```

#### What is in your `dredd.yml`?

```yaml
blueprint: api-description.apib
endpoint: 'http://localhost:8080'
path: []
```

#### What's your `dredd --version` output?

```
dredd v4.5.0 (Darwin 16.7.0; x64)
```

#### Does `dredd --level=debug` uncover something?

No. The output is:

```
info: Configuration './dredd.yml' found, ignoring other arguments.
warn: Parser warning in file 'api-description.apib': (warning code 13) HTTP header 'Authorization' has no value on line 91
warn: Parser warning in file 'api-description.apib': (warning code 13) HTTP header 'Authorization' has no value on line 215
info: Beginning Dredd testing...
fs.js:653
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/Users/alex/src/globalia/ila/fake-countries/fake-countries-rest/non-existing-dir/dredd.xml'
    at Object.fs.openSync (fs.js:653:18)
    at Object.fs.writeFileSync (fs.js:1300:33)
    at Object.fs.appendFileSync (fs.js:1346:6)
    at appendLine (/usr/local/lib/node_modules/dredd/lib/reporters/x-unit-reporter.js:154:15)
    at EventEmitter.<anonymous> (/usr/local/lib/node_modules/dredd/lib/reporters/x-unit-reporter.js:47:9)
    at emitTwo (events.js:130:20)
    at EventEmitter.emit (events.js:213:7)
    at Dredd.emitStart (/usr/local/lib/node_modules/dredd/lib/dredd.js:291:39)
    at /usr/local/lib/node_modules/dredd/lib/dredd.js:139:26
    at /usr/local/lib/node_modules/dredd/lib/dredd.js:282:16
    at /usr/local/lib/node_modules/dredd/node_modules/async/dist/async.js:421:16
    at iteratorCallback (/usr/local/lib/node_modules/dredd/node_modules/async/dist/async.js:998:13)
[...]

PR #907, opened 2017-11-09
Create output directory if non-existent

jceplaras:

Filed in https://github.com/apiaryio/dredd/issues/899. Create directories if non-existent.

Resolved https://github.com/apiaryio/dredd/issues/899

- [x] To write docs
- [x] To write tests

### DIFF
--- a/test/unit/reporters/x-unit-reporter-test.coffee
+++ b/test/unit/reporters/x-unit-reporter-test.coffee
@@ -5,10 +5,11 @@ proxyquire = require('proxyquire').noCallThru()
 
 loggerStub = require '../../../src/logger'
 fsStub = require 'fs'
-
+mkdirpStub = sinon.spy((path, cb) => cb())
 XUnitReporter = proxyquire '../../../src/reporters/x-unit-reporter', {
   './../logger' : loggerStub,
   'fs' : fsStub
+  'mkdirp' : mkdirpStub
 }
 
 describe 'XUnitReporter', () ->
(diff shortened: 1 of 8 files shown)


test/unit/reporters/x-unit-reporter-test.coffee
```diff
@@ -65,10 +66,11 @@ describe 'XUnitReporter', () ->
     afterEach () ->
       fsStub.appendFileSync.restore()
 
-    it 'should write opening to file', (done) ->
+    it.only 'should write opening to file', (done) ->
```
honzajavorek: Forgotten `.only` 🙂

jceplaras: Will fix! Sorry about that. :)

jceplaras:
@honzajavorek Already addressed requested changes and also added an integration test for a non-existent directory.

honzajavorek:
I'm afraid we have a little blocker here: https://github.com/substack/node-mkdirp/pull/138

honzajavorek:
I'm sorry I suggested `mkdirp` in the beginning. I should have inspect the quality of the package first and not to suggest the first thing npm search offered me.

I think [make-dir](https://github.com/sindresorhus/make-dir) should be fine, [fs-extra](https://github.com/jprichardson/node-fs-extra) as well. The latter includes many common utilities we could use in Dredd in other places as well - especially in the `./scripts`, which we would like to eventually keep in form of JavaScript rather than bash, for Windows compatibility. @michalholasek which one would you prefer? Do you think `fs-extra` is a smarter choice in this case? I've seen you already using `rimraf` somewhere.

michalholasek:
@honzajavorek Both look good to me, but [fs-extra](https://github.com/jprichardson/node-fs-extra) is more "futureproof" (featurewise) for us.

Issue #888, opened 2017-10-02
Misleading reporter name - JUnit/XUnit


felpel:
#### Describe your problem

I'm trying to integrate Dredd in my continuous integration. In my configuration, if I specify `junit` as the reporter, it will output correctly the report in XML format. Although, if we look at the code in the repository, it is listed as `XUnitReporter`, which seems odd to me. It might be for backward compatibility, but I'm sure it will be confusing for new adepts like me.

#### What command line options do you use?

```
$ dredd
```

#### What is in your `dredd.yml`?

In order to make it work, I configured it this way:
```yaml
# other properties skipped
reporter: [junit]
output: [./reports/report.xml]
```
Although, I think it should be this way:
```yaml
# other properties skipped
reporter: [xunit]
output: [./reports/report.xml]
```

#### What's your `dredd --version` output?

```
dredd v4.4.0 (Windows_NT 10.0.14393; x64)
```

#### Does `dredd --level=debug` uncover something?

Not related to the command line call.

#### Can you send us failing test in a Pull Request?

The test outcome will simply not be reported in the `reports` folder.

honzajavorek:
Thanks for noticing this! I think this is clearly a typo/mistake, which unfortunately bubbled up to the public interface. I think we should rename it to xunit everywhere (also everywhere in docs). For backward compatibility, we should accept junit as an option as well though, at least for some time.

PR #908, opened 2017-11-09
Added support for reporter xunit parameter

jceplaras:

`xunit` is the better parameter for reporter rather than `junit`. `junit` is still supported until it is decided to be removed (https://github.com/apiaryio/dredd/issues/888#issuecomment-337265911).

Resolves https://github.com/apiaryio/dredd/issues/888

- [x] To write docs
- [x] To write tests


honzajavorek:
When this gets merged, we should create a new issue "Remove junit from codebase" so we don't forget about it. It could be easily done with the next major version.

PR #914, opened 2017-11-13
Non obvious link added to the docs for swagger

jdcrecur:

- [x] To write docs
- [x] To write tests


honzajavorek:
I think I'll solve the problem in a different way - 2e74abc, but thanks for pointing out this is a problematic part for the reader. Feedback on broken docs is always very valuable! 👍

honzajavorek:
@jdcrecur Would you mind reading the new docs and letting me know if it's better now?

- https://dredd.readthedocs.io/en/latest/how-it-works.html#choosing-http-transactions
- https://dredd.readthedocs.io/en/latest/how-to-guides.html#multiple-requests-and-responses

jdcrecur:
Awesome, much clearer thanks!

PR #940, opened 2017-12-18
Throw error when additional path doesn't exist

### LINKED ISSUE #378: If additional files are given in --path and they don't exist, Dredd silently ignores them
Is this bug or feature? In case Dredd should ignore them, it should be documented and there should be at least warning logged. In case Dredd should fail, then it does not and it's a bug.

``` shell
$ ls
apiary.apib
$ dredd ./apiary.apib http://localhost:8000 --path=./non-existent.apib
...
$ echo $?
0
```

honzajavorek:

Dredd should treat non-existing files (or globs evaluating to no files) as user error. This PR just finalizes @jgjadaoag's amazing work in https://github.com/apiaryio/dredd/pull/909.

We want to have this done as soon as possible so there are no open Pull Requests and we could decaf the codebase - #705.

- Fixes #378
- Closes https://github.com/apiaryio/dredd/pull/909

- [ ] ~~To write docs~~
- [x] To write tests

### DIFF
--- a/src/add-hooks.coffee
+++ b/src/add-hooks.coffee
@@ -1,7 +1,6 @@
 require 'coffee-script/register'
 path = require 'path'
 proxyquire = require('proxyquire').noCallThru()
-glob = require 'glob'
 fs = require 'fs'
 async = require 'async'
 clone = require 'clone'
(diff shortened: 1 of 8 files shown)


src/add-hooks.coffee
```diff
@@ -101,30 +101,38 @@ addHooks = (runner, transactions, callback) ->
...
-    )
+    try
+      files = hookfiles.reduce((result, unresolvedPath) ->
+        # glob.sync does not resolve paths, only glob patterns
+
+        unresolvedPaths = if glob.hasMagic(unresolvedPath) then glob.sync(unresolvedPath) else
+          if fs.existsSync(unresolvedPath) then [unresolvedPath] else []
+
+        if unresolvedPaths.length == 0
+          throw new Error("Hook file(s) not found on path: #{unresolvedPath}")
```
michalholasek: I don't like the fact that `throw new Error(...)` is used inside a `reduce` method. I would rather seen it rewritten in `forEach` or some imperative equivalent.

honzajavorek: I added some commits. Not sure if I actually addressed this comment though, because the resulting flow avoids reduce, but now the `throw` still happens in a `.map()` iteratee.

I think throwing exceptions isn't anything a pure function couldn't do. It's not mutation, it's just one of the possible outcomes of the pure function.

PR #943, opened 2018-01-06
CoffeeScript on NPM has moved to "coffeescript"

polc:

It just remove a warning at dredd installation : `warning dredd > coffee-script@1.12.6: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)`

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "caseless": "0.12.0",
     "chai": "4.0.2",
     "clone": "2.1.1",
-    "coffee-script": "1.12.6",
+    "coffeescript": "1.12.6",
     "cross-spawn": "5.0.1",
     "dredd-transactions": "5.0.4",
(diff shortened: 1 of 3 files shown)


honzajavorek:
Hi, thanks for the contribution! However, we're just in the process of migrating the whole project from CoffeeScript to JavaScript - #705, so ideally this would wait after that's done. CoffeeScript is still necessary as a dependency, because it's supported as a first-class citizen in JS hooks, but I would rather separate those two changes (language migration & package migration).

polc:
Great ! Thanks you for your work

honzajavorek:
@polc Would you mind to rebase this with current master? We're JavaScript now! 🎉 https://github.com/apiaryio/dredd/issues/705#issuecomment-363105927

polc:
Rebased !

Issue #754, opened 2017-03-20
Support for Node 8 when it becomes LTS


honzajavorek:
Based on https://github.com/apiaryio/dredd/pull/750 and https://github.com/snikch/goodman/pull/22#issuecomment-281798027, there could be issues with Dredd on Node 7. It would be nice to add Node 7 to the testing matrix in Travis CI and to bump also various different places to mention 7 instead of 6 (e.g. AppVeyor builds).

Anyone willing to contribute this?

honzajavorek:
According to [this document](https://github.com/nodejs/LTS) it's better to wait for Node 8 to become LTS and then to switch.

honzajavorek:
I tried to build Dredd on Node 8, it fails - https://travis-ci.org/apiaryio/dredd/jobs/292155038 Needs more work than just adding to the build matrix.

honzajavorek:
Update: Everything works except some nuances in the sandboxed hooks (specificially, the `test/unit/transaction-runner-test.coffee` file). The Pitboss library doesn't seem to support Node 8 correctly, but I am yet to verify that. Two possible solutions:

- Fix [Pitboss NG](https://github.com/apiaryio/pitboss)
- Remove Pitboss in favor of pure [vm](https://nodejs.org/api/vm.html)

honzajavorek:
Node 8 is already in `.travis.yml` for some time.

PR #946, opened 2018-01-11
Add Node 8, drop Node 4

honzajavorek:

- Node 8 is the latest even production-ready version and we should be sure Dredd supports it correctly.
- Node 4 end of life is April 2018 and there is no reason in supporting it anymore.
- We should not rely on loose dependency versioning - because there's no guarantee the interface, license, etc. won't change in minor or patch version. Semantic versioning is a nice idea, but not everyone follows it. Let's turn on [Greenkeeper](https://greenkeeper.io/) for this repo and let's pin all dependencies. Also adding `package-lock.json` as that's the standard way to work with `npm@5` now.

- Closes https://github.com/apiaryio/dredd/issues/754
- https://github.com/apiaryio/pitboss/issues/13

- [x] To write docs
- [x] To write tests


honzajavorek:
I suspect a0c49c6 to fix the original issue with sandboxed hooks, but to break a lot of other things 😞 I'll continue later with this.

honzajavorek:
@michalholasek If you like the changes, leaving up to you whether you want to merge this now or cherry-pick the commits at the end of your decaff PR and merge those together when the thing is ready.

michalholasek:
@honzajavorek I am ok with merging it before decaff branch is done, just one note - shouldn't we consider this a breaking change, since we are dropping Node 4 support?

honzajavorek:
@michalholasek It is a breaking change:

image

Issue #894, opened 2017-10-13
Add support for Open API 3.0


johngluckmdsol:
It appears that version 3.0 is not yet supported.  We're hoping to move to Open API 3.0 because we are better able to represent our JSON Schema with it.  It would be great if your tool could suppor it.

honzajavorek:
Regarding the Open API 3.0, we plan to support it, the question is when and the answer is yet to be determined. Of course anyone can come and contribute the support at any time, and we'll be there to help them - see https://github.com/apiaryio/dredd/issues/837#issuecomment-320020909

Related issues: https://github.com/apiaryio/fury-adapter-swagger/issues/131

JurRutten:
+1

veewee:
+1

XVincentX:
Same here. I'd love to put the express-gateway API under tests but our specs are written in Open API 3.0!

tombevers:
+1

jonas-depop:
+1

HvonP-QLIK:
+1

honzajavorek:
@tombevers @jonas-depop @JurRutten @veewee We understand this is a wanted feature, but next time please use [GitHub Reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) to express your feelings. Thanks! ✨

honzajavorek:
Just to reiterate, adding support for a new API description format goes like this...

Dredd consumes [API Elements](https://api-elements.readthedocs.io/en/latest/). It uses [Fury.js](https://github.com/apiaryio/fury.js) to turn any API description format (currently API Blueprint and Swagger 2.0) into API Elements.

To produce API Elements for various formats, Fury.js uses [adapters](https://github.com/search?q=topic%3Afury+org%3Aapiaryio&type=Repositories). So the first and the most important step to add support for OpenAPI 3.0 to Dredd would be to create an adapter for Fury, similar to [the one for Swagger](https://github.com/apiaryio/fury-adapter-swagger). The Fury.js maintainers (@apiaryio/adt) would definitely give a hand and advice to anyone who starts with this.

-------

**Note:** A scheme of Dredd's parsing stack is in https://github.com/apiaryio/dredd/issues/820, I'd be happy to re-draw it and add it to docs if there's someone who actually needs it.

altcatalin:
+1

ringgelerch:
+1

YGeorge:
Any update on this?

[...]

PR #950, opened 2018-01-17
Fix reference to the OpenAPI Specification

earth2marsh:
The specification format formerly known as Swagger is now the OpenAPI Specification

None

- [x] To write docs
- [ ] To write tests

Since this is a docs-only fix, no tests or linting required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ documentation.
 ### Supported API Description Formats
 
 - [API Blueprint][]
-- [Swagger][]
+- [OpenAPI (fka Swagger)](https://github.com/OAI/OpenAPI-Specification)
 
 ### Supported Hooks Languages
 


honzajavorek:
Hi, there are hundreds of references to Swagger throughout the docs, so if we want to change it, let's change it everywhere.

I'm wondering whether we want to do this at all, but I'm open to discussion. My thoughts: the format is still very well known and searched for as "Swagger". I'm aware of the renaming, but so far I decided to go with the naming like this:

- Reference the 2.0 version of the format consistently as Swagger
- Reference the 3.0 version of the format consistently as OpenAPI

The latter isn't in effect as #894 isn't done yet. When Dredd supports both, I'd call the formats Swagger 2.0 and OpenAPI 3.0, to clearly distinguish them. What do you think?

honzajavorek:
Hi, thanks again for bringing it up. I filed a separate issue for discussing this: https://github.com/apiaryio/dredd/issues/994

Issue #765, opened 2017-04-07
Redesign CLI reporter


honzajavorek:
Dredd's CLI reporter, which is what most users see as its default output, has several flaws. You can list some of them using the <kbd>[CLI reporter](https://github.com/apiaryio/dredd/labels/CLI%20reporter)</kbd> label. The aim of this meta-issue is to design a new reporter.

### Requirements / Internal Redesign

-   Use [debug](https://www.npmjs.com/package/debug) for Dredd application logging. Separate logging from the CLI reporter completely.
    -   Output debugging info only if asked.
    -   Fixes brievity of the dot reporter (https://github.com/apiaryio/dredd/issues/593), the `--names` option, and other parts of Dredd.

### CLI Reporter Redesign

-   Print short progress information first, use ✔ or number in case of a failure. Make the progress information very brief. Something like:

    ```
    Machines > Machines collection > Create Machines
      → POST /path/to/something (application/json)
      ← 200 (application/json)
      ✔ PASS ... 234ms

    Machines > Machines collection > Get Machines
      → GET /path/to/something?limit=34
      ← 200 (application/json)
      1) FAILURE ... 34ms
    ```

    If possible (terminal support, `--color` / `--no-color`), use colors (green, red) to denote pass or failure. The requested URL displayed in the output is the real one, with templates filled, with all changes applied, not the one from API description (https://github.com/apiaryio/dredd/issues/405). Every test starts with a transaction name (https://github.com/apiaryio/dredd/issues/430).

- List failures at the end, reference by failure numbers.
- Each failure in the list of failures contains:
    - Transaction name
    - Details about the request made
        - Method, headers, body used for the request
        - Filename of the API description document with `:line-number` at the end (clickable on some terminals, can be opened in an editor)
        - _Idea: Information (yes / no) whether the transaction was modified in hooks or whether it was left intact as specified in the API description_
        - _Idea: Debugging hint which `curl` command would perform exactly the same HTTP request as Dredd did during testing._
    - Details about the failure and explanation of the error. All common errors (JSON Schema validation errors, unparseable JSON errors, APIB/Swagger parse errors and warnings, ...) should be presented in such way that user of Dredd clearly has an idea on what happened, what is the problem and **what to do next**. (See examples of unreadable failed tests in https://github.com/apiaryio/dredd/issues/745, https://github.com/apiaryio/dredd/issues/665)
[...]

PR #969, opened 2018-02-21
Update dependencies that were effectively downgraded & remove some unused dependencies

realityking:

In the move to fixed dependency versions (f0f200b) some dependencies were effectively downgraded as there was a newer, in-range, version at the time.

While looking at the dependencies, I also found some unused dev dependencies:
* `coffee-coverage` isn't used anymore after decaffeinating

Many more things that are outdated but this seem like obvious wins 🙂

- [ ] To write docs
- [ ] To write tests


honzajavorek:
Thanks! This is definitely a good idea. It will take me some time to check licensing etc. of the packages, will track the progress here:

- [x] async@2.6.0
- [x] chai@4.1.2
- [x] cross-spawn@5.1.0
- [ ] glob@7.1.2
- [ ] js-yaml@3.10.0
- [ ] markdown-it@8.4.1
- [ ] proxyquire@1.8.0
- [ ] uuid@3.2.1
- [ ] which@1.3.0
- [ ] winston@2.4.0
- [ ] eslint-config-airbnb-base@11.3.0

honzajavorek:
@realityking I think once the `eslint-config-airbnb-base` is sorted, we're good to go, but _without_ winston for now. With regard to #765 and the long-term intention to replace winston with something like [debug](https://github.com/visionmedia/debug) I'm honestly a bit lazy to proceed with the upgrade internally right now.

realityking:
Thanks @honzajavorek! I pushed a new version without the bump for `winston`.

As for `eslint-config-airbnb-base`, that's already a dependency through `eslint-config-airbnb` but I'm happy to split this into another PR.

honzajavorek:
> I'm happy to split this into another PR

Let's wait just a little bit, my hunch would be this one could get resolved in a timely manner. It's Airbnb, they could care about licensing 😄

ljharb:
Note that the "license" field in package.json, which is SPDX-compliant, is *more* than sufficient, the actual LICENSE file is not required. (that said, I've merged https://github.com/airbnb/javascript/pull/1746)

honzajavorek:
@ljharb Thanks! 👍 I answered to your comment under airbnb/javascript#1746.

realityking:
Based on the above discussion I removed the `eslint-config-airbnb` since I assume you'll want to update to a version that includes the License file.

I'll open a separate PR to track that.

ljharb:
With or without the license file, the dependency is MIT licensed; there should be no need to remove it.

honzajavorek:
@realityking btw I'm sorry for the failing Windows builds, they got very flaky recently. I plan to focus on fixing them, but until that time I'll keep retrying the builds for you 😕

realityking:
@honzajavorek Looks like this all passes now :)

realityking:
@honzajavorek Would it be possible to release a new minor version that includes this PR? 🙂

honzajavorek:
Ha, I haven't noticed it's classified as `chore:`. If it was `feat:` or `fix:`, it gets autoreleased. Hmm...

realityking:
Ah, good to know. Now I wish I had made #979 a `fix:` - I was debating that 😆

realityking:
I had another PR I wanted to make this week - let me do that now and I'll put it in as a fix :)

honzajavorek:
Yeah, I think `fix` fits better. Maybe I can find time to do [upgrades on Dredd Transactions](https://github.com/apiaryio/dredd-transactions/pulls) and to release a new Dredd with those propagated.

honzajavorek:
Decided not to block on DT, this resolves the issue: https://github.com/apiaryio/dredd/pull/980

PR #973, opened 2018-02-22
fix(reporters): Use correct context after faulty CoffeeScript conversion

boennemann:

Both the html and markdown reporters are currently broken with `TypeError: Cannot read property 'level' of undefined at title`.

This is caused by a faulty coffeescript conversion where a fat arrow function was converted to a regular function without taking into effect that `@level`/`this.level` would then be broken.
Using an arrow functions brings back the old behavior.

-

- [ ] To write docs
- [ ] To write tests


boennemann:
🎉

PR #975, opened 2018-02-24
Update aslant and related dependencies

realityking:
#### :rocket: Why this change?

eslint and related dependencies are out of date. This updates them to their newest version. Part of this PR was originally part of #969.

This probably can't be merged until there's been a new release of `eslint-config-airbnb-base` (see #969 for discussion)

Other changes:
- Switch from `eslint-config-airbnb` to `eslint-config-airbnb-base`. This project is not using the React related parts and this drops some indirect dev dependencies.
- Fix new style issues due to the update to `eslint-config-airbnb-base`
- Disable `prefer-destructuring`. The rule is now part of `eslint-config-airbnb-base` but this didn't seem worth fixing. (Could always be done at a later time)
- Fix use of the `Buffer` constructor, also a new error from `eslint-config-airbnb-base`

#### :white_check_mark: What didn't I forget?

- [ ] To write docs
- [ ] To write tests
- [X ] To put [Conventional Changelog](https://dredd.readthedocs.io/en/latest/contributing/#sem-rel) prefixes in front of all my commits and run `npm run lint`

PR #979, opened 2018-03-02
chore: Don't use the deprecated Buffer constructor

realityking:

The `Buffer` constructor is deprecated (it will likely start throwing warnings in Node.js 10), instead `Buffer.{from|alloc|allocUnsafe}` should be used. 

Discussion of the API change: https://medium.com/@jasnell/node-js-buffer-api-changes-3c21f1048f97
Runtime deprecation planning: nodejs/node#19079

This is also preparation for #975.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -104,7 +104,7 @@ function applyConfiguration(config) {
   configuration.options.method = configuration.options.method.map(method => method.toUpperCase());
 
   if (configuration.options.user) {
-    const authHeader = `Authorization:Basic ${new Buffer(configuration.options.user).toString('base64')}`;
+    const authHeader = `Authorization:Basic ${Buffer.from(configuration.options.user).toString('base64')}`;
     configuration.options.header.push(authHeader);
   }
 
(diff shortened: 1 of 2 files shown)


ChALkeR:
Uh-oh. `dredd` apparently sends a chunk of uninitialized memory when `configuration.options.user` is a number (it's read from a yml config afaik). Doesn't seem to be serious due to use-cases, but still not a nice thing.

realityking:
@ChALkeR Only if `configuration.options.user` is a number. I assume it's a string so it'll be this behaviour: https://nodejs.org/api/buffer.html#buffer_new_buffer_string_encoding

realityking:
@honzajavorek The tests on Node 8 failed. Could you restart that job?

honzajavorek:
@realityking Just did :)

honzajavorek:
> if `configuration.options.user` is a number

Can we just always coerce it to a string [in the configuration](https://github.com/apiaryio/dredd/blob/master/src/configuration.js#L106)?

realityking:
> Can we just always coerce it to a string in the configuration?

That'd be possible. Behaviour with the patch as it is would be to throw the error `TypeError: "value" argument must not be a number`.

PR #980, opened 2018-03-05
Release updated dependencies

honzajavorek:
Because https://github.com/apiaryio/dredd/pull/969 was mistakenly classified as `chore`, it wasn't auto-released. This is an empty commit to trigger Semantic Release.


realityking:
Alternatively you could merge #982 which I classified as a fix.

PR #992, opened 2018-03-24
Decaffinate some test scripts

realityking:

Removing some more cofffeescript bits & pieces.

Main advantage is that eslint lints these new JS files while nothing was linting the coffeescript files.

- [ ] To write docs
- [ ] To write tests


honzajavorek:
Amazing, thanks! 👍  The CoffeeScript files used to be linted by coffeelint, but it has been removed.

Issue #952, opened 2018-01-18
Feature request: Quieter CLI reporter


mlambley:
I'm not sure whether this will be its own feature or part of https://github.com/apiaryio/dredd/issues/765 so I'm just logging it here for now.

In https://github.com/apiaryio/dredd/issues/765 there are separate sections for Progress and List of Failures.

What I'd like is to have the option to hide the List of Failures section from the CLI reporter. The reason is that my expected/actual json is quite long, and when dumped into my terminal it destroys all meaningful information there.

What I'd prefer is to be able to use both the html and cli reporters, have the failures dumped into the html reporter (as it is now) but make the cli reporter only show the Progress and not the List of Failures.

I had a few thoughts. One is to create a new reporter with the failure data omitted. I created one locally and can submit a PR if you wish.

Another is to modify the parameters in the cli-reporter.js - such as changing the variable inlineErrors to displayErrors which would be either none, inLine or atEnd. But since this would be a breaking change it might be better to incorporate it as part of https://github.com/apiaryio/dredd/issues/765.

Another is to have a means whereby a dev can create their own custom reporters. This might involve configuring a directory for the system to scan and dynamically load all reporter js files in it. I am yet to do this but could develop and submit a PR for this if you wish. However since you are considering switching to TAP this may be redundant.

honzajavorek:
Hi, wouldn't the [built-in dot reporter](https://github.com/apiaryio/dredd/issues/765#issuecomment-293612257) combined with html/apiary reporter solve your issue?

mlambley:
Dot reporter still displays failed tests anyway, so the issue of my long json cluttering up the terminal isn't abated by that.

Also, I like seeing pass/fail/skip next to each action/url/duration. I still want the CLI reporter, just without the clutter.

mlambley:
I've gone ahead and created a PR for the least intrusive option. Please give feedback. Thank you.

PR #1005, opened 2018-04-27
feat: Option for hiding errors in the cli.

mlambley:
Create a non-breaking option which will override inline-errors such that errors can be hidden in the cli altogether.
Useful to reduce clutter in the cli, if another reporter (eg. html or apiary) is reporting the errors.

To reduce clutter in the CLI when dealing with large API specifications.

https://github.com/apiaryio/dredd/issues/952

- [X] To write docs
- [X] To write tests


honzajavorek:
Thanks! It might take us a while to go through this. First let's get the Windows build passing - just needs retries, we have some flaky tests 😞

pksunkara:
@honzajavorek What's the status of this?

honzajavorek:
@pksunkara I'd be interested in what is exactly the use case, because if we need this, I think there's more stuff wrong with Dredd's output. I've been delaying this because I wanted to revisit #765 until the end of this year.

PR #1025, opened 2018-06-07
fix: Update dependencies

pksunkara:


honzajavorek:
@pksunkara I think this won't pass, because as far as I remember updating Mocha on Dredd gets the Windows builds running forever, for some reason 😞

PR #1029, opened 2018-06-13
fix: update Gavel

jesperronn:
This updates Gavel to most recent release 2.1.2 which does not make use of the `git+https` protocol, since it not supported for old Git versions, and therefore blocking the use of Dredd on ci servers with old Git versions (Git 2.1.2 affected, but Git 2.3 is good)

Gavels update is a Major release, which is breaking because it stopped supporting Node 4. But since this project requires at least Node 6, we should be good.

More details found in https://github.com/apiaryio/gavel.js/pull/102

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "dredd-transactions": "6.1.3",
     "file": "0.2.2",
     "fs-extra": "5.0.0",
-    "gavel": "1.1.1",
+    "gavel": "^2.1.2",
     "glob": "7.1.2",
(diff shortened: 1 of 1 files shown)


honzajavorek:
Oh! You're fast 😄 I already created https://github.com/apiaryio/dredd/pull/1028. It seems the new Gavel causes Dredd tests to fail for some reason. I need to investigate on why is that.

jesperronn:
Thanks I close this in favor of #1028

PR #1082, opened 2018-07-08
Remove automated compiling of CoffeeScript

realityking:

This change removes automated compiling of CoffeeScript and instead lets users add any kind of compiler, this would, for example, be useful to support Babel. It also remove the dependency on coffeescript which is nice for us non-coffeescript users :)

This will require a major version release.

- [X] To write docs
- [x] To write tests

I didn't add any new tests as the feature is covered with existing tests that use coffeescript.


honzajavorek:
It might take me a while to review and think through all the consequences, but I love this! One thing which is sure is this needs a `BREAKING CHANGE: ` section in the commit as it is a breaking change for anyone using the CoffeeScript hooks.

honzajavorek:
> I didn't add any new tests as the feature is covered with existing tests that use coffeescript.

Are there dedicated tests to verify hooks work with CoffeeScript as such, or do we rely on the rest of the tests using not-yet-decaffeinated hooks?

Since we have both CS and Babel in the (dev)deps, it would be nice to have two dedicated e2e test cases to be sure at least those two compilers are supported and tested.

package.json
```diff
@@ -71,6 +70,7 @@
     "babel-core": "6.26.0",
     "babel-preset-env": "1.6.1",
     "body-parser": "1.18.3",
+    "coffeescript": "1.12.7",
```
honzajavorek: Once we have it in the tool chain only to verify `--require=coffeescript/register` works, we could actually upgrade it to 2.x to get rid of the annoying warnings. But that's just a note for future from the top of my head, nothing to care about now.

src/dredd.js
```diff
@@ -104,6 +105,16 @@ https://dredd.readthedocs.io/en/latest/how-it-works/#using-https-proxy
...
 
+    if (this.configuration.options.require) {
+      let mod = this.configuration.options.require;
+      const abs = fs.existsSync(mod) || fs.existsSync(`${mod}.js`);
+      if (abs) {
+        mod = path.resolve(mod);
+      }
+      // eslint-disable-next-line global-require, import/no-dynamic-require
+      require(mod);
+    }
```
honzajavorek: Can we test and ensure whether Dredd is able to gracefully handle errors here?

- Dredd aborting with a sane message in case the path is wrong
- Dredd aborting with a sane massage in case the required module blows up (at least try/catch synchronously or something)

realityking:
I figured this would take a while to merge because it does make sense to batch it with some other breaking changes.

Currently there's no dedicated test to cover this functionality, I'll add some. (Might be a few days until I get to this)

package.json
realityking: Yeah I thought the same but I think this is better left for a separate PR.

honzajavorek:
> I figured this would take a while to merge because it does make sense to batch it with some other breaking changes.

I don't see an issue with that. I'm fan of continuous delivery. When debugging, it's easier to identify what change caused the problems, for both the maintainers and the users (they can try granular versions and say which is the last one working and it's clear what change the version brings).

Just add the `BREAKING CHANGE:` section to the end of the commit, with an explanation and possibly a migration hint. Once it's merged, it gets autoreleased and I'm totally fine with major bump.

> Might be a few days until I get to this

No problem, it took me quite a few days to get to reviewing this gem 💎, so I should be apologizing mysefl. Let's skip apologies and make awesome stuff when we have the time 😉

realityking:
Changes I've made:

- [x] Added error handling
- [x] Test for coffeescript
- [x] Test for babel-register
- [x] test for error case
- [x] added breaking change to the commit  

This leaves the sandbox case, I need to play with that and understand how it works.

realityking:
Hmm, tests are failing on Travis but working locally. Any idea what would cause this?

honzajavorek:
I see some [linter errors](https://travis-ci.org/apiaryio/dredd/jobs/415206816), but I guess that's not the real issue. Both AppVeyor and Travis agree there's a problem with

```
  Specifying neither response body nor schema in the API Blueprint
    when the server returns non-empty responses
TypeError: Cannot read property 'code' of null
    at runDredd (/home/travis/build/apiaryio/dredd/test/integration/require-test.js:67:28)
    at recordLogging (/home/travis/build/apiaryio/dredd/test/integration/helpers.js:141:7)
    at fn (/home/travis/build/apiaryio/dredd/test/integration/helpers.js:35:5)
    at configuration.emitter.emit (/home/travis/build/apiaryio/dredd/src/dredd.js:29:3501)
```

on multiple Node versions. Doesn't look flaky to me. Let me see if I can find a while to reproduce or debug this. For now I'd focus on looking for discrepancies between your local setup and the CI setup.

realityking:
I discovered that the difference is in whether the test suite is called directly (`npm run test`) or with coverage (`npm run test:coverage`).

I tried the naive thing of replacing Istanbul with nyc, but that just got me a new set of errors.

I think this is the point where I have to give up, the test suite (and dredd itself) are too complex. I'll leave this open in case you wanna take a stab at it, otherwise feel free to close it.

honzajavorek:
I’m sad to read that, but I suppose it’s true. The tests are way too complex sometimes, especially when taken as the whole suite.

I’ll definitely leave this open and hopefully it gets into my pipeline soon.

honzajavorek:
I'd like to revive this soon. I know I neglected it for quite some time, so I'm ready to take over the responsibility of rebasing etc., but any help is appreciated 🙏

realityking:
I took care of the rebasing and new eslint errors, I hope that makes things a little easer for you. The error from https://github.com/apiaryio/dredd/pull/1082#issuecomment-412450512 is still present.

realityking:
Looks like I actually got it work 😳 

The errors now are only from coveralls. I'll see sometime later this week if things still work when I go back to Istanbul, otherwise I'll see what's up with nyc.

honzajavorek:
I'm actually wondering whether the current setup with Coveralls makes any sense. Coveralls don't report back under PRs and I wasn't able to debug it. Without that integration, it's just a useless number on a README badge, which requires a lot of complexity in how we run tests.

Also, after [discussing the usefulness of coverage](https://twitter.com/Stephen_Mizell/status/1076162260677152768) with @smizell on Twitter I have heretic thoughts about removing coverage altogether...

realityking:
I pretty much rewrote half the changes but it wasn't too bad 😆 

As for coverage, if you wanna keep it around, I had a much better time with codecov than with coveralls.

honzajavorek:
I think I've considered these as alternatives in the past, but didn't finish the evaluation:

- https://www.codacy.com/
- https://codeclimate.com/
- https://codecov.io/

realityking:
It’s green 😭🥳

That leaves the problem of sandboxes hooks. I saw the proposal to simple rip them out (#1178). Is that an option?

honzajavorek:
👏👏

Already in action: https://github.com/apiaryio/dredd/commit/d0a2957

realityking:
Perfect!

honzajavorek:
Here it is: https://github.com/apiaryio/dredd/pull/1196

realityking:
Curios, why didn‘t you make this PR part of 6.0?

honzajavorek:
I forgot it’s breaking change as well 🤦‍♂️

Let's do it as a next step then. We can rebase on top of current master and make another release, I’m not against it.

Actually, I should drop the “major release ceremony” sentiment altogether. With a release for every change people will have version granularity when upgrading and it’s easier for them to debug problems on the way.

I’ll release this early next week.

realityking:
Rebased on top of master.

PR #1094, opened 2018-07-20
Verify whether Dredd handles binary data

honzajavorek:

This is to close the <kbd>[Epic: empty responses and binary files](https://github.com/apiaryio/dredd/labels/Epic%3A%20empty%20responses%20and%20binary%20files)</kbd> epic. Since the empty responses have been fixed, it is (probably!) already possible to test binary data with Dredd or to easily skip the testing. However, this isn't verified in Dredd's test suite and there are no examples in the docs.

#### 🚧 To Do

- [x] test Dredd with "higher bytes", images, etc. https://github.com/apiaryio/dredd/pull/836 and with non-JS hooks https://github.com/apiaryio/dredd/issues/617
- [x] implement encoding of bodies in case they're binary (or a different solution)
- [x] refactor and unit test the implementation
- [x] unit test the performRequest itself
- [x] move the API Blueprint multipart fix to Dredd Transactions https://github.com/apiaryio/dredd-transactions/pull/177
- [x] move multipart unit tests to Dredd Transactions
- [x] get all tests passing
- [x] write docs with examples

- Closes https://github.com/apiaryio/dredd/issues/617
- Closes https://github.com/apiaryio/dredd/issues/87
- Deprecates https://github.com/apiaryio/dredd/pull/836

- [x] To write docs
- [x] To write tests


src/performRequest.js
```diff
@@ -0,0 +1,161 @@
...
+  const httpOptions = Object.assign({}, options.http || {});
+  httpOptions.proxy = false;
+  httpOptions.followRedirect = false;
+  httpOptions.encoding = null;
+  httpOptions.method = transactionReq.method;
+  httpOptions.uri = uri;
+
+  try {
+    httpOptions.body = getBodyAsBuffer(transactionReq.body, transactionReq.bodyEncoding);
+    httpOptions.headers = setContentLength(transactionReq.headers, httpOptions.body);
```
michalholasek: I think that `setContentLength` is a little bit misleading here, `normalizeHeaders` or something like that would be more fitting in this case.

honzajavorek: It doesn't interfere with any other headers. It detects the name of the content-length header if it exists, and sets it to the correct value. In case it doesn't exists, it adds the header. IMHO the name is right.

src/performRequest.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+  const lcEncoding = encoding.toLowerCase();
+  if (lcEncoding === 'utf-8' || lcEncoding === 'utf8') {
+    return 'utf-8';
+  }
+  if (lcEncoding === 'base64') {
+    return 'base64';
+  }
+  throw new Error(`Unsupported encoding: '${encoding}' (only UTF-8 and `
+    + 'Base64 are supported)');
```
michalholasek: Nit: I would personally use `switch` instead `if`(s) wherever possible, something like:
```javascript
switch (lcEncoding) {
  case 'utf-8':
  case 'utf8':
    return 'utf-8'
  case 'base64';
    return 'base64';
  default:
    throw new Error(`Unsupported encoding: '${encoding}' (only UTF-8 and `
      + 'Base64 are supported)');
}
```
But it is not an issue, just an idea.

src/performRequest.js
```diff
@@ -0,0 +1,161 @@
...
+  // hooks implementations.
+  return body.toString().includes('\ufffd') ? 'base64' : 'utf-8';
+}
+
+
+performRequest.normalizeBodyEncoding = normalizeBodyEncoding;
+performRequest.getBodyAsBuffer = getBodyAsBuffer;
+performRequest.setContentLength = setContentLength;
+performRequest.createTransactionRes = createTransactionRes;
+performRequest.detectBodyEncoding = detectBodyEncoding;
```
michalholasek: When we are exposing private/internal APIs for testing purposes, I think it would be good idea to prefix the names with leading underscores `__` - see [this](https://github.com/facebook/react/blob/d17091ddebf317173eb30e3ab4a8e28e124c8f50/src/umd/ReactWithAddonsUMDEntry.js#L22) for inspiration 🙂

honzajavorek: lol (inspiration). I did some research and dropped some thoughts on this to https://github.com/apiaryio/dredd-transactions/pull/179#discussion_r206897044. Let's continue there.

src/performRequest.js
honzajavorek: 🤔

src/performRequest.js
honzajavorek: I'll trust you this is more _javascriptonic_  😄 Python [doesn't have `switch`](https://www.quora.com/Why-does-python-not-have-a-switch-statement) so I never think of using it, even in situations where it probably makes sense.

honzajavorek: ae58682

src/performRequest.js
honzajavorek: 1f6c251

honzajavorek:
@michalholasek Addressed your comments. Would you mind looking at the additional commits and saying 👍 or 👎 ?

PR #1085, opened 2018-07-09
Update request to 2.87

realityking:
#### :rocket: Why this change?

Looks like the licensing conflicts are solved according to #1059.

#### :memo: Related issues and Pull Requests

Closes #1059

#### :white_check_mark: What didn't I forget?

- [ ] To write docs
- [ ] To write tests
- [x] To put [Conventional Changelog](https://dredd.readthedocs.io/en/latest/contributing/#sem-rel) prefixes in front of all my commits and run `npm run lint`

Issue #1052, opened 2018-06-19
sinon cannot be upgraded to 6.0.0 (license issues)


honzajavorek:
The issues were detected using the [license-checker](https://github.com/davglass/license-checker) tool:

### sinon@6.0.0
 - License: BSD-3-Clause
 - Copyright Notice: Copyright (c) 2010-2017, Christian Johansen, christian@cjohansen.no
 - Dependencies: 13
 - Description: JavaScript test spies, stubs and mocks.
 - Package: https://npmjs.com/package/sinon
 - Repo: https://github.com/sinonjs/sinon

### Problematic Licenses: 1
 - `just-extend@1.1.27` (MIT) - missing full license text

honzajavorek:
This has been closed already! 🙀 Closing now.

PR #1107, opened 2018-09-03
Update dependencies

honzajavorek:

- After merging https://github.com/apiaryio/dredd/pull/1085 the lock file is out of sync
- Upgrade sinon

- https://github.com/apiaryio/dredd/pull/1085
- #1052

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -117,6 +117,30 @@
         }
       }
     },
+    "@sinonjs/commons": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.0.2.tgz",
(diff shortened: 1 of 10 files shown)


package-lock.json
```diff
@@ -1995,7 +1995,7 @@
         },
         "lodash": {
           "version": "4.17.4",
-          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.4.tgz",
+          "resolved": "http://registry.npmjs.org/lodash/-/lodash-4.17.4.tgz",
```
kylef: Some of these changes don't seem right, moving away from https? Did you perhaps use `npm config set registry http://...` when switching from internal registry?

honzajavorek: I don't think so 🤔 I've seen similar changes in the past, not sure why this happens.

PR #1136, opened 2018-11-23
Simplify links

honzajavorek:

Some pages are really hard to orientate in, especially because a large number of long links targeting various specs. This PR introduces one extension and shared `_links.rst` file to tackle the problem.

- [x] To write docs
- [x] To write tests


docs/_extensions/specs.py
```diff
@@ -0,0 +1,72 @@
...
+SPECS = {
+    'apib': ('API Blueprint', 'https://apiblueprint.org/documentation/specification.html#{anchor}'),
+    'mson': ('MSON', 'https://apiblueprint.org/documentation/mson/specification.html#{anchor}'),
+    'openapi2': ('OpenAPI 2', 'https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#user-content-{anchor}'),
+    'openapi3': ('OpenAPI 3', 'https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#user-content-{anchor}'),
+}
+
+
+# https://docutils.readthedocs.io/en/sphinx-docs/howto/rst-roles.html
+def link(name, rawtext, text, lineno, inliner, options={}, content=[]):
```
kylef: Not usually a good idea to use `{}` and `[]` as default arguments to a function in Python. Advisable to use `None` and then handle it later in the body of the function:

```python
def foo(content=None):
    ...content or []
```

This is problematic because the the same instance of these instantiated array/dict will be passed as the default argument to link each time, any mutation will apply to further invoations of the function.

Take the following example:

```python
def addHello(content=[]):
    content.append('hello')
    return content
```

```python
>>> addHello()
['hello']
>>> addHello()
['hello', 'hello']
>>> addHello()
['hello', 'hello', 'hello']
```

docs/_links.rst
```diff
@@ -0,0 +1,33 @@
...
+
+.. OpenAPI specification
+
+.. _OpenAPI 2: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md
+.. _OpenAPI 3: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md
+
+
+.. JSON Schema
+
+.. _Draft 3: https://tools.ietf.org/html/draft-zyp-json-schema-03
```
kylef: Down to you, but I feel like these references are a little ambiguous `Draft 3` could mean anything. I think it would be better to name them with `JSON Schema` in the name.

docs/_extensions/specs.py
kylef: I also realise I think I let this slip through in separate PR in `ghissue`.

honzajavorek:
@kylef thanks! comments addressed

docs/_links.rst
```diff
@@ -16,13 +17,12 @@
...
-.. _Draft 5: https://tools.ietf.org/html/draft-wright-json-schema-00
-.. _Draft 6: https://tools.ietf.org/html/draft-wright-json-schema-01
-.. _Draft 7: https://tools.ietf.org/html/draft-handrews-json-schema-01
-
 .. _JSON Schema: https://json-schema.org
+.. _JSON Schema Draft 3: https://tools.ietf.org/html/draft-zyp-json-schema-03
+.. _JSON Schema Draft 4: https://tools.ietf.org/html/draft-zyp-json-schema-04
+.. _JSON Schema Draft 5: https://tools.ietf.org/html/draft-wright-json-schema-00
+.. _JSON Schema Draft 6: https://tools.ietf.org/html/draft-wright-json-schema-01
+.. _JSON Schema Draft 7: https://tools.ietf.org/html/draft-handrews-json-schema-01
```
kylef: If I understand correctly, only links to Draft 3 and 4 are used. Perhaps we can just delete these other ones? No use in unused code/links.

PR #1158, opened 2018-12-12
Align file names with the Airbnb style guide

honzajavorek:

Getting the codebase closer to JavaScript standards and further from its CoffeeScript past. The rule:

> [23.6](https://github.com/airbnb/javascript#naming--filename-matches-export)
> A base filename should exactly match the name of its default export.

⚠️ I did not tackle this one yet:

> [23.9](https://github.com/airbnb/javascript#naming--Acronyms-and-Initialisms)
> Acronyms and initialisms should always be all capitalized, or all lowercased.

- [ ] To write docs
- [ ] To write tests


bin/dredd
```diff
@@ -12,7 +12,7 @@ var path = require('path');
 
 
 var dir = process.env.COVERAGE_DIR ? 'src' : 'lib';
-var CLI = require('../' + dir + '/cli');
+var CLI = require('../' + dir + '/CLI');
```
kylef: > Acronyms and initialisms should always be all capitalized, or all lowercased.

Wouldn't this change be going in opposite direction? (although personally I dislike this rule)

src/HooksWorkerClient.js
```diff
@@ -4,10 +4,10 @@ const path = require('path');
 const spawnArgs = require('spawn-args');
 const { EventEmitter } = require('events');
 
-const getGoBin = require('./get-go-bin');
+const getGoBin = require('./getGoBin');
```
kylef: Maybe it would be nicer to rename this `getGoBinary`, it's a lot clearer on what it does.

test/unit/reporters/CliReporter-test.js
```diff
@@ -5,8 +5,8 @@ const { EventEmitter } = require('events');
 
 const loggerStub = require('../../../src/logger');
 
-const CliReporter = proxyquire('../../../src/reporters/cli-reporter', {
-  './../logger': loggerStub
+const CliReporter = proxyquire('../../../src/reporters/CliReporter', {
```
kylef: 💭 We're using `CLI` elsewhere so this becomes inconsistent (`CLIReporter`).

src/handleRuntimeProblems.js
```diff
@@ -1,4 +1,4 @@
-const blueprintUtils = require('./blueprint-utils');
+const blueprintUtils = require('./blueprintUtils');
```
kylef: Maybe it would be better to:

```js
const { warningLocationToRanges, rangesToLinesText } = require('./blueprintUtils');
```

Not gonna comment on the whole "utils" module :).

src/configureReporters.js
```diff
@@ -1,11 +1,11 @@
...
-const DotReporter = require('./reporters/dot-reporter');
-const HtmlReporter = require('./reporters/html-reporter');
-const MarkdownReporter = require('./reporters/markdown-reporter');
-const NyanCatReporter = require('./reporters/nyan-reporter');
-const XUnitReporter = require('./reporters/x-unit-reporter');
+const ApiaryReporter = require('./reporters/ApiaryReporter');
+const BaseReporter = require('./reporters/BaseReporter');
+const CliReporter = require('./reporters/CliReporter');
+const DotReporter = require('./reporters/DotReporter');
+const HtmlReporter = require('./reporters/HtmlReporter');
```
kylef: I know this is what the AirBNB rules say, but wish this rule wouldn't exist and it would be valid to use `HTMLReporter`.

bin/dredd
honzajavorek: Funny. So from the [examples](https://github.com/airbnb/javascript#naming--Acronyms-and-Initialisms) it looks like they understand uppercase as "capitalization", even though Wikipedia [describes it](https://en.wikipedia.org/wiki/Capitalization) in the sense in which you understood it. The reasoning,

> Why? Names are for readability, not to appease a computer algorithm.

also supports the idea that they mean uppercase. I think I will send a PR 😄

honzajavorek: https://github.com/airbnb/javascript/pull/1972

src/handleRuntimeProblems.js
honzajavorek: Once I get to rewrite this to use source maps with line numbers, I think I can get rid of the whole module.

PR #1162, opened 2018-12-18
Introduce smoke tests

honzajavorek:

I want to fiddle with how Dredd is built and packaged. From my very own experience, in 99% cases this leads to silly mistakes breaking the installability of the resulting package (discrepancies in `.npmignore` and `.gitignore`, wrong Babel setting, etc.). For this reason https://github.com/apiaryio/dredd/issues/528 was filed and the aim of this PR is to implement the proposed smoke test using an additional Travis CI job.

- Close https://github.com/apiaryio/dredd/issues/528

- [ ] To write docs
- [x] To write tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,60 @@
-# GH_TOKEN and NPM_TOKEN are expected to be set
-# in https://travis-ci.org/apiaryio/dredd/settings
-dist: "trusty"
 language: "node_js"
-node_js:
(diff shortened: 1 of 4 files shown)


honzajavorek:
I forgot to turn off CircleCI after playing with it. Please ignore the build error there, I believe I'd have to open a new PR to get rid of it now.

Issue #1159, opened 2018-12-15
[docs] Installation


floggle:
The install docs are woefully inadequate.
A simple "npm install dredd" doesn't work.

It seems that the node install needs to be in the system PATH or one of the requirements fails (protagonist).
Haven't done that, it's giving a big page of error that seems to equate to requiring python installed and in the path.
```
gyp ERR! configure error
gyp ERR! stack Error: Can't find Python executable "python", you can set the PYTHON env variable.
```
Although looking closer at the end, it seems to have worked.

Could I suggest improving the "installation" docs to make these requirements clearer, or otherwise not provide error messages of this sort when an optional dependency fails?

edit: And if I do provide a Python path, it still fails, this time with a SyntaxError!

honzajavorek:
@floggle Hi! Thanks for trying to use Dredd and thanks for the issue!

> The install docs are woefully inadequate.

Could you please be more specific which docs exactly you mean? I recently updated the [installation docs](http://dredd.org/en/latest/installation.html) to be as adequate and detailed as possible, but I might have miss something of course, or you may be referring to other docs (READMEs etc.). In that case I'd very much like to fill the gaps.

You are experiencing problems with the C++ dependencies, which have their [dedicated section](http://dredd.org/en/latest/installation.html#installing-c-dependencies).

Issue #1149, opened 2018-12-09
Native compilation failure of optional dep makes installation super slow


ianwremmel:
**Describe the bug**
Having Dredd present in a project using Node 10 leads to an attempt to natively compile protagonist (or drafter, or snowcrash, or something, I'm not quite sure which is actually the culprit). For reasons not entirely clear to me, this native compile is attempted not just when installing dredd, but when installing other modules (maybe anytime the dredd tree in package-lock gets touched?)

The issue [seems to be fixed in protagonist](https://github.com/apiaryio/protagonist/issues/199) but Dredd needs to update to consume it.

**To Reproduce**
Use Node 10. Run `npm install dredd`. You'll see a very long install time with a failed native compile (but a zero exit code). 

**Expected behavior**
Installs go quickly because the native module (which is an optional dependency) cannot install. 

**What is in your `dredd.yml`?**
 Not Relevant
**What's your `dredd --version` output?**

Same thing happens when installing 5.3.0 on heroku and on debian

```
dredd v5.3.0 (Darwin 18.2.0; x64)
```

**Does `dredd --level=debug` uncover something?**
Not relevant

**Can you send us failing test in a Pull Request?**
Not Applicable

honzajavorek:
Hi @ianwremmel, the API Blueprint parser (let's refer to all the projects with this name to make it easier) is delivered in two forms - a compiled parser written in C++, and an equivalent, slower, but more portable JS parser compiled from the same C++ code.

It's true the parser doesn't support Node 10 correctly in the version Dredd uses, but even if the compilation fails, the parser should work correctly as there's fallback to the JS form. Making the compilation pass won't speed-up the installation though, it only speeds up Dredd's parsing of API Blueprint documents.

I'd say that what you really want (and I really want as well) is to get https://github.com/apiaryio/drafter-npm/pull/16 finished so that Dredd installs with the JS form by default, which avoids any compilation when managing the lifecycle of your project with npm, and so that the C++ form is an opt-in for deployments of Dredd where speed of parsing of API Blueprint documents is more important than the speed of installation.

See also [docs on this topic](http://dredd.org/en/latest/installation.html#installing-c-dependencies). Pinging @apiaryio/adt so they know this is still an issue for Dredd users and we should prioritize it.

ianwremmel:
[...]

Issue #659, opened 2016-11-08
Installation of protagonist finishes with errors, which confuses users


borekb:
When installing Dredd 2.2.3 on Windows 10, I see these errors:

![image](https://cloud.githubusercontent.com/assets/101152/20079928/58d1d53a-a547-11e6-8293-1273a3dc2b54.png)

The installation still seems to work fine, interestingly (I am able to run `dredd` from the console). Also to note is that I tried on another Windows 10 PC and there, all was fine (no errors).

honzajavorek:
Thanks! My first idea would be following: Dredd internally uses [Drafter](https://github.com/apiaryio/drafter) to parse API Blueprint. There are two versions of Drafter - one faster in C++11 and one slower (but more portable) in pure JavaScript. The [npm package for drafter](https://github.com/apiaryio/drafter-npm/) decides which one to use based on success of installation of the C++11 version. If it fails, it continues to use the JavaScript version.

Your installation seems to be the case. However, usually people do not have C++ compiler at all, so the failure doesn't look so severe. My hypothesis would be that you have some kind of C++ compiler present in the system, but it's not able to compile C+11, so the C++11 drafter installation fails with some unusual, terrible hairy errors. At the same time, the Dredd installation continues, finishes (I can't say, because the very bottom of the output isn't on your screenshot) and Dredd then runs correctly.

@w-vi What do you think?

--------------

_Mandatory disclaimer: Dredd isn't officially supported on Windows yet (it's planned!) and especially non-JS hooks probably won't work for you. However, there are people happily using Dredd on Windows._

borekb:
For the record, on my laptop where it all works fine, I've got [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) installed.

pksunkara:
@borekb Can you make sure you have C++ installed from the latest visual studio community edition and try again?

borekb:
Should the C++ Build Tools linked above suffice? I'm hesitant to install the full Visual Studio (even the Express edition is large) on my PC.

w-vi:
@borekb I think so.

honzajavorek:
[...]

Issue #503, opened 2016-05-31
Dredd should not use protagonist at all


pksunkara:
It currently has protagonist as development dependency

honzajavorek:
It's there just because of [this test](https://github.com/apiaryio/dredd/blob/0b1721a889eb420b39b7ec68dbaf46b808f5b574/test/unit/blueprint-utils-test.coffee). I think the test could use Fury. Willing to contribute? :stuck_out_tongue_winking_eye:

argos83:
I'd love to have the protagonist dependency removed. I know it can be skipped when installing dredd with `--no-optional`. However there's no way to specify that from package.json (see https://github.com/npm/npm/issues/14185). My builds take much longer because of compilation. 

And currently in my local dev env with node 10, the compilation fails with:

```
../src/parse_async.cc:128:14: error: no matching constructor for initialization of 'v8::TryCatch'
    TryCatch try_catch;
```
So every time I try to install any node module (not just dredd), npm tries to re-compile protagonist over and over again.

honzajavorek:
@argos83 I'm sorry, this issue is about something slightly different. Previously, Dredd was using `protagonist` as a way to parse API Blueprint. This has changed with introducing `Fury.js` to the codebase, which encapsulates all parsing, for both API Blueprint and OpenAPI 2. This issue reminds us there's still one place left in the tests where Dredd depends on `protagonist` directly, which shouldn't be.

Actually, it is interesting you found the `dredd > gavel > curl-trace-parser > api-blueprint-http-formatter > protagonist` dependency path, I was not aware of that one. And it's strange, because installing Dredd with `--no-optional` does avoid any compilation, even though according to this dependency path it should still compile `protagonist` as a dependency of the `api-blueprint-http-formatter` 🤔 

Anyway, what you want is to avoid `protagonist` as a dependency completely to avoid parsing. That is not possible as of now. Currently it works as opt-out. A solution would be to help the @apiaryio/adt to finalize https://github.com/apiaryio/drafter-npm/pull/16, which installs `drafter.js` as default and makes the `protagonist` an opt-in. See also [Why Is the Installation So Slow?](https://dredd.readthedocs.io/en/latest/installation.html#why-is-the-installation-so-slow) and [Compiled vs pure JavaScript](https://dredd.readthedocs.io/en/latest/contributing.html#compiled-vs-pure-javascript).

[...]

PR #1168, opened 2018-12-19
Remove C++ dependencies using distributed npm-shrinkwrap.json

honzajavorek:

To mitigate all the various problems discussed in https://github.com/apiaryio/dredd/issues/1149.

- Close https://github.com/apiaryio/dredd/issues/659
- Close https://github.com/apiaryio/dredd/issues/1149
- Close https://github.com/apiaryio/dredd/issues/1159
- Close https://github.com/apiaryio/dredd/issues/503

- [x] To write docs
- [x] To write tests


opichals:
@kylef @pksunkara Considering how much people just hate to have this to be build again and again for no most of the time no benefit I'd like to suggest to consider dropping the whole `protagonist` inside `drafter.js` idea and add an extra package which would get picked up if installed as peer dependency (along with documenting the fact in README).

honzajavorek:
@opichals The plan is to have only the WASM version and no compilation https://github.com/apiaryio/dredd/issues/1149#issuecomment-448181314 What you're talking about is attempted here https://github.com/apiaryio/drafter-npm/pull/16, but it isn't as simple (I don't remember all the concerns @kylef expressed in previous discussions). I don't know what's the best solution for everyone for the interim period. I decided the best solution for Dredd is this one, and I only hope I won't get surprised by some side-effects I did not anticipated when this gets merged.

scripts/postshrinkwrap.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+// Postprocesses the lockfile to remove C++ dependency and to enforce https://
+
+const fs = require('fs');
+const path = require('path');
+const lockfile = require('../npm-shrinkwrap');
+
+// Force all installations of Dredd to use only the pure JavaScript version
+// of the API Blueprint parser. It has slower performance, but it solves
+// roughly a billions of installation & distribution problems.
```
kylef:
```suggestion
// roughly billions of installation & distribution problems.
```

Not quite billions either but ok 😄

scripts/postshrinkwrap.js
```diff
@@ -0,0 +1,21 @@
...
+delete lockfile.dependencies.drafter.requires.protagonist;
+delete lockfile.dependencies.protagonist;
+
+const json = JSON.stringify(lockfile, null, 2)
+  // Sometimes the npm registry resolves to http:// URLs for no obvious reason.
+  // There are multiple discussions and bugs about it at https://npm.community
+  // and it seems to be the registry's infrastructure problem. Responsible
+  // people suggested following as a safe workaround:
```
kylef: I have seen this too and was quite confused by it. This explains a lot!

test/unit/blueprintUtils-test.js
```diff
@@ -139,24 +142,82 @@ describe('blueprintUtils', () => {
...
-            }
-            allRanges.push(blueprintUtils.warningLocationToRanges(location, blueprint));
-          }
-          done();
-        });
-      });
+      for (const annotation of annotations) {
+        let location = [];
+        for (const sourceMap of annotation.attributes.sourceMap) {
+          location = location.concat(sourceMap.content);
```
kylef: What about using [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Examples) instead of imperitive loop/concat?

```js
const location = annotation.attributes.sourceMap.flatMap((sourceMap) => sourceMap.content);
```

scripts/postshrinkwrap.js
honzajavorek: I think the `a` got in because it was `a billion`, but then I changed it to `billions` 😂

test/unit/blueprintUtils-test.js
honzajavorek: Oh this is just unindented. I don't like the code, I could make a comment about every single line of it myself, but I don't want to go into rewriting it now, especially given I think I should be able to drop the whole file soon.

honzajavorek:
So I guess I'm merging this! Merry Christmas, everyone 🎄

scripts/postshrinkwrap.js
honzajavorek: Oh I forgot to commit this. Will do in a follow-up.

honzajavorek:
Well it still compiles locally when I do `npm install something --save` or `npm uninstall something --save` etc. Not sure how to get rid of that. But when I install Dredd from npm, it doesn't compile anymore. Except it takes eternity because of dev deps, but fix for that is on its way https://github.com/apiaryio/dredd/pull/1170

realityking:
One unfortunate side effect of this: it breaks deduplication of dependencies. All of dredds dependencies are kept around as a second copy even if they're already used somewhere else. This added ~1000 lines to the package-lock of our app.

I like @opichals suggestion. It plays much nicer with how people expect npm packages to behave.

PR #1173, opened 2018-12-20
Do not use 'console'

honzajavorek:

Currently the only warnings linter produces on Dredd's codebase are warnings about the `no-console` rule. The reasoning of the rule is following:

> In JavaScript that is designed to be executed in the browser, it’s considered a best practice to avoid using methods on console. Such messages are considered to be for debugging purposes and therefore not suitable to ship to the client. In general, calls using console should be stripped before being pushed to production.

While Dredd's code (currently) isn't destined to run in browsers, `console` is often used for debugging purposes (at least by lame people like me) and I see the benefit in having the linter warning me there's such call in the codebase. For that reason and also to follow a common convention, I think it is better not to use the `console` object directly in the production code.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/test/integration/cli/hookfiles-cli-test.js
+++ b/test/integration/cli/hookfiles-cli-test.js
@@ -201,7 +201,7 @@ describe('CLI', () => {
           // TCP server echoing transactions back
           const hookHandler = net.createServer((socket) => {
             socket.on('data', data => socket.write(data));
-            socket.on('error', err => console.error(err));
+            socket.on('error', err => process.stderr.write(`${err}\n`));
           });
 
           const args = [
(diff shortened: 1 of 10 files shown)


test/integration/cli/hookfiles-cli-test.js
```diff
@@ -201,7 +201,7 @@ describe('CLI', () => {
           // TCP server echoing transactions back
           const hookHandler = net.createServer((socket) => {
             socket.on('data', data => socket.write(data));
-            socket.on('error', err => console.error(err));
```
kylef: I wonder, is this as useful for debugging?

I thing coercing an error to string won't show stacktrace like `console.error` will so this might make it harder to debug tests. But it is late so I didn't dig into this, will proceed with review tomorrow.

honzajavorek: Hmmm, I didn't think of this behavior of `console.error`.

PR #1174, opened 2018-12-20
Remove unnecessary bitwise operators

honzajavorek:

Because `~` in combination with `.indexOf()` comes from the stone age.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/test/unit/reporters/MarkdownReporter-test.js
+++ b/test/unit/reporters/MarkdownReporter-test.js
@@ -80,7 +80,7 @@ describe('MarkdownReporter', () => {
 
     it('should write the title to the buffer', done =>
       emitter.emit('start', '', () => {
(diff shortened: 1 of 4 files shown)


test/unit/reporters/MarkdownReporter-test.js
```diff
@@ -80,7 +80,7 @@ describe('MarkdownReporter', () => {
 
     it('should write the title to the buffer', done =>
       emitter.emit('start', '', () => {
-        assert.isOk(~mdReporter.buf.indexOf('Dredd'));
+        assert.isOk(mdReporter.buf.includes('Dredd'));
```
kylef: I don't really know chai's assert that well, but should this be `assert.isTrue`?

PR #1183, opened 2019-01-04
Test the hooks example

honzajavorek:

The examples used in the docs should be tested. I introduced a new example in #1181. I'm experimenting with the idea of using Cucumber for specifying Dredd's end-to-end behavior and for testing it. Let me know what you think about this initial stub. Flaws:

- not tested on Windows
- ~this particular test doesn't necessarily need to be e2e~ https://github.com/apiaryio/dredd/issues/1424
- no coverage is collected during e2e tests

- https://github.com/apiaryio/dredd/pull/1181

- [x] To write docs
- [x] To write tests


features/step_definitions/hooks-js.js
```diff
@@ -0,0 +1,42 @@
...
+
+Given('an implementation, which requires authorization on "POST /articles"', function () {
+  this.dredd.args.push('--server=node test/fixtures/blog/app.js');
+});
+
+Given('I have hooks adding authorization to "POST /articles"', function () {
+  this.dredd.args.push(`--hookfiles=test/fixtures/blog/hooks${this.hooksExt}`);
+});
+
+When('I run Dredd', { timeout: 10 * 1000 }, function (callback) {
```
honzajavorek: Ideally, this wouldn't need to be an end-to-end test. We could call `new Dredd()`, then run it and capture what happens through a custom `EventEmmiter`, but I wasn't really in a mood to get it working now as Dredd's configuration through JavaScript is ugly. This is an initial proof-of-concept of this approach and we'll need e2e tests later as well anyway, so this code is going to be useful.

features/step_definitions/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  rules: {
+    'func-names': 'off',
```
honzajavorek: This is necessary to prevent ESLint warnings on step definitions, where it makes no sense to give names to every function and we need the `function` expression as we need to access `this`.

features/step_definitions/hooks-js.js
honzajavorek: Not possible now, because of https://github.com/apiaryio/dredd/issues/1424

features/step_definitions/hooks-js.js
```diff
@@ -0,0 +1,42 @@
+const { spawn } = require('child_process');
+const { expect } = require('chai');
+const { Given, When, Then } = require('cucumber');
+
```
honzajavorek: I want the step definitions to be like LEGO pieces one day, but I don't want to LEGO-ize them prematurely. I thought I'd start with this, where each step definition is very specific to the feature file in hand, and that we'd LEGO-ize and DRY **when needed** once we're adding more feature files.

features/step_definitions/hooks-js.js
```diff
@@ -0,0 +1,42 @@
...
+  ].concat(this.dredd.args);
+
+  const cli = spawn('node', args);
+  cli.stdout.on('data', (data) => { this.dredd.output += data; });
+  cli.stderr.on('data', (data) => { this.dredd.output += data; });
+  cli.on('exit', (exitStatus) => { this.dredd.exitStatus = exitStatus; callback(); });
+});
+
+Then('the "GET /articles" test passes', function () {
+  expect(this.dredd.output).to.contain('pass: GET (200) /articles');
```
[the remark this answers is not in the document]
honzajavorek: I'd like the tests to use Dredd.js instead of CLI (unless we're testing the CLI itself) and I even had a prototype of how it could work, but I bumped into https://github.com/apiaryio/dredd/issues/1424, so I reverted to end-to-end using the CLI. If Dredd.js was used, the test could listen on events or see the stats as an object. We need to do more decoupling to have the tests nicer.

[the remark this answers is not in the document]
honzajavorek: Added also https://github.com/apiaryio/dredd/issues/1428

Issue #1178, opened 2018-12-28
Remove sandboxed JS hooks


honzajavorek:
Dredd's [sandboxed JS hooks](https://dredd.org/en/latest/hooks/js-sandbox.html) are designed, as opposed to [standard JS hooks](https://dredd.org/en/latest/hooks/js-sandbox.html), to allow running arbitrary code. They provide limited feature set and they take security measures to be able to run the arbitrary code in hooks as a service.

I believe the main reason they were introduced to Dredd was the Post-Deployment feature in Apiary, where Dredd was meant to be ran as a service:

<img width="969" alt="screenshot 2018-12-28 at 17 49 22" src="https://user-images.githubusercontent.com/283441/50522093-72aaa100-0ac9-11e9-9dfa-57bdd6cb4724.png">

This feature never made it to general availability though. We had some further ideas about running Dredd as a service, but running hooks as a service was not part of them, at least in the initial phases.

The code behind sandboxed hooks in Dredd is rusted and makes development and understanding of Dredd's codebase harder. It uses [unmaintained technologies](https://github.com/mdp/pitboss). It is also harder to [make the JS hooks great again](https://github.com/apiaryio/dredd/labels/Epic%3A%20JS%20hooks) with sandboxed hooks in place. Basically if we ever follow up on the Post-Deployment feature or want to fix issues in Dredd's JS hooks, we need to rewrite everything from the ground up anyway.

As the only use case for sandboxed hooks I can imagine is running Dredd as a service, and as even Apiary is not taking advantage of this use case for years, I propose removing the sandboxed hooks from Dredd's codebase.

Issue #168, opened 2015-04-01
Confusing leading whitespace in transaction name when no group is given


netmilk:
When I have following blueprint:
```
FORMAT: 1A
# Machines API
# Machines collection [/machines]
## Get Machines [GET]
- Response 200 (application/json; charset=utf-8)

        [
          {
            "type": "bulldozer",
            "name": "willy"
          }
        ]
```

It results in this transcation name:
```
info:  > Machines collection > Get Machines
```

Issue #1058, opened 2018-06-19
proxyquire cannot be upgraded to 2.0.1 (license issues)


honzajavorek:
The issues were detected using the [license-checker](https://github.com/davglass/license-checker) tool:

### proxyquire@2.0.1
 - License: MIT
 - Copyright Notice: Copyright 2013 Thorsten Lorenz.
 - Dependencies: 6
 - Description: Proxies nodejs require in order to allow overriding dependencies during testing.
 - Package: https://npmjs.com/package/proxyquire
 - Repo: https://github.com/thlorenz/proxyquire

### Problematic Licenses: 1
 - `path-parse@1.0.5` (MIT) - missing full license text

realityking:
PR to fix the licensing issue https://github.com/jbgutierrez/path-parse/pull/2

Looks like the package is essentially unmaintained, I doubt it'll ever get merged. Instead I took a look what it'd take to drop the dependency all together. Not looking to good either: https://github.com/browserify/resolve/issues/168

realityking:
For tests yes, but proxyquire is also used to set up the hook files: https://github.com/apiaryio/dredd/blob/5cd3529a612913e72da71d8fa2d5ee330fb4eafb/src/add-hooks.js#L6

honzajavorek:
Oh, it injects the `hooks`. Hmm. Wondering whether there is an alternative solution to that 🤔

realityking:
I see two options:

1. Simply make hooks available as a global in the execution context. Pitboss supports this directly.
2. Inject a global variable like above with the actual logic but expose an entry point `dredd/hooks` that contains the definitions of all functions. This would have the sweet advantage of giving us working autocomplete in IDEs.

I like option 2 and would be happy to test that out but it would be a breaking change.

honzajavorek:
I like the option 2! We could do this as a separate change, or in the wider context of JS hooks revamp. The current implementation has [many issues](https://github.com/apiaryio/dredd/labels/Epic%3A%20JS%20hooks). If you found any of those worth your time, it would be totally awesome. I'm definitely eager to have breaking changes if they improve the software and people's lives 😄 

Just for the sake of consistency, let's see how other languages do it:

- Perl: `use Dredd::Hooks::Methods;`
- PHP: `use Dredd\Hooks;`
- Python: `import dredd_hooks as hooks`
- Ruby: `include DreddHooks::Methods`
- Rust: `extern crate dredd_hooks;`
- Go: well...
[...]

PR #1196, opened 2019-01-11
Remove sandboxed hooks and old APIs

honzajavorek:

January clean up :) Most of the reasons for these changes are already formulated in https://github.com/apiaryio/dredd/issues/1178.

- bc20b3c removes sandboxed JS hooks
- 4066e2e removes a very old fix for a very old bug from 2015, which I think isn't relevant to any Dredd users anymore - see #168
- 4e5b6c6 removes `blueprintPath` from the JS API, which seems like a remnant of some old APIs as well (I have to admit I found it only because I was searching the code base for the word "legacy")
- e0287ed removes an old API for `beforeAll` and `afterAll` JS hooks, which was already deprecated for ages and printed a big fat warning to anyone still using it

- https://github.com/apiaryio/dredd/issues/1178
- #168

- [x] To ~write~ delete 🙈 docs
- [x] To ~write~ delete 🙈 tests


realityking:
If you're doing a spring cleaning, there's one more candidate. The `junit` reporter is already deprecated. https://github.com/apiaryio/dredd/blob/125be8983ce9dc3c517703c0b7a1d4326f100c10/lib/configureReporters.js#L55

realityking:
It's amazing how much simpler the hooks will be when all of this is ripped out. After this and #1082 are merged I'll take a shot at #1058.

honzajavorek:
Not sure about _spring_ given the weather in Prague just covered me in snow, but otherwise you're right 😀

I’ll add a commit regarding junit later 👍

Maybe we could call it snow cleaning instead of spring cleaning, just saw this car

9c0feece-c6f4-44e1-9fab-6dcb117462fd

realityking:
😱 That's not nice. In Berlin it's just cold, grey and rainy. 🌧

honzajavorek:
Yes, after this I can almost imagine someone even contributing to make the JS hooks better 🙀

honzajavorek:
Ha ha, no problem. I’m glad we still get at least a few days of snow with all the global warming in action 🏜

lib/TransactionRunner.js
```diff
@@ -200,137 +183,14 @@ class TransactionRunner {
...
-      }
-    }
-
-    // Sandboxed mode - hook is a string - only sync API
-    if (typeof hook === 'string') {
-      this.runSandboxedHookFromString(hook, data, callback);
+    if (hook.length === 1) {
+      // Sync api
+      hook(data);
+      callback();
```
kylef: This reminds me how very simple it would be to support async/promise returning callbacks:

```js
const result = callback();

// Once Node >= 10 use https://nodejs.org/api/util.html#util_util_types_ispromise_value
// Or perhaps only support Promises if Node >= 8 or Node >= 10 to simplify code
if (/* node >= 8.2 where callbackify is existant? */ && result && result.then) {
  const {callbackify} = require('util');
  callbackify(result)(() => callback()); // shouldn't we support errors? I saw below async API doesn't?
} else {
  callback();
}
```

lib/addHooks.js
```diff
@@ -91,26 +34,8 @@ Stack: ${error.stack}
...
-`;
-      return callback(new Error(msg));
-    }
-    // No data found, doing nothing
-    return callback();
-  }
-
   // Loading hookfiles from fs
-  const hookfiles = [].concat(runner.configuration.options.hookfiles);
+  const hookfiles = [].concat(runner.configuration.options.hookfiles || []);
```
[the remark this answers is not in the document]
honzajavorek: I don't think there is such case in real world usage, but many existing tests didn't pass without it, because they don't properly init the configuration. So it seemed easier to do this fallback in the codebase than to debug all the (unit)tests and set hookfiles to empty array everywhere `configuration` is being used as an input 😞

lib/TransactionRunner.js
honzajavorek: My plans are to revamp the JS hooks in some future after this. As I commented above, I can even imagine someone contributing relevant changes to the JS hooks when this gets merged. My ideas around the JS hooks API would be:

- sync support (no callback parameter, no return value)
- async callback support (accepts callback)
- async promise support (returns promise)
- tested async/await support
- `transaction.fail = ...` gracefully fails the hook
- thrown error gracefully fails the hook (both sync and async)
- `callback(error)` gracefully fails the hook
- rejected promise fails the hook

Also thinking whether there's a way to make the hooks pure functions in the future, but I'd leave that aside for now as such change is a big one and sort of spans across the languages.

honzajavorek:
@kylef I added two more commits, but I don't want to sneak them in after you approved the PR. Will wait for you to rubber stamp it again.

PR #1209, opened 2019-01-15
Simplify code & reduce dependencies of file based reporters

realityking:

I was going to propose upgrading `fs-extra` to version 7 since now the Node.js version ranges match but while looking at the code I realized it's only used for one method - `mkdirp`. There's a smaller, more focused module (`make-dir`) which does the same job with less dependencies. Soon that module will make use of the native functionality present in Node.js 10.12+.

Similarly `file` (a dependency that's not been updated for 5 months) is only used for one methods, `abspath`. Most of it's functionality is actually in Node core (path.resolve), except for the expansion of `~` to the homedir. I added `untildify` for that purpose. I assume this was not an intended feature so it could be deprecated but for this PR I wanted to maintain compatibility.

- [ ] To write docs
- [ ] To write tests
- [X] To put [Conventional Changelog](https://dredd.org/en/latest/internals.html#sem-rel) prefixes in front of all my commits and run `npm run lint`


realityking:
Sorry about the test failure due to typos. The suite takes so long, I prefer to run it in the cloud if I can.

honzajavorek:
Nothing to be sorry about. It's the same for me :) I'd have ideas on how to make the test suite better / faster, it's just that I think it's better to focus on making the code better first while the tests stay intact. Can't change both at the same time.

realityking:
The last test failure is a bit of mystery to me. The code seems to work correctly if I add some `console.log` statements but ninon doesn't see it. Have you seen something like that before?

honzajavorek:
```
  1) MarkdownReporter
       when ending
         when can create output directory
           should write buffer to file:
     AssertionError: expected false to be truthy
      at Context.it (test\unit\reporters\MarkdownReporter-test.js:101:16)
```

A consistent failure across systems, but only in Markdown reporter? Wow, that's weird. Will check it out.

test/unit/reporters/MarkdownReporter-test.js
```diff
@@ -85,17 +87,17 @@ describe('MarkdownReporter', () => {
     describe('when can create output directory', () => {
       beforeEach(() => {
         sinon.stub(fsStub, 'writeFile');
```
honzajavorek:
```suggestion
        sinon.stub(fsStub, 'writeFile').callsFake((path, data, callback) => callback());
```

test/unit/reporters/MarkdownReporter-test.js
```diff
@@ -105,17 +107,17 @@ describe('MarkdownReporter', () => {
...
```
honzajavorek:
```suggestion
        sinon.stub(fsStub, 'writeFile').callsFake((path, data, callback) => callback());
```

test/unit/reporters/MarkdownReporter-test.js
```diff
@@ -85,17 +87,17 @@ describe('MarkdownReporter', () => {
...
+        makeDirStubImpl = sinon.spy(makeDirStubImpl);
       });
 
       afterEach(() => {
         fsStub.writeFile.restore();
-        fsExtraStub.mkdirp.restore();
+        makeDirStubImpl = makeDirStubImplBackup;
       });
 
       it('should write buffer to file', (done) => {
```
honzajavorek:
```suggestion
      it('should write buffer to file', done => emitter.emit('end', () => {
```

honzajavorek:
Seems like `init.js` uses `fs-extra` as well 🙀

Discovered thanks to the smoke tests. I didn't expect to realize their usefulness so soon 😄

realityking:
Damn. Not sure how I missed that one when I grepped for it. Let's see if it works with this fix.

PR #1099, opened 2018-08-01
Application logging / reporter output separation

michalholasek:
#### :rocket: Why this change?
This PR changes the way how Dredd handles logging information - 'application' logs are now written to stderr (instead of stdout), stdout is used for reporter's output only.
#### To-Do:
* [x] - Move original `logger.js` (Winston) to `reporters/logger.js`
* [x] - Create custom logger implementation using `debug` npm package
* [x] - `--loglevel|-l=debug|info|warn(ing)|error` (renamed from `--level`)
* [x] - Use `logger.debug` instead of `logger.verbose`
* [ ] - Switch Winston for custom logger
* [ ] - Allow command line debugging using `DEBUG=dredd|* dredd` command

#### :memo: Related issues and Pull Requests
https://github.com/apiaryio/dredd/issues/1089

#### :white_check_mark: What didn't I forget?

- [ ] To write docs
- [ ] To write tests
- [ ] To put [Conventional Changelog](https://dredd.readthedocs.io/en/latest/contributing/#sem-rel) prefixes in front of all my commits and run `npm run lint`

Issue #1089, opened 2018-07-16
Separate Dredd's application logging from subprocess and reporter output


honzajavorek:
Currently Dredd uses winston as it's main method of outputting to the command line. It has many flaws, primarily:

- It combines reporters output and application logging
- It outputs to both stdout and stderr inconsistently (certain levels go there, other ones go elsewhere) so it's difficult to parse or redirect
- It pollutes Dredd's output so it's cluttered, ruins brievity of the dot reporter (https://github.com/apiaryio/dredd/issues/593), the `--names` option, and other parts of Dredd
- The [`--level`](https://dredd.readthedocs.io/en/latest/usage-cli.html#level-l) option is inconvenient as it controls output of both application logging and reporter output
- It makes debugging difficult
- It blocks further development of reporters, namely #341 and #765 

The solution:

- Use [debug](https://www.npmjs.com/package/debug) for Dredd application logging. Update docs, tests, and the `.github/ISSUE_TEMPLATE.md` file.
- Use winston only in reporters. Replace it with `process.stdout.write()` later on and remove the `--level` option.
- Take care of capturing the server/hooks output alongside the ideas in https://github.com/apiaryio/dredd/issues/593#issuecomment-251946163 and let reporters to deal with it as they wish (probably a new event).

thesageinpilani:
@honzajavorek , I will be glad to implement the above functionality in dredd, with your approval. Can you please tell me about logging in reporters, so finally dredd won't be using winston at all?

honzajavorek:
@thesageinpilani I think this won't be easy to implement at all and will better be carried out by @michalholasek or me. I don't want to turn you down, I just do not expect this to be a nice experience and we could actually be able to figure it out ourselves.

I did review your previous PR and I'd be very grateful if you could work on that instead so we can fix all the issues mentioned there.

If this issue gets without addressing for 2 or 3 weeks, we failed to deliver the change on time and anyone is very welcome to contribute it if interested. But now I'd prefer if the maintainers could focus on the issue first.

thesageinpilani:
Thanks for the detailed review. I am grateful for guidance and support. 😀

honzajavorek:
@michalholasek one thought - looking at [this](https://github.com/visionmedia/debug#windows-command-prompt-notes) it might be beneficial to keep the `--level` option or rename it to `--debug` and treat it as if setting `DEBUG=dredd` to simplify this for people.

[...]

Issue #1101, opened 2018-08-06
New dredd init & config


honzajavorek:
This is a proposal on how to tackle the [dredd init + dredd.yml epic](https://github.com/apiaryio/dredd/labels/Epic%3A%20dredd%20init%20%2B%20dredd.yml).

### Parts of Dredd affected

- dredd init wizard
- configuration
- dredd.yml

### Requirements

- replace [optimist](https://github.com/substack/node-optimist) with something maintained https://github.com/apiaryio/dredd/issues/1057
- test and improve `dredd init`
  - add questions https://github.com/apiaryio/dredd/issues/320
  - improve wording and behavior https://github.com/apiaryio/dredd/issues/879
  - more robust detection https://github.com/apiaryio/dredd/issues/525
- test and improve generated CI configurations:
  - migrate to circleci 2.0 https://github.com/apiaryio/dredd/issues/878, https://github.com/apiaryio/dredd/issues/978
  - add wercker https://github.com/apiaryio/dredd/issues/878
  - revise travisci https://github.com/apiaryio/dredd/issues/878
  - make it extensible? https://github.com/apiaryio/dredd/issues/878
  - make all of them first-class citizen (AFAIK Dredd won't generate travis in favor of circleci now, only update existing travis config)
  - add commands for hooks https://github.com/apiaryio/dredd/issues/322
  - unify CI docs https://github.com/apiaryio/dredd/issues/353
- **bonus feature, maybe even a separate epic:** allow downloading of the API description document by Dredd directly https://github.com/apiaryio/dredd/issues/366
- refine the CLI and the options
  - allow overriding of blueprint and endpoint https://github.com/apiaryio/dredd/issues/455
  - rename certain outdated options (such as blueprint and endpoint), but keep backward compatibility (at least for a while)
  - test and fix bugs in hierarchical configuration https://github.com/apiaryio/dredd/issues/757, https://github.com/apiaryio/dredd/issues/915
  - revise https://github.com/apiaryio/dredd/issues/997
  - do not require server address for `--names` https://github.com/apiaryio/dredd/issues/563

### Solutions

- migrate to [yargs](https://www.npmjs.com/package/yargs) or something else https://github.com/apiaryio/dredd/issues/845
- use [nconf](https://www.npmjs.com/package/nconf)

honzajavorek:
Depends on https://github.com/sindresorhus/lcid/issues/3

honzajavorek:
- `--hooks=...` (files)
[...]

PR #1213, opened 2019-01-17
Separate "logging" from "output"

honzajavorek:

Because it's needed if we want to revamp the reporters and make them better. See https://github.com/apiaryio/dredd/issues/1089

- Closes https://github.com/apiaryio/dredd/pull/1099
- Closes https://github.com/apiaryio/dredd/issues/1089
- Enables https://github.com/apiaryio/dredd/issues/765

#### 🚧 To do

- [x] make two separate [Winston](https://github.com/winstonjs/winston) logger instances, one for Dredd's "output" and the other for Dredd's "logging"
- [x] remove --level and --silent, replace it with --loglevel and --loglevel=silent, don't allow the option to control Dredd's "output"
- [x] remove all "logging" levels except of error, warning, debug, and silent
- [x] go through the codebase once again and carefully decide about every logger call whether it's "output" or "logging"
- [x] have existing tests passing
- [x] write unit tests for the --loglevel behavior (error, warning, warn, debug, silent, lower/upper case, default is warning)

- [x] To write docs
- [x] To write tests


honzajavorek:
The docs link checker is failing because David is down. Please, don't block the PR on it.

screenshot 2019-01-24 at 19 24 09

test/integration/cli/configuration-cli-test.js
```diff
@@ -54,25 +54,21 @@ function execCommand(custom = {}, cb) {
 describe('CLI class Integration', () => {
   before(() => {
     ['warn', 'error'].forEach((method) => {
-      sinon.stub(loggerStub, method).callsFake((chunk) => { stderr += `\n${method}: ${chunk}`; });
+      sinon
+        .stub(loggerStub, method)
```
kylef: I think it would be nicer if the loggers we're dependency injected into the reporters. It will make it easier to adapt and simplify the testing removing the need for stubbing.

test/integration/cli/hookfiles-cli-test.js
```diff
@@ -533,7 +533,7 @@ describe('CLI', () => {
       });
     });
 
-    describe('when setting the log output level with -l', () => {
+    describe('when setting the log output level with --level', () => {
```
kylef: If I understand correctly, you are changing the test for clarity and the old argument (`-l`) is still supported?

Are we removing test coverage for the `-l` option here? Perhaps worth having tests for both -l and --level

lib/options.json
```diff
@@ -92,21 +92,16 @@
...
-    "default": "info"
+    "description": "The level of logging to output. Options: 'debug', 'warning', 'error', 'silent'.",
+    "default": "warning"
   },
   "timestamp": {
     "alias": "t",
     "description": "Determines whether console output should include timestamps.",
     "default": false
   },
-  "silent": {
```
kylef: Wondering, would be possible / make sense for us to detect this argument still and map it to log level and provide a deprecation warning to the user? Should make it simpler for users to upgrade and understand why some arguments no longer work.

test/unit/Dredd-test.js
```diff
@@ -91,7 +90,7 @@ describe('Dredd class', () => {
         configuration = {
           server: 'http://127.0.0.1:3000/',
           options: {
-            silent: true,
```
kylef: NIT, it would be easier to review if the commit that removed the option also removed it from the code. It's harder to review when things like level is removed but a test using level was updated in separate commit etc. Mostly feedback for consideration in future PRs I don't expecyou to change anything at this point.

Not to mention this separation would make it problematic for anyone to use git-bisect on this project as tests may not pass on some commits.

lib/CLI.js
```diff
@@ -80,7 +80,7 @@ Example:
 
       if (this.exit) {
         this._processExit = (exitStatus) => {
-          logger.verbose(`Exiting Dredd process with status '${exitStatus}'.`);
+          logger.debug(`Exiting Dredd process with status '${exitStatus}'.`);
```
kylef: 🤔 is this log message useful to anyone? Dredd exiting with a status code should already be pretty clear.

lib/CLI.js
```diff
@@ -89,19 +89,19 @@ Example:
         };
       } else {
         this._processExit = (exitStatus) => {
-          logger.verbose(`Exiting Dredd process with status '${exitStatus}'.`);
+          logger.debug(`Exiting Dredd process with status '${exitStatus}'.`);
```
kylef: This line is duplicated and since it contains a log message perhaps it would make sense to use the same string to prevent any potential for typos in messages only if Dredd is used with certain options.

Perhaps something like the following closure and using some form of composition to use it with the other two closures.

```js
const logExit = exitStatus => logger.debug(`Exiting Dredd process with status '${exitStatus}'.`);
```

Outside the scope of this PR but just a thought while reading.

lib/CLI.js
```diff
@@ -156,13 +156,13 @@ Example:
 
     // Show help
     } else if (this.argv.help === true) {
-      logger.silly('Printing help.');
+      logger.debug('Printing help.');
```
kylef: Are these messages really warranted? (same with pritning version, etc).

It should be pretty clear already

lib/HooksWorkerClient.js
```diff
@@ -57,7 +57,7 @@ class HooksWorkerClient {
   }
 
   terminateHandler(callback) {
-    logger.verbose('Terminating hooks handler process, PID', this.handler.pid);
+    logger.debug('Terminating hooks handler process, PID', this.handler.pid);
```
kylef: Some places we're adding fullstop and quotes around variables in the message and other times relying on the logger behaviour of multiple arguments. Should we have some consistency here?

```
logger.debug(`Exiting Dredd process with status '${exitStatus}'.`);
```

lib/reporters/ApiaryReporter.js
```diff
@@ -40,7 +40,7 @@ function ApiaryReporter(emitter, stats, tests, config, runner) {
 
   this.configureEmitter(emitter);
 
-  logger.verbose(`Using '${this.type}' reporter.`);
+  logger.debug(`Using '${this.type}' reporter.`);
 
   if (!this.configuration.apiToken && !this.configuration.apiSuite) {
     logger.warn(`
```
kylef: Should this message be a warning from the reporter? It is a reporter specific warning.

docs/usage-js.rst
```diff
@@ -56,7 +56,7 @@ Let’s have a look at an example configuration first. (Please also see the :ref
 
        'require': null, // String, When using nodejs hooks, require the given module before executing hooks
 
-       'no-color': false,
+       'color': true,
```
kylef: Unsure about the behaviour here, generally in CLI tools it is good idea to detect the default value based on if the shell is TTY (example at https://github.com/kylef/Commander/blob/314f8d7455e25e60e9e9b850040eebd0a87f2527/Sources/Commander/Error.swift#L42-L56, although in this case you should simply be able to use an `isatty` method instead of the TERM parsing, that works around Xcode specific logger that don't apply here).

That way if I do something like:

```
dredd > somefile
```

Colours will automatically be disabled and the written file won't include ANSI codes.

Of course, `--colour` and `--no-colour` would override from automatic behaviour.

test/integration/cli/hookfiles-cli-test.js
honzajavorek: It's only for clarity, `-l` stays. I take it for granted the `optimist` deals with the options correctly, so they're not tested in both versions, usually just with one. If the CLI was better isolated, why not, but at this point it just an overhead.

test/unit/Dredd-test.js
honzajavorek: That's a good point.

docs/usage-js.rst
honzajavorek: Yes, I realized this one when reading one article about CLI apps later after issuing the Pull Request already. I'd like to do it exactly as you write, I think it's [easy to detect TTY in the Node](https://nodejs.org/api/process.html#process_a_note_on_process_i_o).

I wasn't sure though whether to snowball it here or to make the change separately. Also, I'm not sure the `stdout` and `stderr` are completely well separated (they weren't even before this change, the logger seems to print each message to whatever it feels is right... I want to fix that as a next step as well), which could make this complicated now.

lib/HooksWorkerClient.js
honzajavorek: Good point, will do in a separate PR. It doesn't seem that anyone ever thought about consistency in that.

lib/CLI.js
honzajavorek:
```js
// This thing-a-ma-bob here is only for purpose of testing
// It's basically a dependency injection for the process.exit function
```

If I'm ever touching that code, I'll do that in order to find and implement less crazy way how to test how the program exits.

A closure could be an improvement, but I'd rather see the code to go away.

lib/options.json
honzajavorek: I'm intentionally doing evil here by skipping the deprecation period for things to allow myself a major series of breaking changes to tidy up Dredd's codebase.

The major release version should be a warning for people to read the changelog and there are always clear steps they need to take to make their setup work.

So yes, I could deprecate this gradually, but I'm afraid I'd end up managing deprecation warnings in the codebase instead of going forward with cleaning up the project from technical debts the fastest way I can. What do you think of this? Is it way too evil?

test/integration/cli/configuration-cli-test.js
honzajavorek: Yes, exactly my wish as well. At this point I'm only separating the loggers, but ideally the reporters would render the output using something completely else than Winston and whatever it is, it would be dependency injected for sure.

lib/CLI.js
honzajavorek: Tried to improve this in 67b6b2e46ae4e272ef6011fbf0daa9032b42d8a3

lib/CLI.js
honzajavorek: Tried to improve this in 67b6b2e46ae4e272ef6011fbf0daa9032b42d8a3

honzajavorek:
@kylef I replied to your comments and added two commits to address some. Most of your comments lead to follow-up work, which I'm to document somewhere soon.

I'm defending my evil approach in https://github.com/apiaryio/dredd/pull/1213#discussion_r253211888 and I'm interested in your opinion on this before I merge the PR.

lib/options.json
kylef: As a user it would be nice, it doesn’t have to be for long though. 3-6 month period. It’s up to you though not blocking on this.

lib/CLI.js
kylef: This is completely out of the scope of these changes but a potentially useless idea I wanted to share.

I was also thinking / wondering how localisation is done for Node CLI apps (guessing that no one is tackling that). I've dealt with internationalisation / localisation in the past in other languages and environments and you usually end up one with language specific file that declares all of the human readable messages which might be nice to separate them from the logic. Makes it easier for people to change and potentially easier to spell check them in CI.

There's a nice example for iOS I wrote at https://github.com/cocodelabs/palaver-l10n/ (even has Czech locale and linting of the language files). In a past life I also wrote django-request which has localisation (https://github.com/django-request/django-request/tree/master/request/locale).

In this case we could have a language or similiar file that could contain all of the output messages, it might be a nice separation.

```js
// language-en.js

const EXIT_DEBUG_MESSAGE = exitStatus => `Exiting Dredd process with status '${exitStatus}'.`;
```


```js
// some-file.js

const { EXIT_DEBUG_MESSAGE } = require('./locale.js');

console.log(EXIT_DEBUG_MESSAGE(1));
```

lib/options.json
honzajavorek: Addressed in 01b042e

lib/configuration.js
```diff
@@ -44,6 +44,44 @@ function applyLoggingOptions(options) {
 }
 
 
+function deprecate(config = {}) {
+  if (config.options) {
+    if (config.options.c) {
+      throw new Error('DEPRECATED: Dredd does not support '
```
kylef: Since this is erroring, this is not deprecation. It's removal. Deprecation would be when it still works but is discouraged and potentially scheduled for removal in the future:

> In several fields, deprecation is the discouragement of use of some terminology, feature, design, or practice, typically because it has been superseded or is no longer considered efficient or safe, without completely removing it or prohibiting its use. It can also imply that a feature, design, or practice will be removed or discontinued entirely in the future

From [Wikipedia on Deprecation](https://en.wikipedia.org/wiki/Deprecation).

kylef: To fix, if you want to keep this behaviour I think you can rename it to now include "deprecated". Alternatively you could actually deprecate it and set `config.options.equivilent option = foo` etc

lib/CLI.js
honzajavorek: Seems nice! I also worked on some localized apps before, mostly using [Babel](https://pythonhosted.org/Flask-Babel/), which leverages gettext. Using constants like this could be a nice poor mans gettext. Would you file it as an idea to issues?

lib/configuration.js
honzajavorek: You're right. But while I could translate some of the options to the new ones, certain features are just gone. What word should I use to alert the user about the fact the feature is gone? If I don't do this, `optimist` silently passes even with extra options (yeah, facepalm).

I want Dredd to fail fast in case I know someone just tried to use a feature which is not in place anymore.

honzajavorek: > I think you can rename it to now include "deprecated"

Took me a moment to parse the sentence, but if you meant

> I think you can rename it to **not** include "deprecated"

then I could do that, yes 😄

honzajavorek: 072ce78

honzajavorek:
The smoke test is failing as the dredd-example uses `--level` 😐 I'll remove it from there for the time of merging this PR and then I'll replace it with `--loglevel`.

honzajavorek:
I added some ugly code to deal with the deprecations 🙈 At least it's tested. I'll think of a better way to do this in the future, but it probably won't get much better unless I decide to dive into #1101

lib/configuration.js
honzajavorek: ff55694

honzajavorek:
Here it is: ff55694

honzajavorek:
@kylef please re-review 🙏

test/unit/configuration-test.js
```diff
@@ -156,9 +162,409 @@ describe('configuration.applyLoggingOptions()', () => {
...
     });
-
     it('the application logger is set to add timestamps', () => {
       assert.isTrue(logger.transports.console.timestamp);
     });
   });
 });
+
+
+describe('configuration._coerceRemovedOptions()', () => {
```
kylef: I don't like the way we are testing the specific unit for a few reasons (firstly we have to export this internal method for testing), but the biggest reason is it is perhaps error prone. Especially considering Dredd does not error out for invalid options. For example, if I invoke dredd with `--foo` or even a typo of a valid option, dredd will continue instead of erroring out:

```
$ dredd api-description.apib http://127.0.0.1:3000 --dry-run --foo
info: Beginning Dredd testing...
...
complete: Tests took 5ms
```

IMO this is a huge anti pattern because if a user typos or misenters an option it could result in undesired behaviour which could potentially be destructive.

I'll show you perhaps an example outside of Dredd scope:

```shell
$ rm -fr --dry-run /
rm: illegal option -- -
usage: rm [-f | -i] [-dPRrvW] file ...
       unlink file
```

Now, if `--dry-run` (perhaps I typoed `--dri-run`) doesn't exist the potentially destructive command did not proceed. An example closer to Dredd could be perhaps attempting to use Dredd CLI options to skip certain destructive methods. For example if I wanted to run Dredd on production and wanted to skip method DELETE for obvious reasons and I typoed the skip command I would hope that Dredd errors out due to invalid arguments instead of running a mis-understood command.

---

Anyway, my actual point (sorry got distracted on side-bug story/tangent 😀) is what if `_coerceRemovedOptions` created a invalid option that doesn't exist we would never know. Since the test is for the unintegrated unit (and also Dredd doesn't validate invalid options) we could have room for invalid code. Perhaps right now we can be 100% sure with careful review, but what about in a few years times when (future) you or someone else decides to remove an option `--color` because it is spelt wrong :trollface: and they add `--colour` these tests would continue to work so they cannot protect against this type of bug.

lib/CLI.js
kylef: I didn't nessecerily mean supporting localisation. I was thinking of it as a way to expose all the human readable output in a single source file instead of the user facing strings skattered across the code base.

test/unit/configuration-test.js
kylef: Anyway, tl;dr I am not sure these specific tests add much value over what the code does. They do not protect you against future refactoring.

lib/CLI.js
honzajavorek: Oh, hmm. I don't think that's very useful for application logging. It seems to me it would rather obscure it for the reader. Perhaps for the output of the reporters, IDK. I don't see that much value in having the strings at one place if we don't do localisation. It's easier to read and edit it in place, within the context of the code which triggers it.

test/unit/configuration-test.js
honzajavorek: I have to admit I've written the tests _ex post_. They were useful in catching bugs in the function I wrote prior to them 😓 

Your points are completely valid. I consider the whole file to be a brownfield. The fact Dredd is in general silent about mistyped options is terrible, but it is an existing behavior. It's solvable probably only by switching the CLI options parser and changing how the configuration is processed 😞

Perhaps `_coerceRemovedOptions` could have been checked also by an integration test, but ideally I remove all the checks after a period of time with a major release and introduce a better configuration management for Dredd (#1101).

honzajavorek: tl;dr I completely agree, but I think fixing this is out of the scope of this PR. I added ugly things, I know it, and I hope I'll have time to improve the CLI later. And I hope if I won't have time to do it, future generations of people working on Dredd at least find these comments and won't hate me that much 🙏 😄

Issue #1217, opened 2019-01-24
Move the longest running intergration tests from 'beforeEach' to 'before'


honzajavorek:
This is a note that the Server Process tests in `server-process-cli-test.js` are the slowest tests and should be moved from using `beforeEach()` to `before()`.

honzajavorek:
The same is valid for "Regression: Issues #319 and #354" tests

PR #1230, opened 2019-02-01
Speedup tests

honzajavorek:

In the pursuit of extreme test isolation, the test suite is extremely slow compared to what's in there. Sure, integration tests with process spawning etc. are destined to be slow, but otherwise there's a lot of room for improvement if we agree `beforeEach()` really isn't needed everywhere.

------------

Before:

image

After:

image

Close https://github.com/apiaryio/dredd/issues/1217

- [ ] To write docs
- [x] To write tests


kylef:
Nice! That has a decent impact.

PR #1231, opened 2019-02-06
Fix JSON syntax in options to be consistent

kylef:

I opened the file and it was hard to read.

Nupe.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/lib/options.json
+++ b/lib/options.json
@@ -113,8 +113,12 @@
     "default": []
   },
   "help": {
-    "description": "Show usage information."  }, "version": {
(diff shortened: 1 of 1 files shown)


honzajavorek:
I usually opened the file very quickly, only jumping to the option I was interested in by search in my editor. That's why I probably didn't notice it. Definitely not the state I'd like the file to be in.

honzajavorek:
Looks like there's an issue with AppVeyor 🤷‍♂️ We have the push build passing, so I'm merging.

screenshot 2019-02-07 at 13 52 25

PR #1235, opened 2019-02-11
Fix the example code of OpenAPI 2 (api-description.yml) in Quickstart…

hama1080:

Dredd cannot parse the api-description.yml on the Quickstart page currently.

```
> dredd .\api-description.yml http://localhost:3000
error: Parser error in file './api-description.yml': Cannot read property 'version' of null on line 1
error: Error when processing API description.
```

The problem is caused by the grammatical error of yaml file. Therefore I update it using Swagger Editor.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -44,31 +44,31 @@ First, let’s design the API we are about to build and test. That means you wil
 
       .. code-block:: openapi2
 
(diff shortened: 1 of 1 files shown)


honzajavorek:
Oh yeah, the troublesome indentation 🙂 Thank you!

PR #1245, opened 2019-02-21
Update dependencies

realityking:

For some reason these updates where not picked up by Greenkeeper.

`make-dir` is the most interesting one of these updates, as it will use the native Node.js API to recursively create directories, if that API is available.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "README.md"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


honzajavorek:
Thanks! Somehow Greenkeeper feels unreliable 😕 I already had to restart it once on the repos to get it working again.

PR #1258, opened 2019-03-01
Fix and test how annotations are propagated

honzajavorek:

It's better to have this tested as I'm going to make further changes to the code soon.

- [ ] To write docs
- [x] To write tests


lib/Dredd.js
```diff
@@ -49,6 +49,7 @@ class Dredd {
     };
     this.transactions = [];
     this.runner = new Runner(this.configuration);
+    this.logger = logger;
```
honzajavorek: Not a _great_ solution, but will do for now. I want to refactor this part of the codebase further in future PRs.

lib/handleRuntimeProblems.js
```diff
@@ -1,7 +1,10 @@
 const blueprintUtils = require('./blueprintUtils');
-const logger = require('./logger');
+const defaultLogger = require('./logger');
+
+
+module.exports = function handleRuntimeProblems(blueprintData, logger) {
```
honzajavorek: I plan to remove this whole file in the future.

lib/handleRuntimeProblems.js
```diff
@@ -27,10 +30,11 @@ module.exports = function handleRuntimeProblems(blueprintData) {
         log(message);
       } else {
         const transactionName = [
+          annotation.origin.apiName,
```
kylef: We're adding new component to the transaction name which is a feature? This commit is tagged as a fix.

honzajavorek: It is only for the purpose of printing it out in the annotation. So if the annotation text is considered to be a public API, then this should be `feat`, but otherwise no. I don't know why this code creates the transaction name from scratch, but since I'm about to delete it, I don't care that much.

PR #1259, opened 2019-03-01
Upgrade Dredd Transactions to v7

honzajavorek:
Depends on https://github.com/apiaryio/dredd/pull/1258

-------

#### :rocket: Why this change?

This upgrades Dredd Transactions and uses their new public interface. The `compileTransactions` function is a mess, but refactoring that is a challenge for another day.

#### :white_check_mark: What didn't I forget?

- [ ] To write docs
- [x] To write tests
- [x] To put [Conventional Changelog](https://dredd.org/en/latest/internals.html#sem-rel) prefixes in front of all my commits and run `npm run lint`

Issue #1257, opened 2019-02-28
Verify how Dredd deals with non-2xx responses in OpenAPI 3


honzajavorek:
Dredd skips non-2xx responses for OpenAPI 2:

- https://dredd.org/en/latest/how-it-works.html#id12
- https://dredd.org/en/latest/how-to-guides.html#id13

A similar behavior should be implemented for OpenAPI 3 as well. Currently the code only deals with OpenAPI 2.

artem-zakharchenko:
Do I get it correctly that it's about extending this conditional skipping in `TransactionRunner`?

https://github.com/apiaryio/dredd/blob/f7e5ee45bccd1097a65c68884cfefff250a859ca/packages/dredd/lib/TransactionRunner.js#L344-L355

honzajavorek:
I'd say the steps are:

1. Learn about OpenAPI 3 and see whether the way it generates transactions is similar enough to OpenAPI 2 so that it makes sense to approach the underlying problem the same way.
2. If yes, then write tests verifying the desired behavior.
3. Then the media type in `transaction.apiDescription.mediaType` should give you relevant media type for the OpenAPI 3 (a different one, not containing `swagger`, that's why it probably doesn't work correctly now). See the [OAS3 adapter code](https://github.com/apiaryio/api-elements.js/blob/c7b61228f370575739c5f806611e00e896296f5f/packages/fury-adapter-oas3-parser/lib/adapter.js#L7).

The underlying problem this tries to solve is that OpenAPI generates transactions in big amounts (especially through `produces` and `consumes`) and a lot of them might be for 4xx status codes. People usually don't want to test those, or in other words, they usually cannot be tested out of the box, without hooks. That's why Dredd skips non-2xx transactions by default and lets users to unskip them in hooks. That's as close to the behavior most of the people would expect out of the box as possible.

hedleysmith:
Dredd seems to be working fairly well for OAS3 for me, except for this issue (so far). I've started a PR which gets things working for OAS3: https://github.com/apiaryio/dredd/pull/1885

IvanGuardado:
I'm having this problem and I'm figuring out how to solve it, since I think it's important to validate all API responses, not just  2XX.

What do you think about linking requests and responses by the example id? For example:

```yaml
responses:
  '200':
    description: OK
    content:
      application/json; charset=utf-8:
        schema:
          $ref: '#/components/schemas/Token'
        examples:
          example-200:
            value:
[...]

PR #1260, opened 2019-03-01
Rename 'configuraton.data' to 'configuration.apiDescriptions'

honzajavorek:
Depends on https://github.com/apiaryio/dredd/pull/1259

---------

This is a beginning of a larger refactoring, which will allow for simpler code at various places.

- Introducing `apiDescriptionMediaType` as part of the transaction object allows to remove a fragile dependency on the `configuration` object in `configureTransaction` inside the transaction runner
- The `apiDescriptions` name makes more sense to me than `data`, the docs and the configuration has been adjusted accordingly
- The configuration _is_ a mess, the tests _are_ ugly, but those are challenges for another day
- The `.includes('swagger')` doesn't scale, but will be addressed separately - https://github.com/apiaryio/dredd/issues/1257

- [x] To write docs
- [x] To write tests

### DIFF
--- a/docs/usage-js.rst
+++ b/docs/usage-js.rst
@@ -33,71 +33,55 @@ Let’s have a look at an example configuration first. (Please also see the :ref
    {
      server: 'http://127.0.0.1:3000/api', // your URL to API endpoint the tests will run against
      options: {
-
-       'path': [],       // Required Array if Strings; filepaths to API description documents, can use glob wildcards
-
+       path: [],         // Required Array if Strings; filepaths to API description documents, can use glob wildcards
        'dry-run': false, // Boolean, do not run any real HTTP transaction
-       'names': false,   // Boolean, Print Transaction names and finish, similar to dry-run
-
-       'loglevel': 'warning', // String, logging level (debug, warning, error, silent)
-
-       'only': [],      // Array of Strings, run only transaction that match these names
-
-       'header': [],    // Array of Strings, these strings are then added as headers (key:value) to every transaction
-       'user': null,    // String, Basic Auth credentials in the form username:password
-
-       'hookfiles': [], // Array of Strings, filepaths to files containing hooks (can use glob wildcards)
-
-       'reporter': ['dot', 'html'], // Array of possible reporters, see folder lib/reporters
-
-       'output': [],     // Array of Strings, filepaths to files used for output of file-based reporters
-
+       names: false,     // Boolean, Print Transaction names and finish, similar to dry-run
+       loglevel: 'warning', // String, logging level (debug, warning, error, silent)
+       only: [],         // Array of Strings, run only transaction that match these names
+       header: [],       // Array of Strings, these strings are then added as headers (key:value) to every transaction
+       user: null,       // String, Basic Auth credentials in the form username:password
+       hookfiles: [],    // Array of Strings, filepaths to files containing hooks (can use glob wildcards)
+       reporter: ['dot', 'html'], // Array of possible reporters, see folder lib/reporters
+       output: [],       // Array of Strings, filepaths to files used for output of file-based reporters
        'inline-errors': false, // Boolean, If failures/errors are display immediately in Dredd run
-
-       'require': null, // String, When using nodejs hooks, require the given module before executing hooks
-
-       'color': true,
+       require: null,    // String, When using nodejs hooks, require the given module before executing hooks
+       color: true,
      },
-
-     'emitter': EventEmitterInstance, // optional - listen to test progress, your own instance of EventEmitter
-
-     'data': {
-       'path/to/file': '...'
-     }
+     emitter: new EventEmitter(), // listen to test progress, your own instance of EventEmitter
+     apiDescriptions: ['FORMAT: 1A\n# Sample API\n']
    }
 
-Properties
-----------
+.. js:data:: configuration
 
-server (string)
-~~~~~~~~~~~~~~~
+.. js:attribute:: configuration.server
 
-Your choice of the API endpoint to test the API description against. It must be a valid URL (you can specify ``port``, ``path`` and http or https ``protocol``).
+   The HTTP(S) address of the API server to test against the API description(s). A valid URL is expected, e.g. ``http://127.0.0.1:8000``
 
-options (object)
-~~~~~~~~~~~~~~~~
+   :type: string
+   :required: yes
 
-Because ``options.path`` array is required, you must specify options. You’ll end with errors otherwise.
+.. js:attribute:: configuration.options
 
-.. _optionspath-object:
+   Because :js:attr:`configuration.options.path` array is required, you must specify options. You’ll end with errors otherwise.
 
-options.path (object)
-^^^^^^^^^^^^^^^^^^^^^
+   :type: object
+   :required: yes
 
-**Required** Array of filepaths to API description documents. Or it can also be an URL to download the API description from internet via http(s) protocol.
+.. js:attribute:: configuration.options.path
 
-data (object)
-^^^^^^^^^^^^^
+   Array of paths or URLs to API description documents.
 
-**Optional** Object with keys as ``filename`` and value as ``blueprint``-code.
+   :type: array
+   :required: yes
 
-Useful when you don’t want to operate on top of filesystem and want to pass code of your API description as a string. You get the point.
+.. js:attribute:: configuration.emitter
 
-.. code-block:: javascript
+   Listen to test progress by providing your own instance of `EventEmitter <https://nodejs.org/api/events.html#events_class_eventemitter>`__.
 
-   {
-     'data': {
-       './api-description.apib': 'FORMAT: 1A\n\n# My String API\n\nGET /url\n+ Response 200\n\n        Some content',
-       './directory/another-api-description.apib': '# Another API\n\n## Group Machines\n\n### Machine [/machine]\n\n#### Read machine [GET]\n\n...'
-     }
-   }
+   :type: EventEmitter
+
+.. js:attribute:: configuration.apiDescriptions
+
+   API descriptions as strings. Useful when you don't want to operate on top of the filesystem.
+
+   :type: array
--- a/lib/Dredd.js
+++ b/lib/Dredd.js
@@ -20,14 +20,10 @@ const PROXY_ENV_VARIABLES = ['HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY'];
 const FILE_DOWNLOAD_TIMEOUT = 5000;
 
 
-function removeDuplicates(arr) {
-  return arr.reduce((alreadyProcessed, currentItem) => {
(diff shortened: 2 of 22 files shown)


docs/usage-js.rst
```diff
@@ -33,71 +33,55 @@ Let’s have a look at an example configuration first. (Please also see the :ref
...
-       'color': true,
+       require: null,    // String, When using nodejs hooks, require the given module before executing hooks
+       color: true,
      },
+     emitter: new EventEmitter(), // listen to test progress, your own instance of EventEmitter
+     apiDescriptions: ['FORMAT: 1A\n# Sample API\n']
+   }
 
-     'emitter': EventEmitterInstance, // optional - listen to test progress, your own instance of EventEmitter
+.. js:data:: configuration
```
honzajavorek: Using the [JS Sphinx domain](http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-javascript-domain) here

docs/usage-js.rst
```diff
@@ -33,71 +33,55 @@ Let’s have a look at an example configuration first. (Please also see the :ref
...
        'inline-errors': false, // Boolean, If failures/errors are display immediately in Dredd run
-
-       'require': null, // String, When using nodejs hooks, require the given module before executing hooks
-
-       'color': true,
+       require: null,    // String, When using nodejs hooks, require the given module before executing hooks
+       color: true,
      },
+     emitter: new EventEmitter(), // listen to test progress, your own instance of EventEmitter
+     apiDescriptions: ['FORMAT: 1A\n# Sample API\n']
```
honzajavorek: I changed the interface for this to a plain array of strings, as I don't see a reason for the user to provide an artificial filename/URL/... or anything like that. In `configuration.js` this is then coerced to something like

```js
[{ content: 'FORMAT: 1A\n# Sample API\n', location: 'configuration.apiDescriptions[0]' }]
```

lib/Dredd.js
```diff
@@ -5,28 +5,26 @@ const fs = require('fs');
...
-function removeDuplicates(arr) {
-  return arr.reduce((alreadyProcessed, currentItem) => {
-    if (alreadyProcessed.indexOf(currentItem) === -1) {
-      return alreadyProcessed.concat(currentItem);
-    }
-    return alreadyProcessed;
-  }, []);
+
+function unique(array) {
+  return Array.from(new Set(array));
```
honzajavorek: This function is intentionally here and in the `configuration.js` as these files are to be further refactored and I don't want to introduce more dependencies between them or any "utility functions" file just because of this.

lib/configuration.js
```diff
@@ -21,6 +21,11 @@ function coerceToBoolean(value) {
 }
 
 
+function unique(array) {
+  return Array.from(new Set(array));
+}
```
honzajavorek: This function is intentionally here and in the `Dredd.js` as these files are to be further refactored and I don't want to introduce more dependencies between them or any "utility functions" file just because of this.

lib/Dredd.js
```diff
@@ -47,8 +45,10 @@ class Dredd {
       end: 0,
       duration: 0,
     };
+    this.configuration.files = [];
```
honzajavorek: Will be renamed to `this.files` as the next step, as it is something completely local to the `Dredd` class 🙄

lib/TransactionRunner.js
```diff
@@ -38,7 +38,7 @@ class TransactionRunner {
 
   config(config) {
     this.configuration = config;
-    this.multiBlueprint = Object.keys(this.configuration.data).length > 1;
+    this.multiBlueprint = this.configuration.apiDescriptions.length > 1;
```
honzajavorek: I didn't bother refactoring this part now. The `multiBlueprint` is something used only for the following code further down the file, and I didn't have the time to figure out all the consequences of it yet:

```js
    // Backward compatible transaction name hack. Transaction names will be
    // replaced by Canonical Transaction Paths: https://github.com/apiaryio/dredd/issues/227
    if (!this.multiBlueprint) {
      transaction.name = transaction.name.replace(`${transaction.origin.apiName} > `, '');
    }
```

lib/TransactionRunner.js
```diff
@@ -247,7 +241,7 @@ class TransactionRunner {
     // Transaction skipping (can be modified in hooks). If the input format
     // is OpenAPI 2, non-2xx transactions should be skipped by default.
     let skip = false;
-    if (mediaType.indexOf('swagger') !== -1) {
+    if (transaction.apiDescriptionMediaType.includes('swagger')) {
```
honzajavorek: Not robust, but will get fixed later: https://github.com/apiaryio/dredd/issues/1257

lib/configuration.js
```diff
@@ -166,6 +193,13 @@ function applyConfiguration(inConfig) {
     },
   };
 
+  // Transform apiDescriptions from strings to an array of objects
+  outConfig.apiDescriptions = coerceToArray(inConfig.apiDescriptions)
+    .map((apiDescription, i) => ({
+      location: `configuration.apiDescriptions[${i}]`,
+      content: apiDescription,
+    }));
```
honzajavorek: I'd like to have more separation between what user provides to Dredd and what's Dredd's resolved, internal configuration, but that requires to pretty much revamp how Dredd gets configured. So doing this for now, coercing user's input to a data structure Dredd uses internally.

test/unit/Dredd-test.js
```diff
@@ -5,13 +5,18 @@ const proxyquire = require('proxyquire').noCallThru();
 const requestStub = require('request');
 const sinon = require('sinon');
 const { assert } = require('chai');
+const parse = require('dredd-transactions/parse');
+const compile = require('dredd-transactions/compile');
 
-const dreddTransactionsStub = require('dredd-transactions');
 const loggerStub = require('../../lib/logger');
```
honzajavorek: I didn't really try to make the tests better, just to make them pass.

test/unit/Dredd-test.js
```diff
@@ -116,20 +124,31 @@ describe('Dredd class', () => {
 
       it('should load file contents on paths to config', done => dredd.run((error) => {
         if (error) { return done(error); }
-        assert.isObject(dredd.configuration.data);
-        assert.property(dredd.configuration.data, './test/fixtures/multifile/greeting.apib');
-        assert.isObject(dredd.configuration.data['./test/fixtures/multifile/greeting.apib']);
-        assert.property(dredd.configuration.data['./test/fixtures/multifile/greeting.apib'], 'filename');
-        assert.property(dredd.configuration.data['./test/fixtures/multifile/greeting.apib'], 'raw');
+        assert.lengthOf(dredd.configuration.apiDescriptions, 3);
+        dredd.configuration.apiDescriptions.sort((ad1, ad2) => ad1.location.localeCompare(ad2.location));
```
honzajavorek: There is sorting on multiple places, because Dredd currently resolves the apiDescriptions in an asynchronous way (downloading remote API description files, resolving glob patterns, ...), i.e. in non-deterministic order. To make the resolution deterministic would mean to rewrite half of the `Dredd.js` file, so I opted for sorting the array in tests for now.

test/unit/transactionRunner-test.js
```diff
@@ -298,11 +290,7 @@ describe('TransactionRunner', () => {
       beforeEach(() => {
         transaction.response.status = 400;
         transaction.origin.filename = filename;
-
-        if (!runner.configuration.data) { runner.configuration.data = {}; }
-        if (!runner.configuration.data[filename]) { runner.configuration.data[filename] = {}; }
-        runner.configuration.data[filename].mediaType = 'text/plain';
-
+        transaction.apiDescriptionMediaType = 'text/plain';
```
kylef: Shouldn't this be a Swagger media type?

> when processing other than OpenAPI 2 document and given transaction has non-2xx status code

test/unit/transactionRunner-test.js
```diff
@@ -279,11 +275,7 @@ describe('TransactionRunner', () => {
         beforeEach(() => {
           transaction.response.status = status;
           transaction.origin.filename = filename;
-
-          if (!runner.configuration.data) { runner.configuration.data = {}; }
-          if (!runner.configuration.data[filename]) { runner.configuration.data[filename] = {}; }
-          runner.configuration.data[filename].mediaType = 'application/swagger+json';
-
+          transaction.apiDescriptionMediaType = 'application/swagger+json';
```
kylef:
```
const filename = 'api-description.yml';
```

Shouldn't the media type be the YAML variant?

test/unit/Dredd-test.js
kylef: Is `localeCompare` correct to use here? Perhaps the tests could fail on some users system where they use a locale that is pretty different to our assumptions in the file order as we're expecting the result to be consistent with assertions below.

lib/TransactionRunner.js
kylef: Outside scope of this PR but brings it to my attention, should the same rule apply to OAS3?

honzajavorek: That's a subject of discussion / pending design. I need to read the spec first to properly understand what's possible to describe with it and what default behavior makes sense for the users.

Eventually, I'd like to unify the behavior across the formats, but that's a long way to go as that's something which depends on the transaction names 'epic' 😬

test/unit/Dredd-test.js
honzajavorek: I guess I could use a custom sorting function in that case. It's such a pain to sort in JS 😞 I miss Python's `key` approach.

test/unit/transactionRunner-test.js
honzajavorek: "other than OpenAPI 2 document", so it should be anything else than Swagger

test/unit/transactionRunner-test.js
honzajavorek: For consistency, yes. Or the filename to be `.json`.

test/unit/transactionRunner-test.js
kylef: Oh my bad, misread.

test/unit/transactionRunner-test.js
honzajavorek: Oh, looking into the code now and it's because Dredd takes whatever Fury returns as the first media type from the `.detect()` function. That way it always gets `application/swagger+json`, even for YAML files.

It's not an issue for Dredd, but it _is_ an issue once it's propagated to the transaction object. This is actually a blocking problem 😞

honzajavorek: Solved in a dirty way in 32f3864

honzajavorek:
Now the PR contains a commit using `feat:` and then a `refactor:` commit reverting the feature. I'm wondering what's the best way to solve this, perhaps <kbd>Squash & merge</kbd>, where I squash to something like this?

```
refactor: rename 'configuraton.data' to 'configuration.apiDescriptions'
```

PR #1270, opened 2019-03-09
fix: (dredd.yml) Uses default "language" value when not prompted

### LINKED ISSUE #1239: dredd init generates invalid file
**Describe the bug**
I ran `dredd init` which generates a `dredd.yml` config file.
When I run `dredd --dry-run` I get the following error:

```
error: unknown tag !<tag:yaml.org,2002:js/undefined> at line 4, column 47:
     ... g:yaml.org,2002:js/undefined> ''
```

Looking in the generated config there is this line:
```yaml
language: !<tag:yaml.org,2002:js/undefined> ''
```

This seems to be what is broken.

Deleting that line fixes the problem.

```
dredd v8.0.0 (Darwin 18.2.0; x64)
```

kettanaito:

Aims to set a proper value for the `language` key in the generated `dredd.yml` after running the following command:

```bash
dredd init
```

Please refer to the mentioned GitHub issue for the reproduction steps.

- Fixes #1239 

- [x] To write docs (not applicable? I assume current docs cover the use case)
- [x] To write tests

### DIFF
--- a/lib/init.js
+++ b/lib/init.js
@@ -180,7 +180,7 @@ function applyAnswers(config, answers, options = {}) {
   config._[1] = answers.apiHost;
 
   config.server = answers.server || null;
-  config.language = answers.language;
+  config.language = answers.language || 'nodejs';
 
   if (answers.apiary) { config.reporter = 'apiary'; }
   if (answers.apiaryApiKey) { config.custom.apiaryApiKey = answers.apiaryApiKey; }
--- a/test/unit/init/applyAnswers-test.js
+++ b/test/unit/init/applyAnswers-test.js
@@ -38,9 +38,9 @@ describe('init._applyAnswers()', () => {
     const config = applyAnswers(createConfig(), { language: 'python' });
     assert.equal(config.language, 'python');
   });
-  it('sets the language when not prompted', () => {
-    const config = applyAnswers(createConfig({ language: 'python' }), {});
-    assert.equal(config.language, 'python');
+  it('uses default language (nodejs) when none is prompted', () => {
+    const config = applyAnswers(createConfig(), { language: undefined });
+    assert.equal(config.language, 'nodejs');
   });
   it('sets no reporter by default', () => {
     const config = applyAnswers(createConfig(), {});
--- a/test/unit/init/applyAnswers-test.js
+++ b/test/unit/init/applyAnswers-test.js
@@ -38,6 +38,10 @@ describe('init._applyAnswers()', () => {
     const config = applyAnswers(createConfig(), { language: 'python' });
     assert.equal(config.language, 'python');
   });
+  it('sets the language when not prompted', () => {
+    const config = applyAnswers(createConfig({ language: 'python' }), {});
+    assert.equal(config.language, 'python');
+  });
   it('sets no reporter by default', () => {
     const config = applyAnswers(createConfig(), {});
     assert.isUndefined(config.reporter);


lib/init.js
```diff
@@ -164,6 +164,8 @@ function prompt(config, detected, callback) {
     },
   ]).then((answers) => {
     callback(null, answers);
+  }).catch((error) => {
```
kettanaito: I've noticed Promise rejection is not handled, thus the following line will never resolve in case of error:

https://github.com/apiaryio/dredd/blob/07c8c8157de47658d5e8f8f4c58ee10ea1f8ef21/lib/init.js#L25

kettanaito:
Will provide tests. As of now, I'm not sure if grabbing the default value from `config` is a proper solution. Based on tests I can see that config can be empty, and thus the proper default value of the language is, most likely, the `detected.language` value. However, that would complicate the unit testing, as it would require to detect the language as a part of a unit test.

test/unit/init/applyAnswers-test.js
```diff
@@ -38,6 +38,10 @@ describe('init._applyAnswers()', () => {
     const config = applyAnswers(createConfig(), { language: 'python' });
     assert.equal(config.language, 'python');
   });
+  it('sets the language when not prompted', () => {
+    const config = applyAnswers(createConfig({ language: 'phyton' }), {});
```
honzajavorek: Typo: phyton

test/unit/init/applyAnswers-test.js
```diff
@@ -38,6 +38,10 @@ describe('init._applyAnswers()', () => {
...
+    assert.equal(config.language, 'phyton');
```
honzajavorek: Typo: phyton

honzajavorek:
Looking at the code again after reading your comment.

The way it is wired together now is that the `init()` takes output from the [optimist](https://github.com/substack/node-optimist) as the initial `config`. This way it can set CLI arguments with priority.

I took a moment to read the `init.js` code and I found out the `language` should be a single "init question" where the fix is needed. All other questions actually seem to correctly count with the Inquirer behavior, i.e. if `when` doesn't pass, there will be no value and `default` gets ignored.

I think the fix is probably easier than it seems. The problem is caused by the fact that for various historical reasons, `config.language`'s effective "empty" value is `"nodejs"`.

So the flow goes as follows. When user doesn't specify the hook language in the CLI, it defaults to `nodejs` (see `options.json` file). In that case, the line 71 passes and the question about hooks gets asked:

```js
    {
      name: 'hooks',
      message: 'Do you want to use hooks to customize Dredd\'s behavior?',
      type: 'confirm',
      default: true,
      when: () => config.language === 'nodejs',
    },
```

If user wishes to set the hooks language, they are presented with the following question:

```js
    {
      name: 'language',
      message: 'Programming language of the hooks',
      type: 'list',
      default: detected.language,
      choices: [
        { name: 'Go', value: 'go' },
        { name: 'JavaScript', value: 'nodejs' },
        { name: 'Perl', value: 'perl' },
        { name: 'PHP', value: 'php' },
        { name: 'Python', value: 'python' },
        { name: 'Ruby', value: 'ruby' },
        { name: 'Rust', value: 'rust' },
      ],
      when: answers => answers.hooks,
    },
```

There is no "empty" answer. The "empty" answer is `nodejs`. If the question isn't asked, the answer will be `undefined` though. I think we could take it as a possible, valid outcome of the prompting, and just make sure the `applyAnswers()` correctly deals with it:

```diff
- config.language = answers.language;
+ config.language = answers.language || 'nodejs';
```

Then testing this in `applyAnswers-test.js` as follows:

```js
it('sets the language if not provided', () => {
  const config = applyAnswers(createConfig(), { language: undefined });
  assert.equal(config.language, 'nodejs');
});
```

------

The code, and especially the line `when: () => config.language === 'nodejs'` was a bit confusing to me until I realized the `nodejs` is the effective empty value for that particular option. Unfortunately it's something not easy to spot without a larger context of the codebase, making contributions harder 😞

test/unit/init/applyAnswers-test.js
kettanaito: Thanks. Fixed by d64c8d7.

test/unit/init/applyAnswers-test.js
kettanaito: Thanks. Fixed by d64c8d7.

kettanaito:
Thanks for the detailed insights, @honzajavorek.

Okay, in this case we can fallback to a hard-coded `"nodejs"`, as I don't see something as `defaultLanguage` variable any where. That should be fine.

I've fixed the typos, and adjusted the `init.js` and corresponding test. Should be good now :)

honzajavorek:
Just a nitpick - could you please change the commit messages so they [adhere to the Semantic Release convention](https://dredd.org/en/latest/internals.html#commit-message-format)?

```diff
- test: Fixes a typo in "applyAnswers" unit test
+ test: fix a typo in "applyAnswers" unit test
- fix: (dredd.yml) fallbacks to "nodejs" when no language prompted 
+ fix(dredd.yml): fallback to "nodejs" when no language prompted 
```

You can check it locally by `npm run lint` as well. It helps us to employ Semantic Release for automatic releases, which lowers maintenance overhead.

kettanaito:
@honzajavorek, yes, done. I've spotted that's the reason why CI fails, so rewrote the commit history. Hope it's aligned with the rules now.

PR #1274, opened 2019-03-12
Use the same code for resolving globs in paths

honzajavorek:

Dredd now resolves globs for both hookfiles and API descriptions paths. The logic of resolving them is the same, so I think the code taking care of it should be the same as well.

- [ ] To write docs
- [x] To write tests

### DIFF
--- a/lib/resolveHookfiles.js
+++ b/lib/resolveHookfiles.js
@@ -2,31 +2,40 @@ const fs = require('fs');
 const path = require('path');
 const glob = require('glob');
 
-// Ensure platform agnostic path.basename function
+
+// Ensure platform-agnostic 'path.basename' function
 const basename = process.platform === 'win32' ? path.win32.basename : path.basename;
 
-// Expand hookfiles - sort files alphabetically and resolve their paths
-module.exports = function resolveHookfiles(hookfiles, cwd = null) {
+
+function resolveGlob(pattern, workingDirectory = null) {
+  const cwd = workingDirectory || process.cwd();
+
+  // 'glob.sync()' does not resolve paths, only glob patterns
+  if (glob.hasMagic(pattern)) {
+    return glob.sync(pattern, { cwd })
+      .map(matchingPath => path.resolve(cwd, matchingPath));
+  }
+  const resolvedPath = path.resolve(cwd, pattern);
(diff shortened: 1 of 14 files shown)


honzajavorek:
@kylef I was able to fix all Windows tests, the PR is ready for review. Sorry for asking you for a review before everything was passing.

test/unit/Dredd-test.js
```diff
@@ -120,9 +121,9 @@ describe('Dredd class', () => {
         if (error) { return done(error); }
         assert.lengthOf(dredd.files, 3);
         assert.deepEqual(dredd.files, [
-          `${CWD}/test/fixtures/multifile/greeting.apib`,
-          `${CWD}/test/fixtures/multifile/message.apib`,
-          `${CWD}/test/fixtures/multifile/name.apib`,
+          path.join(CWD, '/test/fixtures/multifile/greeting.apib'),
```
kylef: Since you're using path.join, perhaps we shouldn't hard-code the rest of the path separators in this too?

```js
path.join(CWD, 'test', 'fixtures', 'multifile', 'greeting.apib'),
```

Same applies to all joins below.

lib/resolveHookfiles.js
```diff
@@ -7,9 +7,7 @@ const glob = require('glob');
 const basename = process.platform === 'win32' ? path.win32.basename : path.basename;
 
 
-function resolveGlob(pattern, workingDirectory = null) {
-  const cwd = workingDirectory || process.cwd();
-
+function resolveGlob(cwd, pattern) {
```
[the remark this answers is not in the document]
kylef: I later realised that this doesn't always represent the current working directory and the name is maybe tricky.

Perhaps it would be clearer to rename this from `cwd` to something else, like `directory`, `base` or `working_directory`. "cwd" stands for current working directory which I don't believe this is.

test/unit/Dredd-test.js
honzajavorek: If it's for the sake of portability, `path.join()` can correctly handle this on Windows.

(The reason why Windows tests were failing was actually the fact I wasn't using `path.join()` everywhere in the tests, which translates the separators.)

If it's for the sake of readability, the current version feels better (to me).

Issue #913, opened 2017-11-11
Expose "request" options to hook scripts


marc0l92:
The `request` library used to query the server allows to specify several client options. Some of them are related to the client certificate, the self-signed certificate validation and others.

It is possible to make this option configurable from the hook script or from the command line?

#### Possible dirty solution
In order to use your applicaiton with a http server with self signed certificate I added this line in the file `lib/transaction-runner.js`:

```
88:      return function() {
89:        if (_this.hookHandlerError) {
90:          return callback(_this.hookHandlerError);
91:        }
->         if(hooks.configuration) _this.configuration.http = hooks.configuration.http;
92:        return async.timesSeries(transactions.length, function(transactionIndex, iterationCallback) {
93:          transaction = transactions[transactionIndex];
94:          logger.verbose("Processing transaction #" + (transactionIndex + 1) + ":", transaction.name);
95:          logger.verbose('Running \'beforeEach\' hooks');
```

With this change I can now configure the `request` library options from my hook file:
```
...
hooks.beforeAll(function (transactions, done) {
  // Allow server self-signed certificates
  hooks.configuration.http.rejectUnauthorized = false;
  // Set client certificate
  hooks.configuration.http.cert = fs.readFileSync(certFile);
  hooks.configuration.http.key = fs.readFileSync(keyFile);
  done();
});
...
```

honzajavorek:
I would be quite cautious exposing the whole options, but from the description I understand you need it primarily for testing an API with complex SSL setup - certs, keys, etc. This would be related to https://github.com/apiaryio/dredd/issues/323. Let me rename your issue to "Support SSL certificates".

abtris:
You don’t need change code, just set env variable as dirty hack. 

```
process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = 0;
```

PR #1278, opened 2019-03-18
feat: ssl integration

mihikasharma:
Create a mechanism to test the endpoints that are secured with ssl integration. Helps to cover all the endpoint testing.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/lib/performRequest.js
+++ b/lib/performRequest.js
@@ -35,6 +35,10 @@ function performRequest(uri, transactionReq, options, callback) {
   try {
     httpOptions.body = getBodyAsBuffer(transactionReq.body, transactionReq.bodyEncoding);
     httpOptions.headers = normalizeContentLengthHeader(transactionReq.headers, httpOptions.body);
+    httpOptions.cert = transactionReq.cert;
+    httpOptions.key = transactionReq.key;
+    httpOptions.requestCert = transactionReq.requestCert;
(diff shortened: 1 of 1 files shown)


honzajavorek:
Hi, thanks! I think this would tackle #913 in a sense (btw, did you try the workaround mentioned in https://github.com/apiaryio/dredd/issues/913#issuecomment-381419699 ?), but if we're going to implement a solution to SSL certificates, I'd like it to be on a global level (configuration options and CLI options) rather than per transaction.

mihikasharma:
Hi. I was trying to accomodate it to per request basis as we could have different certificates for different requests.

honzajavorek:
Is different certificates for different Dredd requests the reality you're coping with in your project or is it something you think might be useful?

weberr13:
Will this allow client side SSL auth then?  If so it is definitely something that I'm interested in using/seeing. As far as I can tell the only Automated REST testing framework that currently supports client side cert base auth is SoapUI.

reppners:
@honzajavorek I'm looking for a api testing framework that can ingest OpenApi v3 but a majority of endpoints needs authentication via client certificate and it needs to be configurable per request. So I'd really would love to see this going forward - but if certificate settings are global then it would not be usable at all for the APIs I'm working with.

Issue #1288, opened 2019-03-25
Reorganization and cleanup of Dredd.js Integration Tests


honzajavorek:
The current contents of the [`test/integration/dredd-test.js`](https://github.com/apiaryio/dredd/blob/master/test/integration/dredd-test.js) is a mess of various things. This is a proposal on how to restructure it so the tests make more sense. Each heading stands for a separate file.

#### Dredd.js Integration tests
- [x] successful run returns no error and expected stats
- [x] erroneous run returns an error and stats
- [x] requires --require module, deals with an error
- [x] loads ADD from a local file
- [x] loads ADDs from local files specified by globs (expands globs in config)
- [x] loads ADD from URL
- [x] loads ADDs from various sources in their given order
- [x] deals with error on loading local file
- [x] deals with error resolving globs to non-existing files
- [x] deals with error on loading remote file from non-existing server
- [x] deals with error on loading remote file from existing server but 404 path
- [x] sets configuration.apiDescriptions
- [x] warns on warning annotations
- [x] errors and aborts dredd on error annotations
- [x] starts the transaction runner with an expected list of transactions

#### Apiary Reporter integration tests
- [ ] when testing with apiary reporter and custom apiary keys, dredd doesn't log warning about missing settings, should contain auth headers, should send the test run as non-public, should print URL to report, should send results from gavel
- [ ] when testing with apiary reporter and server is not running, shuld print URL to report, should send results from gavel, should let user know the server is down
- [ ] when testing with apiary reporter and server is ok, prints warning about credentials, links to report, links to docs about apiary reporter, sends gavel results

#### OpenAPI 2 specifics
- [ ] when OpenAPI 2 has multiple responses, recognizes all transactions and marks non-200 status codes as skipped
- [ ] when OpenAPI 2 has multiple responses and hooks unskip some of them
- [ ] when OpenAPI 2 and hooks, transaction names are different (contain more info)

Issue #1275, opened 2019-03-15
Proposal: Use promises/async/await in Dredd's codebase


honzajavorek:
Currently Dredd's codebase is written using Node.js-style callbacks, almost exclusively. That's mostly due to historical continuity and the fact most of the parts of the codebase are tightly coupled together, so it's not easy to bring new stuff in gradually. Also, to support promises and async/await nicely means bringing in new dependencies - i.e. even [`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) was only added in Node 8, while Dredd still supports Node 6.

With ongoing refactoring of the codebase, which aims for decoupling things, and with Node 6 going out of maintenance soon, it might make sense to bring promises in, as it's becoming very easy - all of the important stuff is supported in Node 8.

<img width="811" alt="Screenshot 2019-03-15 at 10 19 16" src="https://user-images.githubusercontent.com/283441/54425539-da11dd80-4715-11e9-89bf-b5f2bc9111d2.png">

This is a proposal to:

1. use callbacks and [async.js](https://github.com/caolan/async) consistently everywhere until following steps are taken 
2. deprecate Node 6 as soon as possible (as soon as we're using Node 8 everywhere also internally in @apiaryio)
3. once Node 6 is deprecated, start using async/await (preferably) and promises with newly added code, callbacks becoming a forbidden way of writing asynchronous code
4. once addressing https://github.com/apiaryio/dredd/issues/1200, tackle also all of these: https://github.com/apiaryio/dredd/issues/964, https://github.com/apiaryio/dredd/issues/692, https://github.com/apiaryio/dredd/issues/543

honzajavorek:
I think we can close this now. We're using async/await in the codebase and the Node versions are deprecated. JS hooks are a different story with different, dedicated issues.

Issue #1191, opened 2019-01-10
handleRuntimeProblems & blueprint utils should be merged, renamed, better tested


honzajavorek:
Well the title says it all. Also, we should be more robust about line numbers in annotations. API Elements should provide us soon with an easy, uniform interface to line numbers, so we could just delete a lot of code.

PR #1289, opened 2019-03-25
Refactor Dredd.js

honzajavorek:

I didn't like most of the `Dredd.js` file. This should streamline the pipeline from API description locations to parse results and transactions. The refactoring prepares ground for https://github.com/apiaryio/dredd/issues/1191. With https://github.com/apiaryio/dredd/issues/1275 it could be further refactored and streamlined.

- https://github.com/apiaryio/dredd/issues/1191
- https://github.com/apiaryio/dredd/issues/1275
- https://github.com/apiaryio/dredd/issues/1288

- [ ] To write docs
- [x] To write tests


lib/Dredd.js
```diff
@@ -1,35 +1,94 @@
 const async = require('async');
```
honzajavorek: The code of this file still seems ugly to me at many places and I blame mainly the asynchronous parts. I hope with async/await/promises it could get even more readable and unit-testable in the future.

lib/Dredd.js
```diff
@@ -1,35 +1,94 @@
...
 const logger = require('./logger');
-const Runner = require('./TransactionRunner');
+const TransactionRunner = require('./TransactionRunner');
 const { applyConfiguration } = require('./configuration');
-const options = require('./options.json');
 
 
-const PROXY_ENV_VARIABLES = ['HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY'];
-const FILE_DOWNLOAD_TIMEOUT = 5000;
+function prefixErrors(decoratedCallback, prefix) {
```
honzajavorek: I couldn't come up with a nicer way how to prefix errors with context than decorating the callback like this. I might be just missing a good idea or a best practise. But it's gonna disappear with async/await/promises anyway.

lib/Dredd.js
```diff
@@ -1,35 +1,94 @@
...
+  return (error, ...args) => {
+    if (error) { error.message = `${prefix}: ${error.message}`; }
+    decoratedCallback(error, ...args);
+  };
+}
+
+
+function readLocations(locations, options, callback) {
```
honzajavorek: I feel like readLocations() and parseContent() might get a lot slimmer or even redundant with Promise.all() or Promise.map(), but I decided to stay with callbacks until Dredd drops Node 6 support.

lib/Dredd.js
```diff
@@ -40,215 +99,79 @@ class Dredd {
...
-      callback(err, this.stats);
-      return;
-    }
-
-    // Spin that merry-go-round
-    this.logger.debug('Reading API description files.');
-    this.loadFiles((loadErr) => {
-      if (loadErr) { return callback(loadErr, this.stats); }
+    this.logger.debug('Configuring reporters');
+    configureReporters(this.configuration, this.stats, this.transactionRunner);
```
honzajavorek: The configureReporters() is a rabbit hole of side effects. Leaving it as it is for now.

lib/Dredd.js
```diff
@@ -40,215 +99,79 @@ class Dredd {
...
+      } catch (compileErr) {
+        callback(compileErr);
+        return;
+      }
+
+      callback(null, apiDescriptionsWithTransactions);
+    });
   }
 
   run(callback) {
```
honzajavorek: The Dredd class represents an object with just a single (asynchronous...) method, run(). That's an antipattern and a clear indication it doesn't need to be a class at all. It is interconnected by references with reporters though and it's Dredd's public JavaScript API, so I'm leaving it as a class for now. Also for now ignoring the fact that it calls the callback with both error and the stats (fixing it would be a breaking change). In the future I can imagine something like `const stats = await dredd(config);`

lib/Dredd.js
```diff
@@ -40,215 +99,79 @@ class Dredd {
...
   }
 
-  logProxySettings() {
-    const proxySettings = [];
-    for (const envVariableName of Object.keys(process.env || {})) {
-      const envVariableValue = process.env[envVariableName];
-      if (!Array.from(PROXY_ENV_VARIABLES).includes(envVariableName.toUpperCase())) { continue; }
-      if (envVariableValue === '') { continue; }
-      proxySettings.push(`${envVariableName}=${envVariableValue}`);
+  prepareAPIdescriptions(callback) {
```
honzajavorek: The prepareAPIdescriptions() method isn't a real method of a real object as the Dredd class doesn't need to be a class in the first place. I created it just to make run() shorter. It roughly contains the pipeline turning API description locations into API descriptions, then parsing them, and then compiling them into transactions. Once we have async/await/promises and the code is less tied with 'this.configuration', it might make sense to refactor this further.

test/unit/Dredd-test.js
```diff
@@ -1,759 +1,94 @@
...
-
-
-function compareLocation(ad1, ad2) {
-  if (ad1.location < ad2.location) { return -1; }
-  if (ad1.location > ad2.location) { return 1; }
-  return 0;
-}
-
-
-describe('Dredd class', () => {
```
honzajavorek: I've deleted most of the tests, because I'm not sure they were useful at all. I'm not fan of unit tests based on stubbing and mocking everything around. I think such tests aren't very valuable. Unfortunately Dredd's test suite is full of such tests.

My strategy would be to push as much logic as possible into pure, nicely testable functions, which can get verified by unit tests. Then to test the rest, the glue, by integration tests. Therefore I'm adding to the integration tests rather than stubbing and mocking everything here.

test/unit/CLI-test.js
```diff
@@ -284,77 +284,6 @@ describe('CLI class', () => {
     });
   });
 
-
-  describe('when configuration was saved', () => {
```
honzajavorek: This test is supposed to verify that CLI arguments can override dredd.yml arguments. I don't think this is a valuable test, as it:

- tries to unit test an integration point, at the cost of stubbing and mocking everything (see my other comment)
- tests configuration / CLI behavior using the `Dredd` class, which isn't really related to the subject under test much

I think this should be integration-tested together with revamping the configuration mechanics, not unit-tested here and definitely not tested this way.

lib/TransactionRunner.js
```diff
@@ -42,19 +42,43 @@ class TransactionRunner {
   }
 
   run(transactions, callback) {
-    logger.debug('Sorting HTTP transactions');
-    transactions = this.configuration.options.sorted ? sortTransactions(transactions) : transactions;
+    logger.debug('Starting reporters and waiting until all of them are ready');
+    this.emitStart((emitStartErr) => {
```
honzajavorek: The `emitStart` and `emitEnd` methods were originally part of Dredd.js, but I moved them to the transaction runner. All other events are emitted in the transaction runner, so I thought it's a good idea to streamline the concerns.

lib/TransactionRunner.js
```diff
@@ -42,19 +42,43 @@ class TransactionRunner {
...
+      logger.debug('Reading hook files and registering hooks');
+      addHooks(this, transactions, (addHooksError) => {
+        if (addHooksError) { return callback(addHooksError); }
+
+        logger.debug('Executing HTTP transactions');
+        this.executeAllTransactions(transactions, this.hooks, (execAllTransErr) => {
+          if (execAllTransErr) { return callback(execAllTransErr); }
+
+          logger.debug('Wrapping up testing and waiting until all reporters are done');
+          this.emitEnd(callback);
```
honzajavorek: The `emitStart` and `emitEnd` methods were originally part of Dredd.js, but I moved them to the transaction runner. All other events are emitted in the transaction runner, so I thought it's a good idea to streamline the concerns.

lib/Dredd.js
```diff
@@ -1,35 +1,94 @@
...
+            mediaType: apiDescription.mediaType,
+          },
+        }, transaction))
+    ))
+    // flatten array of arrays
+    .reduce((flatArray, array) => flatArray.concat(array), []);
+}
 
 
 class Dredd {
```
honzajavorek: What I did here:

- removed stuff, which didn't seem to be necessary anywhere: `this.tests`, `this.stats.fileBasedReporters`
- moved some things to configuration, mostly various coercing and preparation of `this.configuration`, and `this.logProxySettings()`
- moved `start` and `end` events to the transaction runner, as all other events are handled there so I thought it's a good idea to streamline the responsibilities
- I tried to make the constructor free of side-effects and any heavy lifting, and to move all that into `this.run()`

The Dredd class is now responsible for:

- coercing the configuration in the constructor
- requiring a module provided by `--require`
- configure reporters
- "API description locations to transactions" pipeline
- print annotations (and halt Dredd in case of errors)
- launch the transaction runner (where all the testing happens) with the list of transactions

lib/Dredd.js
```diff
@@ -40,215 +99,89 @@ class Dredd {
...
+    // ideally the property wouldn't be needed at all.
+    delete this.stats.fileBasedReporters;
+
+    this.logger.debug('Preparing API description documents');
+    this.prepareAPIdescriptions((error, apiDescriptions) => {
+      if (error) { callback(error, this.stats); return; }
+      this.configuration.apiDescriptions = apiDescriptions;
+
+      // TODO https://github.com/apiaryio/dredd/issues/1191
+      const annotationsError = handleRuntimeProblems(apiDescriptions, this.logger);
```
honzajavorek: I want to get rid of `handleRuntimeProblems()` as the next step after this PR, finally unblocking upgrading Dredd Transactions.

lib/TransactionRunner.js
honzajavorek: The code here is original, only start/end is added and the code is indented as a result of that.

lib/TransactionRunner.js
```diff
@@ -42,19 +42,43 @@ class TransactionRunner {
...
+        });
+      });
+    });
+  }
+
+  emitStart(callback) {
```
honzajavorek: Copy-paste from Dredd.js

lib/TransactionRunner.js
```diff
@@ -599,6 +623,14 @@ include a message body: https://tools.ietf.org/html/rfc7231#section-6.3\
       });
     });
   }
+
+  emitEnd(callback) {
```
honzajavorek: Copy-paste from Dredd.js

PR #1292, opened 2019-03-28
Require options where they're used

honzajavorek:

Streamlining how the codebase is written.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -13,7 +13,9 @@ const logger = require('./logger');
 const { applyLoggingOptions } = require('./configuration');
 const { spawn } = require('./childProcess');
 
+const dreddOptions = require('./options.json');
 const packageData = require('../package.json');
+
 
 class CLI {
   constructor(options = {}, cb) {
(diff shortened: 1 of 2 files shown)


lib/CLI.js
```diff
@@ -13,8 +13,10 @@ const logger = require('./logger');
 const { applyLoggingOptions } = require('./configuration');
 const { spawn } = require('./childProcess');
 
+const dreddOptions = require('./options.json');
```
kylef: I don't think the extension here is nessecery, in same way how you don't specify `.js` on the other sources. We can omit for consistency.

```suggestion
const dreddOptions = require('./options');
```

lib/Dredd.js
```diff
@@ -250,5 +249,5 @@ Is the provided path correct?
   }
 }
 
+
 module.exports = Dredd;
-module.exports.options = options;
```
kylef: Is this part of the public API / a breaking change?

honzajavorek: Hmm, you're right. While it's undocumented, this is breaking, I'll amend the message. I want it removed as I don't see any value in having it exported to users like this, the options definitions are internal to Dredd.

lib/CLI.js
honzajavorek: Oh, I thought it is necessary for `.json`. I don't know why.

honzajavorek:
Waiting for https://github.com/apiaryio/dredd/pull/1294 to get merged first.

PR #1293, opened 2019-03-28
refactor: upgrades "winston" package to v3

### LINKED ISSUE #1225: Upgrade winston to 3.x
There's an [upgrade guide](https://github.com/winstonjs/winston/blob/master/CHANGELOG.md). This should be done as part of the efforts started in https://github.com/apiaryio/dredd/pull/1099 and https://github.com/apiaryio/dredd/pull/1213

kettanaito:

Upgrades `winston` package to version 3. Follows the [migration guides](https://github.com/winstonjs/winston/blob/master/UPGRADE-3.0.md) to adjust existing usage for breaking API changes of the `winston` package.

* Uses `winston.createLogger` instead of `winston.Logger`
* Removes deprecated `colors` field from the logger config
* Uses` format` field with `format.colorize()` for logs' colors (retains the previously set colors mapping)

- Closes #1225 

- [x] To write docs
- [x] To write tests

### DIFF
--- a/test/unit/reporters/CLIReporter-test.js
+++ b/test/unit/reporters/CLIReporter-test.js
@@ -15,13 +15,17 @@ describe('CLIReporter', () => {
   let test = {};
 
   before(() => {
-    loggerStub.transports.console.silent = true;
-    reporterOutputLoggerStub.transports.console.silent = true;
+    [loggerStub, reporterOutputLoggerStub].forEach((logger) => {
+      logger.configure({ silent: true });
+    })
   });
 
   after(() => {
-    loggerStub.transports.console.silent = false;
-    reporterOutputLoggerStub.transports.console.silent = false;
+    // Is this really a good idea to mutate logger instances
+    // vs creating logger instances for tests using the same factory function?
+    [loggerStub, reporterOutputLoggerStub].forEach((logger) => {
+      logger.configure({ silent: false });
+    })
   });
 
   describe('when starting', () => {
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,17 +1,19 @@
-const winston = require('winston');
+const { createLogger, transports, format } = require('winston');
 
-module.exports = new (winston.Logger)({
+module.exports = new createLogger({
   transports: [
-    new (winston.transports.Console)({ colorize: true }),
+    new transports.Console({ colorize: true }),
   ],
   levels: {
     debug: 2,
     warn: 1,
     error: 0,
   },
-  colors: {
-    debug: 'cyan',
-    warn: 'yellow',
-    error: 'red',
-  },
+  format: format.colorize({
+    colors: {
+      debug: 'cyan',
+      warn: 'yellow',
+      error: 'red',
+    },
+  }),
 });
--- a/lib/reporters/reporterOutputLogger.js
+++ b/lib/reporters/reporterOutputLogger.js
@@ -1,8 +1,8 @@
-const winston = require('winston');
+const { createLogger, transports, format } = require('winston');
 
-module.exports = new (winston.Logger)({
+module.exports = new createLogger({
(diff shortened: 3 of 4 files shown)


kettanaito:
Hi, @honzajavorek. Could you please assist me in verifying these changes retain the previous behavior of both internal and external (repoter) loggers?

I've used `CLIReporter-test` to verify the changes on local. The test successfully passes after my changes. However, I would like to confirm on your side that the upgrade of the package went successfully on the usage level.

test/unit/reporters/CLIReporter-test.js
```diff
@@ -15,13 +15,17 @@ describe('CLIReporter', () => {
   let test = {};
 
   before(() => {
-    loggerStub.transports.console.silent = true;
-    reporterOutputLoggerStub.transports.console.silent = true;
+    [loggerStub, reporterOutputLoggerStub].forEach((logger) => {
+      logger.configure({ silent: true });
```
kettanaito: I'm particulraly insterested in how to verify this silent toggling.

honzajavorek:
Thanks! I'll take a look tomorrow 👍

test/unit/reporters/CLIReporter-test.js
honzajavorek: **Short:** I think it is in the test only to prevent the logger to print to the terminal during the tests. I wouldn't be bothered about it in this phase. If the `--loglevel=silent` option works correctly for Dredd (important), feel free to omit the silencing and leave the logging chatty in tests (nice to have).

**Long:** What I'm trying to achieve is to use winston for Dredd's "application logging" (roughly equal to stderr), and to replace winston completely for the reporters and any other "application output" (roughly equal to stdout). Also, I'd like the logger or the "output rendering component" (not in place yet) to be testable in better ways than it is now. Currently, magic with `proxyquire()` and `sinon.stub()`-s is necessary every time the tests want to verify what was displayed to the user.

The refactoring is hairy and complex, and especially gradually updating the tests to be able to test the same stuff even after changes to the logging infrastructure is tedious. For that reason, I consider having the logger silent during the tests as something _nice to have_ at this moment, and I omitted it many times in my previous PRs. The result is that the output of `npm test` is a bit messy now, but I count with the fact this gets figured out once we have a better way to test the logger (like dependency injection).

test/unit/reporters/CLIReporter-test.js
```diff
@@ -15,13 +15,17 @@ describe('CLIReporter', () => {
...
+    })
   });
 
   after(() => {
-    loggerStub.transports.console.silent = false;
-    reporterOutputLoggerStub.transports.console.silent = false;
+    // Is this really a good idea to mutate logger instances
+    // vs creating logger instances for tests using the same factory function?
```
honzajavorek: > Is this really a good idea to mutate logger instances

Nope 😄 But as per my previous comment, currently the logging isn't done in a way it would be easy to test. It is planned, but if you want to keep this PR scoped to Winston upgrade, you should restrain yourself from trying to solve this problem as well. We can (should!) do it as one of the next steps.

We can limit the mutations though by ommitting silencing of the logging in tests, which is not for the purpose of the test itself, but only for the purpose of not having the `npm test` output cluttered (like in this case).

kettanaito: Thanks for clarifying. I will remove this mutation and see the test output.

Let's keep this pull request scoped to prevent the complexity of the changes. We can refactor the usage of logging later, I'd be glad to.

kettanaito:
Thanks for the support, Honzo. I don't have the connection at the moment, so we can continue on this upgrade once I'm in the office. Hope that's fine.

kettanaito:
Closed in favor of https://github.com/apiaryio/dredd/pull/1303.

Issue #612, opened 2016-08-20
Possible problem in configuring apiary reporter by dredd.yml


honzajavorek:
User reported following `dredd.yml` configuration led to a public test report:

```
reporter: apiary
custom:
- "apiaryApiKey:<key>"
- "apiaryApiName:<name>"
```

The same values worked as environment values. We should verify whether there's not a bug in picking up these values from YAML.

artem-zakharchenko:
I can see that the current implementation in `configuration.js` attempts to grab `inConfig.custom`:

https://github.com/apiaryio/dredd/blob/a3d103c51d18f8ad30b0c41f79bc88079582091c/lib/configuration.js#L200-L203

However, that is a rool-level field which is explicitly set on the initial config object:

https://github.com/apiaryio/dredd/blob/a3d103c51d18f8ad30b0c41f79bc88079582091c/lib/configuration.js#L159-L161

That is *not* the `custom:` field you set in `dredd.yml`, that one would be `inConfig.options.custom`. However, we loop over `inConfig.custom`.

@honzajavorek Is this intentional, or is it a typo? I'd like to know what is the purpose of `inConfig.custom` prior to changing it. Thanks.

artem-zakharchenko:
Okay, I see now that it's the different field. `inConfig` and `outConfig` got my confused. So, in the end we do map over all options here:

https://github.com/apiaryio/dredd/blob/a3d103c51d18f8ad30b0c41f79bc88079582091c/lib/configuration.js#L209-L211

And there is `outConfig.options.custom` with the list of the provided credentials. The parsing/normalization of config seems to be correct.

honzajavorek:
I have changed this logic very recently. Is it possible I accidentally fixed this? 🤔

honzajavorek:
If we're not able to reproduce this and you inspected the code seems okay, we could close this. Under different circumstances, it would be nice to have a test for this, but I'd like us to rather focus on revamping the configuration soon so it's less confusing.

artem-zakharchenko:
The problem may be in the way we reference `outConfig.custom` or `outConfig.options.custom` further in the code. Also, as we have discussed, I believe some root-level properties on the default options Object aren't properly merged with the custom `dredd.yml` config, which doesn't look nice.

Let's refine what needs to be done in the short and long run, and gradually clean up config resolving.

artem-zakharchenko:
This should be fixed by #1311, need to verify.

honzajavorek:
[...]

PR #1311, opened 2019-04-03
Re-design Dredd configuration handling

artem-zakharchenko:

- To have a configuration handling as a composition of utility functions
- Removes deprecation warnings and drops support for the next CLI options: -c, -q/--silent, -l/--level, -t/--timestamp, -b/--sandbox, --hooksData.

#### Change log

- [x] Separates concerns during Dredd configuration handling: validation, normalization, logging
- [x] Performs config values coercing in a separate layer
- [x] Flattens `config.options` into `config`
- [x] Rewrites config handling functions to pure functions
- [x] Coercion
  - [x] Coerces supported options
  - [x] Coerces `l` and `level` to `loglevel`
  - [x] Coerces `data` to `apiDescriptions`
  - [x] Coerces `user` to Authorization `header`
  - [x] Coerces string value of `c` into boolean `color`
- [x] Validation
  - [x] Produces proper warnings upon deprecated fields
  - [x] Produces proper errors upon unsupported fields
  - [x] Account for options aliases (i.e. `c`, `l`, `q`) in deprecation schemas
- [x] ~~Loggers configuration adjusted~~ (changes are too big as-is, logging to be refactored in a separate pull request)
- [x] Remove leftovers
- [x] Tests
  - [x] Remove unused `_coerceRemovedOptions` and respective tests
  - [x] Retain the usage of nested `options` to ensure it's supported
  - [x] Add unit tests for functions that compose configuration handling
- [x] Set `server` key as `endpoint` key in the root level of config. Set `options.server` as-is

- Relates to https://github.com/apiaryio/dredd/issues/612

- [x] To write docs
- [x] To adjust test suits


artem-zakharchenko:
I need to check Dredd's docs for references to the dropped CLI options. Will mark the respective checkbox in the PR when it's done.

artem-zakharchenko:
Makes sense, @honzajavorek.

Then I will take this opportunity to adjust how validation and normalization of the config is performed currently. I would like to refrain from mutating the config object, and instead produce the next "resolved" config.

honzajavorek:
Go for it 🚀

honzajavorek:
Heads up - I found bug in how the inlined API descriptions are being loaded in the config. This is a fix: https://github.com/apiaryio/dredd/pull/1289/commits/be36252e58e899dc44d29e5ede64a266db6eef64

The `apiDescriptions` property gets accidentaly overriden by `coerceRemovedOptions()`, while it should be the other way around.

I have integration tests for API descriptions loading deep inside https://github.com/apiaryio/dredd/pull/1289/ as well, but unfortunately they're dependent on other changes.

lib/configuration.js
```diff
@@ -50,150 +54,225 @@ function applyLoggingOptions(options) {
...
+ * Returns the errors and warnings relative to the given config.
+ */
+const validateConfig = config => ({
+  errors: flushMessages(deprecationErrors, config),
+  warnings: flushMessages(deprecationWarnings, config),
+});
+
+const sanitizeConfig = R.compose(
+  coerceOptions,
+  removeDeprecatedOptions
```
honzajavorek: I think the best behavior would be:

- If people run Dredd with an unknown option, Dredd fails fast with an error this is unknown configuration.
- If people run Dredd with a previously known but now deprecated option, Dredd fails fast with an error this option has been removed.
- If the configuration code can figure out that a provided value is invalid for its respective option (such as `--loglevel` where it's a limited enum of available levels or `--color/--no-color`, which should be boolean), it should fail fast with error it's an invalid option.
- As a result of the above point, if an option hasn't been removed (like `-l`), but it's semantics had changed, Dredd should be able to fail fast for invalid (deprecated) values.
- If figuring out the validity of the option is not cheap (e.g. figuring out whether a file exists), it is passed along and left up to subsequent Dredd code.

The question is whether we want to implement all that in this PR or whether it's fine to do just a part of it now.

test/unit/configuration-test.js
```diff
@@ -181,8 +200,11 @@ describe('configuration._coerceRemovedOptions()', () => {
     it('gets coerced to color set to boolean true', () => {
       assert.deepEqual(config, { options: { color: true } });
     });
+    it('produces no warnings', () => {
+      assert.lengthOf(coerceResult.warnings, 0);
+    });
     it('produces no errors', () => {
-      assert.deepEqual(coerceResult.errors, []);
+      assert.lengthOf(coerceResult.errors, 0);
```
honzajavorek: I used `deepEqual` as in this case it provides richer diff when the test doesn't pass (it tells you exactly what's wrong instead of just saying 1 != 0), but since everyone is correcting this and pointing it out in reviews, I'm giving up on it 😄 I guess it's not immediately obvious for anyone else, so probably not worth keeping.

lib/configuration.js
artem-zakharchenko: Fail fast approach is very applicable here. I was thinking about that, but I would first rewrite it to the state that would resemble how it worked previously, and then it's up to us to decide whether to pile up this pull request, or merge and deal with that in the next one.

test/unit/configuration-test.js
artem-zakharchenko: Oh, I see. Well, I had no hard opinion on this. The only reason I changed it is because errors/warnings assertions were different through test cases, while testing seemingly same output. I'm okay with using `deepEqual` for *both* of them, if that makes a better DX during test failure debugging.

honzajavorek: I'm fine with leaving lengthOf there. You're not the first one confused by the mixed usage of lengthOf and deepEqual.

lib/configuration.js
honzajavorek: Would it make sense to do it like this then?

1. PR for refactoring the config
2. PR for dropping Node 6 support (= breaking change) & demoting most of the softly deprecated options (errors) to being removed (warnings) and failing fast (= breaking change)

artem-zakharchenko: Yes, absolutely. Let me revert the nodeJS deprecation and focus on configuration composition at the moment.

artem-zakharchenko:
Updated the pull request with coercion of options.

It looks somewhat complicated at certain places, I will go to simplifying it after all tests pass. There are a few other options not migrated (i.e. `c` as a string value), and we can discuss whether we need something like `coerceWithWarning` to differentiate deprecated and removed options.

In a nutshell, I suggest to divide configuration options into three groups:

1. Supported options.
1. Deprecated options. Implying their usage *will get* deprecated in the next major release. Optional coercion may be performed for such options to prevent internals from using them.
1. Removed options. No coercion is performed, the usage of removed options produces an error and exits the process.

artem-zakharchenko:
Finished with rewriting the coercion logic. Tests pass 🎉.
Proceeding with the validation part.

lib/configuration.js
```diff
@@ -50,201 +53,256 @@ function applyLoggingOptions(options) {
...
+    )
+  )
+);
+
+const mapIndexed = R.addIndex(R.map);
+
+const coerceApiDescriptions = R.compose(
+  mapIndexed((content, index) => ({
+    location: `configuration.apiDescriptions[${index}]`,
+    content: R.when(R.has('content'), R.prop('content'), content),
```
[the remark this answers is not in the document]
honzajavorek: The intention is, externally, I'd like Dredd to have this as a plain array of strings:

```js
["... content ...", "... content ..."]
```

Internally, I'd like it to be a more complex structure, an array of objects:

```js
[
  { content: "...", location: "...", ... },
  { content: "...", location: "...", ... },
]
```

The `location` is assigned property, user cannot specify it. When the API descriptions are provided inlined in the configuration of the Dredd class, it is assigned to `configuration.apiDescriptions[original array index]`. If the API descriptions come from files or URLs, the location is set to `foobar.apib` or `http://example.com/foobar.apib`. I can imagine having `apiDescriptions` supported in `dredd.yml` as well, so perhaps we could have `dredd.yml` like this:

```yaml
apiDescriptions:
  - "... content ..."
  - "... content ..."
```

I think there's no reason to accept the more complex, internal structure on the input. If anything else than array of strings (or, for that matter, a single string) comes on the input, Dredd should error as misconfigured. I don't think that was the case currently, but if we're fixing this, I'd restrain from supporting the complex structure on input.

lib/configuration.js
```diff
@@ -50,201 +53,274 @@ function applyLoggingOptions(options) {
...
+  'hooks-worker-term-timeout': 5000,
+  'hooks-worker-term-retry': 500,
+  'hooks-worker-handler-host': '127.0.0.1',
+  'hooks-worker-handler-port': 61321,
+};
 
-  // Abort Dredd if there has been any errors
-  if (coerceResult.errors.length) {
+function resolveConfig(config) {
+  const inConfig = R.compose(
```
honzajavorek: I don't think we should call it `inConfig` anymore, it isn't the object on input of the `resolveConfig` function. Perhaps something along the lines of `configWithDefaults`?

lib/configuration.js
```diff
@@ -50,201 +53,274 @@ function applyLoggingOptions(options) {
...
+ * Any coercion will not be performed, as they are removed prior to coercion.
+ */
+const removeUnsupportedOptions = R.compose(
+  R.dissoc('q'),
+  R.dissoc('silent'),
+  R.dissoc('t'),
+  R.dissoc('timestamp'),
+  R.dissoc('blueprintPath'),
+  R.dissoc('b'),
+  R.dissoc('sandbox')
```
honzajavorek: `sandbox` didn't have a short option? 🤔

test/integration/openapi2-test.js
```diff
@@ -11,15 +12,19 @@ let output = '';
...
+  const dreddOptions = R.mergeDeepRight(defaultConfig, options);
+
+  new Dredd(dreddOptions).run((error) => {
     if (!finished) {
       finished = true;
       if (error ? error.message : undefined) {
         output += error.message;
       }
-      cb();
+      cb(error);
```
honzajavorek: I don't think this is wanted, because this will halt the tests. We might have tests inspecting the error as a valid outcome of the test. (Perhaps doesn't matter here in particular, but a similar idiom is on many other places where Dredd is called in tests).

test/integration/response-test.js
```diff
@@ -232,7 +232,7 @@ const Dredd = require('../../lib/Dredd');
     it('prints four warnings for each of the responses', () => assert.equal(runtimeInfo.dredd.logging.match(
       /HTTP 204 and 205 responses must not include a message body/g
     ).length, 4));
-    it('prints four failures for each non-matching status code', () => assert.equal(runtimeInfo.dredd.logging.match(
+    it('prints four failures for each non-matching status code', () => console.log(runtimeInfo) || assert.equal(runtimeInfo.dredd.logging.match(
```
honzajavorek: Is `console.log(runtimeInfo) || ` a leftover?

test/unit/Dredd-test.js
```diff
@@ -266,11 +263,11 @@ GET /url
         done();
       }));
 
-      describe('and I also set configuration.options.path to an existing file', () => {
+      describe('and I also set configuration.path to an existing file', () => {
         let localdredd;
         beforeEach(() => {
-          if (!configuration.options) { configuration.options = {}; }
-          configuration.options.path = ['./test/fixtures/apiary.apib'];
+          // if (!configuration) { configuration = {}; }
```
honzajavorek: Leftover?

lib/configuration.js
```diff
@@ -256,14 +332,17 @@ function applyConfiguration(inConfig) {
...
   applyConfiguration,
   applyLoggingOptions,
 
   // only for the purpose of unit tests
-  _coerceRemovedOptions: coerceRemovedOptions,
+  _normalizeConfig: normalizeConfig,
+  _validateConfig: validateConfig,
+  _resolveConfig: resolveConfig,
+  _coerceRemovedOptions: coerceDeprecatedOptions,
+  _coerceDeprecatedLevel: coerceDeprecatedLevelOption,
```
honzajavorek: Shouldn't we rename these?

test/unit/configuration-test.js
```diff
@@ -171,405 +171,226 @@ describe('configuration.applyLoggingOptions()', () => {
 
 describe('configuration._coerceRemovedOptions()', () => {
```
honzajavorek: The function doesn't exist anymore, right? Should we rename the tests?

test/unit/configuration-test.js
```diff
@@ -614,31 +430,49 @@ describe('configuration._coerceRemovedOptions()', () => {
...
+      assert.lengthOf(errors, 0);
     });
-    it('produces one warning', () => {
-      assert.lengthOf(coerceResult.warnings, 1);
+  });
+});
+
+describe('configuration._resolveConfig()', () => {
+  it('retains class instances', () => {
+    const normalizedConfig = configuration._resolveConfig({});
```
honzajavorek: Would be nice to also test whether providing a custom emitter works (if I'm reading the test correctly, it only verifies whether the default emitter gets through okay)

lib/configuration.js
```diff
@@ -50,201 +53,274 @@ function applyLoggingOptions(options) {
...
+  token => `Authorization: Basic ${token}`,
+  user => Buffer.from(user).toString('base64')
+);
+
+const updateHeaderWithUser = R.compose(
+  R.unnest,
+  R.adjust(0, getUserHeader),
+  R.values,
+  R.pick(['user', 'header'])
+);
```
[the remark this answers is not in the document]
artem-zakharchenko: `header` has been confusing to me as well. List variables must be named in plural, imo.

test/unit/configuration-test.js
artem-zakharchenko: Yes. I'm planning to add unit tests for those functional utils which now compose configuration handling. I have a few of them unstaged, will push once the pull request is cleaned and your comments are taken into account.

lib/configuration.js
```diff
@@ -50,201 +53,274 @@ function applyLoggingOptions(options) {
...
+  )
+);
+
+const coerceDeprecatedDataOption = R.when(
+  R.propSatisfies(R.complement(R.isNil), 'data'),
+  R.compose(
+    R.dissoc('data'),
+    R.over(
+      R.lens(
+        R.compose(
```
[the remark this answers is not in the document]
artem-zakharchenko: What do you think if we create a `configuration` directory to group all related utils? I'm all hands for breaking big compositions to consist from different functions to read properly. But I hate to put it into a single module, it makes it hard to read, defiling the purpose of readability.

test/unit/configuration-test.js
artem-zakharchenko: Agree. There are a few branches of configuration handling logic that were not tested previously.

artem-zakharchenko:
Thanks for review, @honzajavorek. I will address your comments and provide the changes.

1. Unit tests for granular configuration utils will be added.
2. Actually, we’ve removed the `options` nesting internally, but for tests I need to make sure we preserve the nesting to ensure that it won’t break anything for users who still use nested `options`.
3. Pull request description updated.
4. Will update the docs.
5. Sorry, let’s just use `@todo` if you don’t mind.
6. Apart from being a preference, `compose` is closer to the mathematical functional composition. It also more comfortable to read and apply functions right-to-left, since currying also works RTL by design. I would choose `compose` over `pipe`.

lib/configuration.js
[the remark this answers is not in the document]
honzajavorek: https://github.com/apiaryio/dredd/issues/1327

lib/configuration.js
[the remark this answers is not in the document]
honzajavorek: Sure 🚀 I'm all for having a `configuration` folder with specific functions in their own files, unit tested, etc. I did similar things with `resolveLocations.js`, `resolveModule.js`, `resolvePaths.js`, etc.

lib/configuration.js
artem-zakharchenko: Will do. Also will add those functions that are subjected to unit tests.

honzajavorek:
Perhaps bikeshedding, but is `@todo` a JS Doc spec or something? I kind of find it hard to use for oneliners, like

```
const foo; // TODO: just this small thing
```

Also, there might already be some `// FIXME:`-s somewhere in the codebase. I don't think this is important though and I would never block the PR because of it. We can discuss this separately and then change it everywhere in Dredd to the one we choose, in a separate PR, so it's consistent.

lib/configuration/applyConfiguration.js
```diff
@@ -0,0 +1,111 @@
...
+const logger = require('../logger');
+const getProxySettings = require('../getProxySettings');
+const applyLoggingOptions = require('./applyLoggingOptions');
+const validateConfig = require('./validateConfig');
+const { normalizeConfig } = require('./normalizeConfig');
+
+const defaultConfig = {
+  http: {},
+  server: null,
+  // emitter: new EventEmitter(), // TODO Merge + preserve prototype?
```
honzajavorek: Nitpick: When I think about it, I'd appreciate if the comment was more verbose. In case we keep it here for a while, it might get murky when we forget the context. Don't be afraid to overcomment in this case:

```js
const defaultConfig = {
  http: {},
  server: null,
  // TODO Merge + preserve prototype?
  // 
  // If the next line was uncommented, the subsequent merging machinery wouldn't be able to deal with
  // the fact the EventEmitter isn't a plain object. We worked around it by making the code to treat the event emitter
  // as a special case, but a more robust solution would be nicer.
  //
  // emitter: new EventEmitter(),
  ...
```

Just an example, perhaps it could be more brief 😄 I'm sometimes way too verbose with my writing.

In case of weird things, special cases, work arounds, etc., I'm all for writing tomes of wisdom for the next generations of readers into the code. Otherwise the suffering you went through to discover and solve the problem will be forgotten and someone else in the future will have to go through it again.

lib/configuration/applyConfiguration.js
```diff
@@ -0,0 +1,111 @@
...
+  'hooks-worker-connect-retry': 500,
+  'hooks-worker-after-connect-wait': 100,
+  'hooks-worker-term-timeout': 5000,
+  'hooks-worker-term-retry': 500,
+  'hooks-worker-handler-host': '127.0.0.1',
+  'hooks-worker-handler-port': 61321,
+};
+
+/**
+ * Flattens giving configuration Object, removing nested "options" key.
```
honzajavorek: Nitpick:

```diff
- giving
+ given
```

lib/configuration/applyConfiguration.js
```diff
@@ -0,0 +1,111 @@
+const R = require('ramda');
+const { EventEmitter } = require('events');
+const logger = require('../logger');
+const getProxySettings = require('../getProxySettings');
+const applyLoggingOptions = require('./applyLoggingOptions');
+const validateConfig = require('./validateConfig');
+const { normalizeConfig } = require('./normalizeConfig');
+
+const defaultConfig = {
```
honzajavorek: I wonder if this shouldn't be called `DEFAULT_CONFIG` so it's clearer in the code below this is a global static thing rather than a local variable

lib/configuration/applyConfiguration.js
```diff
@@ -0,0 +1,111 @@
...
+      + 'https://dredd.org/en/latest/how-it-works/#using-https-proxy'
+    );
+  }
+
+  return resolvedConfig;
+}
+
+module.exports = {
+  applyConfiguration,
+  resolveConfig,
```
honzajavorek: Should the `resolveConfig` be exported as equal to `applyConfiguration`, when the file is called `applyConfiguration.js`, or is it an util? Also, similar to my other comment, I think these isolated things should be possible to require as `const applyConfiguration = require('./applyConfiguration')`.

lib/configuration/normalizeConfig.js
```diff
@@ -0,0 +1,184 @@
...
+  )
+);
+
+const coerceDataToApiDescriptions = R.compose(
+  R.unnest,
+  R.values,
+  R.evolve({
+    data: R.compose(
+      R.map(([location, content]) => {
+        const nextDesdcription = (typeof content === 'string')
```
honzajavorek: typo: nextDesdcription

honzajavorek: and description is too generic word, should be `nextAPIdescription` or perhaps just `apiDescription` would be enough

test/integration/configuration/resolveConfig-test.js
```diff
@@ -0,0 +1,48 @@
...
+    it('contains default options', () => {
+      assert.hasAllKeys(config, Object.keys(_utils.defaultConfig).concat('emitter'));
+    });
+    it('overrides default options with custom ones', () => {
+      assert.deepEqual(config.path, ['./foo.apib']);
+    });
+    it('preserves "emitter" instance', () => {
+      assert.isNotNull(config.emitter);
+      assert.isNotNull(config.emitter.on);
+      assert.typeOf(config.emitter.on, 'Function');
```
honzajavorek: Perhaps `assert.instanceOf(config.emitter, EventEmitter)` could be a replacement for the three lines above?

test/integration/configuration/resolveConfig-test.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    describe('deep merges "custom" properties', () => {
+      it('preserves default "cwd" property', () => {
+        assert.equal(config.custom.cwd, process.cwd());
```
honzajavorek: I wouldn't bind the test to `process.cwd()`, since we already use `_utils.defaultConfig` in this test file, we could compare it to `_utils.defaultConfig.custom.cwd`

test/unit/configuration-test.js
honzajavorek: It makes sense to keep the test so we know we didn't break anything with the refactor, but we should still rename the test, perhaps at least to just "configuration" as it basically tests the "configuration" folder now.

test/unit/configuration/normalizeConfig-test.js
```diff
@@ -0,0 +1,218 @@
...
+      describe('coerces to "silent"', () => {
+        it('when given "silent"', () => {
+          assert.propertyVal(_utils.coerceDeprecatedLevelOption({ l: 'silent' }), 'loglevel', 'silent');
+        });
+      });
+
+      describe('coerces to "warn"', () => {
+        it('in all other cases', () => {
+          assert.propertyVal(_utils.coerceDeprecatedLevelOption({ l: null }), 'loglevel', 'warn');
+          assert.propertyVal(_utils.coerceDeprecatedLevelOption({ l: null }), 'loglevel', 'warn');
```
honzajavorek: These two are identical lines. Perhaps you wanted to test `null`, `warn`, `warning`, `foobar`, IDK

test/unit/configuration/normalizeConfig-test.js
```diff
@@ -0,0 +1,218 @@
...
+            apiDescriptions: [
+              {
+                location: 'filename.api',
+                content: 'FORMAT: 1A\n# Sample API\n',
+              },
+            ],
+          });
+        });
+
+        it('with both "data" and "apiDescriptions"', () => {
```
honzajavorek: 👍 👏

lib/configuration/applyConfiguration.js
artem-zakharchenko: No hard opinion on that :) We are not reusing it anywhere, so it's not global per say, it remains scoped to this module.

test/integration/configuration/resolveConfig-test.js
artem-zakharchenko: 🤦‍♂️ Indeed. Thanks for pointing out.

test/unit/configuration/normalizeConfig-test.js
artem-zakharchenko: These should be fixed, probably a stale state of pull request.

lib/configuration/applyConfiguration.js
artem-zakharchenko: Yes, good call. I've provided detailed description below, where we explicitly set `emitter` after merging with default config. I will mention it here as well.

lib/configuration/normalizeConfig.js
```diff
@@ -0,0 +1,184 @@
...
+);
+
+const normalizeConfig = R.compose(
+  coerceOptions,
+  removeUnsupportedOptions
+);
+
+module.exports = {
+  normalizeConfig,
+  _utils: {
```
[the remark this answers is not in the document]
artem-zakharchenko: Both ways make them exposed in the module's export. There is a fundamental question about the publicity of this module, as `normalizeConfig` is used internally only.

I am against modifying `normalizeConfig` because it's a *function*, and setting properties on it effectively modifies that function. Named exports keep modules immutable, and serve for access purposes.

Alternatively, we can move all utils into separate directly, and export/import from there. Tests would import utils directly, and main `normalizeConfig` module would import, but won't export them. That's a slight overkill, however, to bring such separation for tests alone.

lib/configuration/applyConfiguration.js
honzajavorek: We're exporting it and using it in the respective test file and I think the constant-like letter casing would convey better what the variable represents

test/unit/configuration/normalizeConfig-test.js
honzajavorek: Yup, you made the commits meanwhile

lib/configuration/applyConfiguration.js
artem-zakharchenko: At this point there is no usage for `resolveConfig`, so it can be treated as test-only util. However, it may be useable for cases when you want to get a configuration but omit related side-effects (such as applying it to loggers).

lib/configuration/normalizeConfig.js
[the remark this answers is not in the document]
honzajavorek: I see your point in having the functions immutable, but I think this isn't really a mutation which would matter, I perceive it more like a workaround, syntax sugar. Bascially, with imports, it's what the "default export" gives you.

My requirements would be developer experience and encapsulation of the code.

I think each module should be encapsulated in a way that it represents one function which is useful outside of that module, is named according to that function, and I can always require it as const fooBarFunction = require('./fooBarFunction.js');` without thinking and without knowing anything about its internals (blackbox with a suitable naming so I can infer what it does without reading the code). And it doesn't matter if that particular function is used just within a certain folder (configuration) or across Dredd (although I think folders should behave in a similar way, hiding things which shouldn't be used by outside code).

If that function composes from helpers relevant only to that function, those should be kept in a file together with it. So far all good. The problem lands when we need to unit test these helper functions (which we definitely should be doing).

I think the only way to retain the blackbox developer experience, the uniform export, and to have the helpers exported and unittested, is to have the "default export" and some hidden ones. While ES6 has a syntax for that, with requires we need a hack: assigning the unittested functions as properties to the main function, which effectively then acts as a default export.

Since we never do this in our code for any other purpose - modifying function properties, I'd say this is an acceptable convention to solve a real problem. Also, while it is mutation of the function instance, I think you could also take it as a way of [reflection](https://en.wikipedia.org/wiki/Reflection_%28computer_programming%29). You are exporting a function, which exposes its helpers in a "protected" property. These helpers belong to that function, they are its parts, so it doesn't seem wrong to me to expose them for the purpose of unit tests in a context of that function.

lib/configuration/applyConfiguration.js
honzajavorek: If it's not used now, I'd keep it as util. If it's needed on multiple places in the future, let's give it it's own file. I'd like to keep the exports as a single function (corresponding with the file's basename) if possible (things like `configuration/index.js` being exceptions).

lib/configuration/normalizeConfig.js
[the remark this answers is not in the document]
artem-zakharchenko: As long as injected util names are not function's native methods (`toString`, `name`, etc.), it should be fine. I will update the pull request.

test/integration/configuration/resolveConfig-test.js
```diff
@@ -12,11 +14,13 @@ describe('resolveConfig()', () => {
       },
     },
   });
+  const { config } = resolveConfig(withOverrides({}));
```
honzajavorek: I don't like how the tests order matters here and how some of the assertions reuse this global config while some of them use a different, local one. It feels like we're not isolating each test properly from each other.

I think each `describe` should have it's own `resolveConfig()` call and dedicated `config` variable, even if that feels like duplication (duplication in tests is often desirable).

test/integration/configuration/resolveConfig-test.js
```diff
@@ -27,9 +31,57 @@ describe('resolveConfig()', () => {
...
+    describe('replaces "server" with "endpoint"', () => {
+      it('removes "server" key', () => {
+        assert.notProperty(config, 'server');
+      });
+
+      it('sets value to "endpoint" key', () => {
+        assert.propertyVal(config, 'endpoint', 'http://127.0.0.1');
+      });
+
+      it('allows to set "options.server" command', () => {
```
honzajavorek: This `it` (and the subsequent ones) feels like it should rather be a `describe`, which has certain config as an input and asserts for both `server` and `endpoint`, so the expected behavior is clearer:

- server = abc ➡️ endpoint = abc, server = default
- server = abc, options.server = xyz ➡️ endpoint = abc, server = xyz
- options.server = abc ➡️ endpoint = default, server = abc
- endpoint = abc ➡️ endpoint = abc, server = default
- endpoint = abc, options.server = xyz ➡️ endpoint = abc, server = xyz
- endpoint = abc, server = xyz ➡️ endpoint = ???, server = default (??? in reality should never happen, might make sense to error and abort)

lib/configuration/applyConfiguration.js
```diff
@@ -122,10 +122,7 @@ function applyConfiguration(config) {
...
 
-module.exports = {
-  applyConfiguration,
-  resolveConfig,
-  _utils: {
-    DEFAULT_CONFIG,
-  },
-};
+applyConfiguration.resolveConfig = resolveConfig;
+applyConfiguration.DEFAULT_CONFIG = DEFAULT_CONFIG;
```
honzajavorek: I see your point about collisions with builtin function properties. What we did in other files was underscore dangling so it never happens:

```js
applyConfiguration._resolveConfig = resolveConfig;
applyConfiguration._DEFAULT_CONFIG = DEFAULT_CONFIG;
```

I wonder whether it's needed or not. Looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function I don't think we're likely to hit any of the prototype properties 🤔 Maybe we could remove the dangling from other files then (if you agree, we could do it in a separate PR and then we can remove https://github.com/apiaryio/dredd/blob/master/.eslintrc.js#L22-L25 as well).

honzajavorek: I see your point about collisions with builtin function properties. What we did in other files was underscore dangling so it never happens:

```js
applyConfiguration._resolveConfig = resolveConfig;
applyConfiguration._DEFAULT_CONFIG = DEFAULT_CONFIG;
```

I wonder whether it's needed or not. Looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function I don't think we're likely to hit any of the prototype properties 🤔 Maybe we could remove the dangling from other files then (if you agree, we could do it in a separate PR and then we can remove https://github.com/apiaryio/dredd/blob/master/.eslintrc.js#L22-L25 as well).

docs/usage-cli.rst
```diff
@@ -75,7 +75,7 @@ See below how sample configuration file could look like. The structure is the sa
    blueprint: api-description.apib
```
honzajavorek: ⏫ Ouch, we're not ignoring `__pycache__` folders and they were added to your commit by accident. Please remove the files from the commit, I'll issue a PR to prevent this, at once.

docs/usage-cli.rst
```diff
@@ -75,7 +75,7 @@ See below how sample configuration file could look like. The structure is the sa
...
    endpoint: "http://127.0.0.1:3000"
 
-.. note::
+.. warning::
```
honzajavorek: I might have confused you by pointing at random things on your display and saying what should be warning 😄 I think this should stay as `note`

docs/usage-js.rst
```diff
@@ -26,48 +26,42 @@ As you can see, ``dredd.run`` is a function receiving another function as a call
...
+     output: [],       // Array of Strings, filepaths to files used for output of file-based reporters
+     'inline-errors': false, // Boolean, If failures/errors are display immediately in Dredd run
+     require: null,    // String, When using nodejs hooks, require the given module before executing hooks
+     color: true,
      emitter: new EventEmitter(), // listen to test progress, your own instance of EventEmitter
      apiDescriptions: ['FORMAT: 1A\n# Sample API\n']
    }
 
+.. note::
+   The usage of nested `options` key is deprecated. Please list options under the root of the configuration.
```
honzajavorek: I think this should be `warning`

honzajavorek: I'm missing a sentence about the `server`. Perhaps something like:

```restructuredtext
The former top-level ``server`` property should be now specified as ``endpoint``, not to be confused with the :option:`--server` option.
```

Also, beware, reST does

```restructuredtext
``options``
```

not

```markdown
`options`
```

artem-zakharchenko: Adjusted.

docs/usage-cli.rst
artem-zakharchenko: Resolved. Thanks :)

test/integration/configuration/resolveConfig-test.js
artem-zakharchenko: I've broken test cases for `resolveConfig` more granularly 74091dcc3763bb5a990a71d5017754c2280db1cb

test/integration/configuration/resolveConfig-test.js
artem-zakharchenko: Broken down test suits for `resolveConfig` 74091dcc3763bb5a990a71d5017754c2280db1cb

docs/usage-cli.rst
artem-zakharchenko: Removed 0006046c118db32c08e50288923a12500fc5984e

honzajavorek:
I see all my comments addressed. What's the state of the short options handling? Otherwise I think it's good to go 🚀 

One nitpick, if you want to scope a commit, the parentheses are before the colon:

```diff
-  test: (resolveConfig) breaks down options test suits 
+ test(resolveConfig): breaks down options test suits 
```

This is optional, in Dredd there is no convention around scoping the conventional changelog commits.

Another nitpick, it would be nicer to rebase the branch and squash/drop stuff not to have the `__pycache__` in the git history at all.

artem-zakharchenko:
Comments addressed, handling of options aliases provided. I will rebase to get rid of `__pycache__`, and it should be ready to merge.

artem-zakharchenko:
Marked `@todo` notes now include links to the GitHub issues associated with them.

lib/configuration.js
```diff
@@ -21,16 +18,22 @@ function coerceToBoolean(value) {
...
     logger.transports.console.colorize = false;
     reporterOutputLogger.transports.console.colorize = false;
   }
 
   // Handling the 'loglevel' value
-  if (options.loglevel) {
-    const loglevel = options.loglevel.toLowerCase();
+  /**
+   * @todo Can use "applySpec" here to produce next options?
+   */
```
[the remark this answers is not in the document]
artem-zakharchenko: Let's discuss it in the dedicated issue :) It's but a thought from the top of my head.

lib/configuration/applyConfiguration.js
```diff
@@ -9,7 +9,7 @@ const { normalizeConfig } = require('./normalizeConfig');
 
 const DEFAULT_CONFIG = {
   http: {},
-  server: null,
+  endpoint: null,
```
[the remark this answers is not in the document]
artem-zakharchenko: I will add the default value for `server` today.

PR #1312, opened 2019-04-03
Move events from Dredd.js to the Transaction Runner

honzajavorek:

The `emitStart` and `emitEnd` methods were originally part of Dredd.js, but I moved them to the transaction runner. All other events are emitted in the transaction runner, so I thought it's a good idea to streamline the concerns of the components.

I didn't rewrite the code, only moved it around and indented what was necessary. The result is not meant to be beautiful at this stage.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/test/unit/Dredd-test.js
+++ b/test/unit/Dredd-test.js
@@ -1,5 +1,3 @@
-const bodyParser = require('body-parser');
-const express = require('express');
 const fsStub = require('fs');
 const proxyquire = require('proxyquire').noCallThru();
 const requestStub = require('request');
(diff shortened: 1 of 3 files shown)


test/unit/Dredd-test.js
```diff
@@ -588,112 +588,14 @@ GET /url
...
+    afterEach(() => dredd.transactionRunner.executeTransaction.resetHistory());
 
     it('should execute the runtime', done => dredd.run(() => {
-      assert.isOk(dredd.runner.executeTransaction.called);
+      assert.isOk(dredd.transactionRunner.executeTransaction.called);
       done();
     }));
   });
-
-  describe('#emitStart', () => {
```
honzajavorek: I removed these unit tests now as they're specific to Dredd.js and would be hard to move to the transaction runner tests. In fact I have integration tests replacing most of this file, covering the basic use cases, waiting in the https://github.com/apiaryio/dredd/pull/1289

PR #1314, opened 2019-04-03
Refactor and test requiring user-provided module

honzajavorek:

- [ ] To write docs
- [x] To write tests

### DIFF
--- a/lib/resolveModule.js
+++ b/lib/resolveModule.js
@@ -1,0 +1,10 @@
+const fs = require('fs');
+const path = require('path');
+
+
+module.exports = function resolveModule(workingDirectory, moduleName) {
+  const absolutePath = path.resolve(workingDirectory, moduleName);
+  return fs.existsSync(absolutePath) || fs.existsSync(`${absolutePath}.js`)
+    ? path.resolve(absolutePath)
+    : moduleName;
+};
--- a/lib/resolveModule.js
+++ b/lib/resolveModule.js
@@ -5,6 +5,6 @@ const path = require('path');
 module.exports = function resolveModule(workingDirectory, moduleName) {
   const absolutePath = path.resolve(workingDirectory, moduleName);
   return fs.existsSync(absolutePath) || fs.existsSync(`${absolutePath}.js`)
-    ? path.resolve(absolutePath)
+    ? absolutePath
     : moduleName;
 };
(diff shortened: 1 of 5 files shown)


lib/resolveModule.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
+const path = require('path');
+
+
+module.exports = function resolveModule(workingDirectory, moduleName) {
+  const absolutePath = path.resolve(workingDirectory, moduleName);
+  return fs.existsSync(absolutePath) || fs.existsSync(`${absolutePath}.js`)
+    ? path.resolve(absolutePath)
```
artem-zakharchenko: Why should we resolve already resolved path once more?

test/unit/resolveModule-test.js
```diff
@@ -0,0 +1,30 @@
...
+    assert.equal(
+      resolveModule(workingDirectory, 'requiredModule.js'),
+      path.join(workingDirectory, 'requiredModule.js')
+    );
+  });
+
+  it('resolves an installed module name', () => {
+    assert.equal(
+      resolveModule(workingDirectory, 'coffeescript/register'),
+      'coffeescript/register'
```
artem-zakharchenko: A minor thing, but maybe it's more future-proof to mock some generic module (i.e. `mocha`), the one that is unlikely to be removed as the library's dependency. I assume `coffeescript` is a legacy, so it's going to be removed in the future. It's nice not to hook up to legacy things in the tests.

honzajavorek: CoffeeScript is definitely a legacy as Dredd's direct dependency, but it is an existing JS transpiler (it's still developed and some people still use it). The `--require` option was introduced to mainly support transpilers, similar to `mocha --compilers js:babel-core/register`.

Any transpiler would do to verify the feature works, but `coffeescript` was already in dev dependencies from the past. Its objective advantage is it has zero dependencies itself, so using it to test the feature is very "cheap" (compared to e.g. `babel-core`, which has ~14 dependencies right now, and might require additional settings or packages to actually transpile something).

Also... it is a tiny little tribute to Dredd's origins... okay, with this one I admit I'm just making up things now 😄

lib/resolveModule.js
honzajavorek: That's a good question. Looks like a leftover from some furious changes in code. I'll play with the code for a while and get back to address this comment.

honzajavorek: Okay, verified it has no particular reason. Addressed in da645b2

PR #1315, opened 2019-04-03
Split dredd-test.js into multiple files

honzajavorek:

The dredd-test.js is a mess of multiple things. This splits it into three files:

- `test/integration/apiary-reporter-test.js` - integration tests for the `apiary` reporter
- `test/integration/dredd-test.js` - `Dredd.js` integration tests
- `test/integration/openapi2-test.js` - testing specifics of the OpenAPI 2 support

⚠️ The code is copy-paste. This PR doesn't change it, only splits 1 ugly test file into 3

- Partly addressing https://github.com/apiaryio/dredd/issues/1288
- Factored out from https://github.com/apiaryio/dredd/pull/1289/

- [ ] To write docs
- [x] To write tests


test/integration/openapi2-test.js
```diff
@@ -0,0 +1,152 @@
...
+      if (error ? error.message : undefined) {
+        output += error.message;
+      }
+      cb();
+    }
+  });
+}
+
+
+function record(transport, level, message) {
```
artem-zakharchenko: Not critical, however, during the upgrade to `winston@3` I've noticed we declare this `record` function multiple times, doing the very same operations. I suggest to consider declaring just one `record` function (i.e. in some test utils) to decrease the refactoring surface of any future logging upgrades.

```js
// Convert to high-order function to preserve call signature of winston's callback.
function record(output) {
  return (transport, level, message) => {...}
}
```

honzajavorek: Agree 👍 I'd like to consolidate the ways logging gets inspected in tests together with further improving the logging infrastructure after it was split between logs and output.

PR #1316, opened 2019-04-04
Use explicit variable names

honzajavorek:

Because I think `response` is better than `res`.

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/lib/performRequest.js
+++ b/lib/performRequest.js
@@ -5,7 +5,7 @@ const defaultLogger = require('./logger');
 
 
 /**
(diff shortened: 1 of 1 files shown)


kylef:
There's also a lot of "err" used everywhere instead of "error":

```
$ git grep '\berr\b' | wc
     459
```

honzajavorek:
Yes, it's everywhere. We can either rename it in bulk or we can try to get rid of it using async/await 😄

PR #1369, opened 2019-05-21
test: uses "hasAllKeys" for order-insensitive keys assertion

### LINKED ISSUE #1368: Order assertion of results keys in sanitization test
**Describe the bug**
In the current `sanitization-test.js` we assert that the validation result has all the specified keys:

https://github.com/apiaryio/dredd/blob/306528427176c6b6d3f7e2eed8a27afdd5b063b7/test/integration/sanitation-test.js#L497

However, we achieve that through `deepEqual`, which makes the order of keys matter.

**To Reproduce**
```bash
mocha test/integration/sanitation-test.js
```

**Expected behavior**
We should assert that `event[2].test.results` contains all the keys, without relying on the exact order of keys. I suggest to use `hasAllKeys` instead:

```js
assert. hasAllKeys(events[2].test.results, ['general', 'headers', 'body', 'statusCode']);
```

**What is in your `dredd.yml`?**

Doesn't influence the issue.

```
11.1.0
```

> Internally, this issue has occurred because the order of validation changed in `gavel`. However, this may be a good chance to re-consider certain tests to be bulletproof.

artem-zakharchenko:

Uses `hasAllKeys` for assertion of validation results object keys in sanitization tests. Makes assertion order-insensitive, improving the tests' sustainability. Also fixes a logical flaw where object's keys are asserted in specific order.

- Fixes #1368 

- [x] To write docs
- [x] To write tests

### DIFF
--- a/test/integration/sanitation-test.js
+++ b/test/integration/sanitation-test.js
@@ -455,7 +455,7 @@ describe('Sanitation of Reported Data', () => {
       assert.equal(events[2].test.status, 'fail');
       assert.include(events[2].test.results.general.results[0].message.toLowerCase(), 'fail');
     });
-    it('emitted test data results contain just \'general\' section', () => assert.deepEqual(Object.keys(events[2].test.results), ['general']));
+    it('emitted test data results contain just \'general\' section', () => assert.hasAllKeys(events[2].test.results, ['general']));
     it('sensitive data cannot be found anywhere in the emitted test data', () => {
       const test = JSON.stringify(events);
       assert.notInclude(test, sensitiveKey);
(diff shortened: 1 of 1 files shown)


test/integration/sanitation-test.js
```diff
@@ -494,7 +494,8 @@ describe('Sanitation of Reported Data', () => {
       assert.equal(events[2].test.status, 'fail');
       assert.include(events[2].test.results.general.results[0].message.toLowerCase(), 'fail');
     });
-    it('emitted test data results contain all regular sections', () => assert.deepEqual(Object.keys(events[2].test.results), ['general', 'headers', 'body', 'statusCode']));
+    it('emitted test data results contain all regular sections', () => assert.hasAllKeys(events[2
+    ].test.results, ['general', 'headers', 'body', 'statusCode']));
```
honzajavorek: Strange newline, but given the test is decaffeinated and not humanly formatted, this is a minor issue.

artem-zakharchenko: Fixed. Thanks :)

PR #1372, opened 2019-05-21
fix: uses explicit validation result sections list for logging output

### LINKED ISSUE #1371: Logging order in transaction runner depends on Gavel's validation order
**Describe the bug**
The order in which the validation result is logged directly depends on the order in which validation is executed in Gavel (see https://github.com/apiaryio/gavel.js/issues/169). This is not okay, since it creates a tight dependency between two libraries, making Dredd fail if Gavel suddenly returns validation results in a different order.

In the mentioned Gavel's issue I reason why the order must never be relied on, and Dredd should specify the required order within its logging.

**To Reproduce**

```
mocha test/integration/response-test.js
```

Execute the test suite and see it pass. Change the order of validators invocation in Gavel and see it fail.

**Expected behavior**
- No tight dependency between Gavel and Dredd order of validators
- Dredd specifies an explicit list of "sections" it outputs during the logging

```
11.1.0
```

artem-zakharchenko:

Introduces an explicit list of validation result sections to be present in the Dredd's transaction runner output. This allows Dredd to be independent from Gavel's validation order, and also makes it resilient to new Gavel releases.

- Fixes #1371

- [x] To write docs
- [x] To write tests

### DIFF
--- a/lib/TransactionRunner.js
+++ b/lib/TransactionRunner.js
@@ -573,15 +573,18 @@ include a message body: https://tools.ietf.org/html/rfc7231#section-6.3\
         let message = '';
         const object = gavelResult || {};
         let validatorOutput;
-        for (const sectionName of Object.keys(object || {})) {
-          // Section names are 'statusCode', 'headers', 'body' (and 'version', which is irrelevant)
+
+        // Order-sensitive list of validation sections to output in the log
+        const resultSections = ['headers', 'body', 'statusCode']
+          .filter(sectionName => Object.prototype.hasOwnProperty.call(object, sectionName));
+
+        resultSections.forEach((sectionName) => {
           validatorOutput = object[sectionName];
-          if (sectionName !== 'version') {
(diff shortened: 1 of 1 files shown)


lib/TransactionRunner.js
```diff
@@ -572,16 +572,20 @@ include a message body: https://tools.ietf.org/html/rfc7231#section-6.3\
...
         let message = '';
         const object = gavelResult || {};
+
         let validatorOutput;
-        for (const sectionName of Object.keys(object || {})) {
-          // Section names are 'statusCode', 'headers', 'body' (and 'version', which is irrelevant)
+
+        // Order-sensitive list of validation sections to output in the log
+        const resultSections = ['headers', 'body', 'statusCode']
+          .filter(sectionName => Object.prototype.hasOwnProperty.call(gavelResult, sectionName));
```
artem-zakharchenko: Since now we have an explicit list of sections we need to filter out those sections that are actually present in the Gavel's results.

PR #1374, opened 2019-05-21
Do not initiate hooks if there are no hookfiles

honzajavorek:

I believe it is a regression. @kylef had to work around it when upgrading Dredd in https://github.com/apiaryio/polls-api/pull/66/commits/4f66c3409757f241334473ed2d13fba0f5c7142a

- https://github.com/apiaryio/polls-api/pull/66

- [ ] To write docs
- [x] To write tests

### DIFF
--- a/lib/addHooks.js
+++ b/lib/addHooks.js
@@ -42,6 +42,11 @@ module.exports = function addHooks(runner, transactions, callback) {
     runner.hooks.transactions[transaction.name] = transaction;
   });
 
+  // No hooks
+  if (!runner.configuration.hookfiles || !runner.configuration.hookfiles.length) {
+    return callback();
+  }
+
   // Loading hookfiles from fs
   let hookfiles;
   try {
--- a/test/unit/addHooks-test.js
+++ b/test/unit/addHooks-test.js
@@ -74,12 +74,26 @@ describe('addHooks()', () => {
 
   it('sets transactionRunner.hooks.configuation', (done) => {
     const transactionRunner = createTransactionRunner();
+    transactionRunner.configuration.hookfiles = [
+      './hooks.js',
+    ];
 
     addHooks(transactionRunner, [], (err) => {
       assert.deepEqual(
         transactionRunner.hooks.configuration,
         transactionRunner.configuration
       );
+      done(err);
+    });
+  });
+
+  it('skips hooks loading when there are no hookfiles', (done) => {
+    const transactionRunner = createTransactionRunner();
+    transactionRunner.configuration.hookfiles = [];
+    transactionRunner.configuration.language = 'python';
+
+    addHooks(transactionRunner, [], (err) => {
+      assert.isUndefined(transactionRunner.hooks.configuration);
       done(err);
     });
   });


honzajavorek:
Wait, it's wrong. `if (!runner.configuration.hookfiles) {` shouldn't work for an array in JavaScript... but I did it test-first, so I'm wondering how it can work 🤔

lib/addHooks.js
```diff
@@ -42,6 +42,11 @@ module.exports = function addHooks(runner, transactions, callback) {
     runner.hooks.transactions[transaction.name] = transaction;
   });
 
+  // No hooks
+  if (!runner.configuration.hookfiles) {
```
artem-zakharchenko: You are right, this branch doesn't cover the scenario when there are no hooks. `unner.configuration.hookfiles` is an empty array `[]`, which resolves to `false` in this condition.

The code follow down, and after:

```js
let hookfiles;
  try {
    hookfiles = resolvePaths(runner.configuration.custom.cwd, runner.configuration.hookfiles);
  } catch (err) {
    return callback(err);
  }

// Override hookfiles option in configuration object with
// sorted and resolved files
runner.configuration.hookfiles = hookfiles;
``` 

After this you mutate the `configuration.hookfiles` with `hookfiles`. `resolvePaths` returns an empty array in case there were no hooks (empty array).

Probably, you could assert the length of hooks:

```js
if (runner.configuration.hookfiles && runner.configuration.hookfiles.length ===0) {
  return callback(err)
}
```

honzajavorek: Yes, fixing the code is easy. Fixing the test to actually test something will be tricky.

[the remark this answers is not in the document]
honzajavorek: I'm afraid the subsequent code needs some of the stuff `addHooks()` do, regardless of whether there are hooks or not. I'd keep it as a surgical fix at the moment.

test/unit/addHooks-test.js
```diff
@@ -83,4 +86,15 @@ describe('addHooks()', () => {
       done(err);
     });
   });
+
+  it('skips hooks loading when there are no hookfiles', (done) => {
+    const transactionRunner = createTransactionRunner();
+    transactionRunner.configuration.hookfiles = [];
+    transactionRunner.configuration.language = 'python';
```
artem-zakharchenko: Why to test on `python`?
It would spawn a hook handling process, if I'm not mistaken, but does it affect the behavior when no hooks are set?

If the behavior is different depending on whether it's javascript hooks or not, then we would need two tests respectively. But if it's the same, then maybe let's use the default options (javascript)?

honzajavorek:
@artem-zakharchenko Changed both the test and the code. I made sure the test fails without implementation. Without implementation if the `if` it actually fails even without the assert, with the same error @kylef reported (it tries to load Python hooks, because I've set the language).

test/unit/addHooks-test.js
honzajavorek: ...like this:

image

artem-zakharchenko: But failed attempt to execute not installed hooks handler means the hooks _were_ attempted to run, isn't it?

artem-zakharchenko: The exception on the screenshot above means the code reached this point:

https://github.com/apiaryio/dredd/blob/0da05fb9c6b0a6a799d5336f9d539eebb9913cf4/lib/addHooks.js#L76-L78

Which, according to the issue, it shouldn't.

honzajavorek: Correct, because this has been "photographed" at the moment when I didn't have the assert in the test in place and the `if` wasn't in place in the implementation. So it reached the line you pointed out and failed with a try to find Python hooks. That's exactly what happened to @kylef. And that's exactly what shouldn't happen.

When the assert line is in place, it still fails with `configuration` not being undefined. With the `if` in place the test passes.

honzajavorek: I was first trying to find a failing test, which replicates the erroneous behavior. Then I added the `if`, which fixes it.

PR #1379, opened 2019-05-24
feat: removes "gavel.isValid" calls

artem-zakharchenko:

- Removes [deprecated `gavel.isValid(https://github.com/apiaryio/gavel.js/issues/166)` method]() usage from Dredd.
- Consequentially, improves performance by validating transaction once, as opposed to twice previously (`gavel.isValid()` used to validate a transaction, but share no results with you).

- Changes are a part of Gavel upgrades and have no corresponding Dredd issue I'm aware of

- [x] To write docs (no need, no public API changes)
- [x] To write tests (tests adjusted)

### DIFF
--- a/lib/TransactionRunner.js
+++ b/lib/TransactionRunner.js
@@ -528,11 +528,11 @@ Not performing HTTP request for '${transaction.name}'.\
 
   validateTransaction(test, transaction, callback) {
     logger.debug('Validating HTTP transaction by Gavel.js');
-    logger.debug('Determining whether HTTP transaction is valid (getting boolean verdict)');
-    gavel.isValid(transaction.real, transaction.expected, 'response', (isValidError, isValid) => {
-      if (isValidError) {
-        logger.debug('Gavel.js validation errored:', isValidError);
-        this.emitError(isValidError, test);
+
+    gavel.validate(transaction.real, transaction.expected, 'response', (validateError, gavelResult) => {
+      if (validateError) {
+        logger.debug('Gavel.js validation errored:', validateError);
+        this.emitError(validateError, test);
       }
 
       test.title = transaction.id;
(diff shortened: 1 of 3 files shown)


lib/TransactionRunner.js
```diff
@@ -528,102 +528,97 @@ Not performing HTTP request for '${transaction.name}'.\
 
   validateTransaction(test, transaction, callback) {
     logger.debug('Validating HTTP transaction by Gavel.js');
-    logger.debug('Determining whether HTTP transaction is valid (getting boolean verdict)');
-    gavel.isValid(transaction.real, transaction.expected, 'response', (isValidError, isValid) => {
-      if (isValidError) {
-        logger.debug('Gavel.js validation errored:', isValidError);
-        this.emitError(isValidError, test);
+
+    logger.debug('Validating HTTP transaction (getting verbose validation result)');
```
artem-zakharchenko: It's questionable whether we need these two similar logs right now (lines 530 and 532). They seem to communicate the very same thing without bearing any extra info.

lib/TransactionRunner.js
```diff
@@ -528,102 +528,97 @@ Not performing HTTP request for '${transaction.name}'.\
...
+        logger.debug('Gavel.js validation errored:', validateError);
+        this.emitError(validateError, test);
       }
 
       test.title = transaction.id;
       test.actual = transaction.real;
       test.expected = transaction.expected;
       test.request = transaction.request;
 
+      const isValid = gavelResult ? gavelResult.isValid : false;
```
artem-zakharchenko: There are occasions when `gavelResult` is `null`. I believe the only scenario for this to happen is when `validateError` exists. It appears validation error doesn't terminate the Dredd process (which is expected, as there are multiple transactions to be validated, and invalid transaction is _expected_ result), but neither it stops the logic flow in this block.

@honzajavorek, please, do you think it's a correct behavior for a single transaction logic to continue if `validateError` exists?

honzajavorek:
Commit message:

```diff
- feat: removes "gavel.isValid" calls
+ perf: validates HTTP messages just once
```

This is perhaps the first time I see we can use `perf:`, btw 😅

lib/TransactionRunner.js
honzajavorek: I think we can remove it, it was only to distinguish isValid and validate steps.

lib/TransactionRunner.js
honzajavorek: Yes, this is correct. The (original) code isn't beautiful, but it works as it's supposed to work. This seems right.

artem-zakharchenko:
@honzajavorek :) I hope it's not the last one either.

I've renamed the commit message, but I would like to emphasize in it that it's going to be included in the next minor version. How would you approach this via semantic release?

honzajavorek:
I'm trying to figure out whether `perf:` bumps / goes to changelog, but I'm struggling to find it anywhere. Everything is so configurable that it's hard to find a definitive source of information.

honzajavorek:
I noticed you are trying to put a "scope" there. That can be done in a following way:

```
- perf: validates HTTP message only once (TransactionRunner)
+ perf(TransactionRunner): validates HTTP message only once
```

I'm going to dig deeper to find out about the `perf`.

artem-zakharchenko:
Unfortunately 

```
perf(TransactionRunner): validates HTTP message only once
```

is not valid per commit citizen spec:

```
scope must be lowercase
```

I don't find this useful if it distorts the file/scope where the changes happen.

honzajavorek:
Oh. I did investigation on perf and that doesn't seem useful either 😞 Looks like we should stick to `feat`. Regarding scopes, I never really started using them. It felt to me that it's redundant information as it's obvious from the commit diff where the change happens. It's the same like with descriptive commit messages, such as "changed A for B", when that's obvious as well from the diff (the message should explain why/what is the result of the change).

honzajavorek:
Closes https://github.com/apiaryio/dredd/pull/1378

PR #1427, opened 2019-07-09
docs: remove david dm

honzajavorek:

We use other tools now to monitor dependencies and the links to David stopped to work.

- [x] To write docs
- [ ] To write tests


honzajavorek:
Fixed. The Windows builds didn't pass yet, but they're not important as the failing check only runs on Travis CI.

Issue #1186, opened 2019-01-08
Future Dredd proposal


honzajavorek:
_Dredd reads the API description document, executes HTTP requests it finds, and validates the HTTP responses it receives._

- “Designing your API prior to its implementation makes it human-friendly”
- “Dredd enables you to design your API prior to its implementation”

------

I think Dredd's architecture should get more modular, decoupled. This is my idea on how it should look like:

### Direction

- [ ] Be active player in OAS3 adoption, make it a first-class citizen, use OAS3 examples in the docs with a priority over OAS2 examples
- [ ] As Dredd gets decoupled, it should have multiple commands - `dredd server` for testing server, `dredd client` for testing clients (aka client-testing mock server: see [this](https://github.com/netmilk?tab=repositories&q=element)), `dredd traffic` for testing arbitrary previously recorded traffic
- [ ] Explore the ability to test whole scenarios (comes with testing arbitrary recorded traffic)
- [ ] Explore the ability to use FaaS for hooks

### Support

- [ ] Write a blog post about how to use Dredd - Apiary blog, dev.to, docs (tutorial), ...
- [ ] Talk about Dredd at meetups and conferences
- [ ] Get a new set of stickers printed and give them away at meetups and conferences
- [x] Monitor Twitter for new articles and mentions, actively RT and reply - TweetDeck search is `(dredd api) OR (dredd openapi) OR (dredd oas)`

### Architecture

- [x] API description parsing is a separate library (done as `dredd-transactions/parse`), the input interface for the Dredd core is [API Elements](https://apielements.org/)
- [x] There is a separate library to turn API Elements into an internal sequence of expected transactions (done as `dredd-transactions/compile`)
- [ ] There is Dredd core, which takes a sequence of expected transactions (https://github.com/apiaryio/dredd/issues/995), performs the HTTP requests, and returns a sequence of real transactions, streamed
- [ ] There is Dredd CLI, which takes care of the command-line interface and provides API Elements together with configuration
- [ ] Dredd Hooks are a separate engine, which can transparently execute arbitrary code in JavaScript, other language, or delegate execution to FaaS
- [ ] All hooks are provided with the real transaction(s) to modify, with the expected transaction(s) to inspect (https://github.com/apiaryio/dredd/issues/1145), and with configuration (https://github.com/apiaryio/dredd/issues/215) 
- [ ] Transaction names are completely agnostic to the API description format and work as a query language rather than paths or pointers
[...]

Issue #784, opened 2017-05-16
Re-introduce integration test suite of hook handlers


honzajavorek:
- This issue sums up ideas in #644, #647, #672, #702, #744.
- The original test suite was removed in https://github.com/apiaryio/dredd/pull/782.

--------

Requirements for future implementation and other notes:

- The test suite should work with contributors's PRs. That means it cannot use any secret credentials. https://github.com/apiaryio/dredd/issues/644, https://github.com/apiaryio/dredd/issues/702
- The test suite should be time-effective, possibly running in parallel with the regular test suite. https://github.com/apiaryio/dredd/issues/647
- The test suite should run also on Windows. That means AppVeyor. https://github.com/apiaryio/dredd/issues/744, https://github.com/apiaryio/dredd/issues/702
- There are no dependent builds in Travis CI. Maybe Circle CI 2.0 would be a solution?

honzajavorek:
https://circleci.com/docs/2.0/workflows/

Issue #704, opened 2017-01-25
Split Dredd into multiple projects


honzajavorek:
For Dredd it would make sense to split into multiple projects:

- [`dredd`](https://github.com/apiaryio/dredd) - command line tool
- `dredd-core` - the core functionality of Dredd, without the command-line interface
- `dredd-hooks` - support for hooks, communication with hook handlers
- [`dredd-transactions`](https://github.com/apiaryio/dredd-transactions) - parses API description and outputs a list of HTTP transactions to execute

This would make easier to use Dredd also in other environments than CLI (imagine all kinds of integrations, such as [the one @XVincentX did for Visual Studio Code](https://github.com/XVincentX/vscode-apielements)).

honzajavorek:
When this happens, investigate whether [lerna](https://github.com/lerna/lerna) wouldn't help with managing. (Thanks @XVincentX for the tip.)

w-vi:
When you say multiple projects, do you mean multiple repos?

honzajavorek:
That is an implementation detail to be decided. I mean multiple packages in the first place.

honzajavorek:
I'm going to make this a ticket tracking our "monorepo" efforts, which are currently under-documented and scattered all over our repositories.

I think the initial ideas were formulated around this conversation https://github.com/apiaryio/dredd/issues/600#issuecomment-324887713, then it was further discussed in Apiary internally ([Slack link](https://apiary.slack.com/archives/C56PT7LDN/p1511284805000595) to the initial message, for employees) and also externally with the original Dredd author @netmilk. This issue was created to make the intention public, but I'm not sure it sums up all the reasoning sufficiently. Issues https://github.com/apiaryio/gavel.js/issues/126 and https://github.com/apiaryio/dredd-transactions/issues/252 were created, but unlinked from any other mentions of the decisions behind. There are additional mentions of the "future monorepo" in https://github.com/apiaryio/dredd-transactions/issues/105#issuecomment-346317514 or in https://github.com/apiaryio/dredd/issues/1186. The [v6](https://github.com/apiaryio/dredd-transactions/releases/tag/v7.0.0) of Dredd Transactions is one of the first steps in the effort, splitting parsing and compilation into two distinct modules (packages, if you want).

### Reasons for the monorepo and additional notes

- The split between `dredd` and `dredd-transactions` projects makes the development more difficult. To have those two under one roof removes overhead for both the maintainers team and contributors.
[...]

PR #1494, opened 2019-09-13
Switch Dredd to CircleCI

honzajavorek:

Checking a box from https://github.com/apiaryio/dredd/issues/1186. We wanted to do this for a long time already. Makes it easier to turn Dredd into a monorepo.

Documented in https://github.com/apiaryio/dredd/issues/1186, enables https://github.com/apiaryio/dredd/issues/784 and https://github.com/apiaryio/dredd/issues/704.

- [x] To write docs
- [x] To write tests


.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+
+
+jobs:
+  test-node12:
+    docker:
+      - image: circleci/node:12
+    steps:
+      - checkout
+      - npm-install
+      - run: npm run ci:test
```
kylef: These are all repeated steps below, I'd suggest using an reusable component:

See https://github.com/apiaryio/api-elements.js/blob/38fdacc56a1a7ef3a8e238e3852ccc987b673996/.circleci/config.yml for working example.

```yaml
aliases:
  - &defaults
    steps:
      - checkout
      - npm-install
      - run: npm run ci:test

jobs:
  test-node12:
    docker: [image: circleci/node:12]
    <<: *defaults

  test-node10:
    docker: [image: circleci/node:10]
    <<: *defaults
```

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+          key: js-deps-{{ checksum "package.json" }}
+      - run: npm install --no-save
+      - save_cache:
+          key: js-deps-{{ checksum "package.json" }}
+          paths:
+            - ./node_modules
+
+  pip-install-docs:
+    steps:
+      - run: sudo pip install --upgrade pip
```
kylef: Is this really needed? What's wrong with the pip version installed in the docker image `circleci/python:3.6`?

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+      - restore_cache:
+          key: py-deps-{{ checksum "docs/requirements.txt" }}
+      - run: "if [ ! -d ./venv ]; then python -m venv ./venv; fi"
```
kylef: Not sure there's much to change here, but wanted to point out that a venv created from `circleci/python:3.6` may not work with `circleci/python:3.7`, since the cache could allow older venvs to newer pythons if the Python version is updated.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+      - image: circleci/node:12
+    steps:
+      - checkout
+      - npm-install
+      - run: npm run e2e:openapi2
+
+  quality-checks:
+    # Using Python 3.6 as that's the version ReadTheDocs is running.
+    docker:
+      - image: circleci/python:3.6-node
```
kylef: Maybe even worth making the docker image for Python reusable so that there's only one place to update.

```yaml
aliases:
  - &docker-python
    docker: [image: circleci/python:3.6-node]

steps:
  quality-checks:
    <<: *docker-python
```

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
+version: 2.1
+
+
+commands:
+  npm-install:
+    steps:
+      - run: sudo npm install npm@6 --global
```
kylef: I think Node 8 onwards ships with NPM 6, so this step is perhaps not nessecery? I don't see a real reason to update packages from the docker image, this can lead to less reproducable build because an update to NPM 6 could cause different behaviour upon CI restarts whereas relying on the NPM from the docker image will always be the same.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+      - restore_cache:
+          key: js-deps-{{ checksum "package.json" }}
```
kylef: If I understand correctly, the npm install can happen on different versions of Node JS and the cache can be loaded from another version? If the test-node12 step saves cache before the restore cache line happens on test-node10 then any compiled dependencies for node12 may cause node 10 to fail.

scripts/commitlint.sh
```diff
@@ -1,16 +1,11 @@
...
 
-
-if [ ! -z "$TRAVIS" ]; then
-  git remote set-branches origin master
-  git fetch --unshallow --quiet
-  git checkout master --quiet
-  git checkout - --quiet
-  ./node_modules/.bin/commitlint-travis
+if [ ! -z "$CIRCLECI" ]; then
+  ./node_modules/.bin/commitlint-circle
```
kylef: Could `npx commitlint-circle` work instead?

.circleci/config.yml
kylef: If we did something like https://github.com/apiaryio/dredd/pull/1494#discussion_r324273941 perhaps a comment around there regarding this, sure it will trip someone up in the future.

PR #1497, opened 2019-09-16
(Archived) Adopt TypeScript

artem-zakharchenko:

- Adds Prettier
- Introduces basic TypeScript stack to Dredd
- Adjusts source code and tests to use ES6 syntax
- Provides a minimum of types annotations to basic utility functions

- Relates to #1186

- [x] To write docs (not applicable, except contribution docs)
- [x] To write tests
- [x] Adjust CircleCI
- [x] Adjust AppVeyor
- [x] Remove Rollup and related dependencies, because the build is performed by TypeScript compiler


.eslintrc.js
```diff
@@ -1,28 +1,20 @@
...
-        'functions': 'never'
-      }
-    ],
-
     // This is to allow a convention for exporting functions solely for
     // the purpose of the unit tests, see
     // https://github.com/apiaryio/dredd-transactions/pull/179#discussion_r206852270
     'no-underscore-dangle': 'off',
 
+    'import/prefer-default-export': 'off',
```
artem-zakharchenko: Allows `export const a = 'anything'` exports from files.

- Makes internals easier to export for testing
- Doesn't pollute the default export

artem-zakharchenko:
CircleCI build is passing. Switching to configure AppVeyor to include build step in the CI.

artem-zakharchenko:
@kylef, could you please review?

Mainly CI configs and build setup, but I'd be thankful for a few files I've migrated to TypeScript too. Note that as of now Dredd won't emit any type definitions because that's incompatible with `allowJs` option that enables js/ts codebase. We would need to rewrite all existing modules to TypeScript for it to emit definition files, it seems.

.circleci/config.yml
```diff
@@ -60,13 +64,17 @@ jobs:
...
       - npm-install
+      - run: npm run build
       - run: npm run e2e:apib
 
   test-e2e-openapi2:
     docker: [<<: *node12]
     steps:
       - checkout
+      - restore_cache:
+          key: build-{{ .Branch }}-{{ .Revision }}-node12
```
kylef: It might be worth updating `node12` string to use environment variable indicating the node version. I'm wondering about someone changing the node12 version somewhere and missing parts in the rest of the configuration file.

.eslintrc.js
```diff
@@ -1,28 +1,20 @@
...
     // Using 'console' is perfectly okay for a Node.js CLI tool and avoiding
     // it only brings unnecessary complexity
     'no-console': 'off',
 
-    // Node 6 does not support dangling commas in function arguments
-    'comma-dangle': [
-      'error',
-      {
-        'arrays': 'always-multiline',
-        'objects': 'always-multiline',
```
kylef: I noticed this rule has gone away in some of the diff below. Is the comma dangle behaviour pushed by prettifier or did you make a decision to cut it out?

.circleci/config.yml
artem-zakharchenko: Good catch. Also in the fact that this cache restoration is unnecessary. I've used it before when I wanted to store the build artifacts in cache, but after discussion with you I've removed any kind of build caching and perform build on each test job.

I will remove that `restore_cache` at all.

artem-zakharchenko: Removed at fd46b03.

.eslintrc.js
artem-zakharchenko: I have explicitly removed it. In combination with `prettier` ESLint config it won't affect anything, delegating proper comma dangling to Prettier.

lib/CLI.js
```diff
@@ -367,16 +419,32 @@ ${packageData.name} v${packageData.version} \
...
+            logger,
+            (key, value) => {
+              // ES6 modules interoperability preserves the module's default export
+              // under the "default" key. When attempted to stringify that,
+              // the process gets into infinite loop due to circular JSON structure.
+              // Explicitly ignoring the "default" key during the conversion to string.
+              return key === 'default' ? null : value;
+            },
+            2
+          )}`
```
kylef: You might want to include `\n` here, writing to file descriptor doesn't suffix with newline unlike `console.error`.

lib/CLI.js
```diff
@@ -367,16 +419,32 @@ ${packageData.name} v${packageData.version} \
 
   exitWithStatus(error, stats) {
     if (error) {
-      if (error.message) { logger.error(error.message); }
+      if (error.message) {
+        process.stdout.write(
+          `logger: ${JSON.stringify(
```
kylef: Why are we printing the logger out?

lib/CLI.js
```diff
@@ -367,16 +419,32 @@ ${packageData.name} v${packageData.version} \
 
   exitWithStatus(error, stats) {
     if (error) {
-      if (error.message) { logger.error(error.message); }
+      if (error.message) {
+        process.stdout.write(
```
kylef:
```suggestion
        process.stderr.write(
```

lib/Dredd.js
```diff
@@ -108,7 +144,7 @@ class Dredd {
       skipped: 0,
       start: 0,
       end: 0,
-      duration: 0,
+      duration: 0
```
kylef: Not to keen on removing the trailing commas because then when you add new entries to objects/arrays you have to include modification of other lines in the diff, which is why it is in our style guide.

lib/Hooks.js
```diff
@@ -1,76 +1,76 @@
...
+    this.before = this.before.bind(this)
+    this.beforeValidation = this.beforeValidation.bind(this)
+    this.after = this.after.bind(this)
+    this.beforeAll = this.beforeAll.bind(this)
+    this.afterAll = this.afterAll.bind(this)
+    this.beforeEach = this.beforeEach.bind(this)
+    this.beforeEachValidation = this.beforeEachValidation.bind(this)
+    this.afterEach = this.afterEach.bind(this)
+    this.log = this.log.bind(this)
+    ;({ logs: this.logs, logger: this.logger } = options)
```
kylef: Doesn't seem like the semi-colons are removed in the other files I've reviewed so far, why are they removed from other lines in this file and inserted at the start of this line instead of the end of the prior line?

lib/childProcess.js
```diff
@@ -1,24 +1,24 @@
...
 const ASCII_CTRL_C = 3;
 const IS_WINDOWS = process.platform === 'win32';
 const TERM_FIRST_CHECK_TIMEOUT_MS = 1;
 const TERM_DEFAULT_TIMEOUT_MS = 1000;
 const TERM_DEFAULT_RETRY_MS = 300;
 
-
 // Signals the child process to forcefully terminate
-function signalKill(childProcess, callback) {
+export const signalKill = (childProcess, callback) => {
```
kylef: What's the reason for changing from a function to closure here? Seems inconsistent with some of the code in prior files such as `function formatLocation(apiDescriptionLocation, annotationLocation) {` above.

lib/Dredd.js
artem-zakharchenko: Yes, completely agree. I've removed trailing commas due to Node6 compliance, but then I've realized that source code is not run in Node, the built version is. Will revert the trailing comma rule.

lib/CLI.js
artem-zakharchenko: Sorry, I don't know the previous intention. Perhaps @honzajavorek can clarify?

lib/Hooks.js
artem-zakharchenko: 1. I will align semicolons consistency and remove them from the source code altogether.
2. Because `({ logs: this.logs, logger: this.logger } = options)` is not a valid expression per formatting tools, so it appends an explicit semicolon in no-semi mode to let it be evaluated properly (and not the part of the previous expression).

lib/CLI.js
artem-zakharchenko: Although it seems like a valid point, I would refrain from making logic changes as the part of this pull request. If this is a serious concern, let's raise a separate issue to adjust it. OK?

lib/Dredd.js
artem-zakharchenko: Files formatting made consistent in b7911c81cfe0411bd399a9f5f6d43ee18e8c8492.

lib/CLI.js
kylef: According to this diff, this entire block is new in your branch?

lib/CLI.js
kylef: https://github.com/apiaryio/dredd/pull/1497/files#r329025441 perhaps I'm confused, but seems this wasn't there before.

Screenshot 2019-09-27 at 12 20 23

lib/CLI.js
artem-zakharchenko: Oh, you are right, my bad. I think I left it for debugging purposes. Will revert to `logger.error()`.

artem-zakharchenko: It appeared to be some debugging left-over. Removed in 7b15907.

lib/CLI.js
artem-zakharchenko: That was some left-overs, as turned out. Removed in 7b15907.

lib/CLI.js
artem-zakharchenko: That is no longer needed, as the whole expression has been removed in 7b15907.

.circleci/config.yml
```diff
@@ -18,31 +17,36 @@ commands:
...
   - &test-steps
     steps:
       - checkout
       - npm-install
+      - run: npm run build
       - run: npm run ci:test
 
-
 jobs:
+  build:
```
honzajavorek: Does it need to be tested separately if it's a part of all the other jobs?

honzajavorek: Besides, this job isn't included in the workflow, so it isn't currently being ran.

.circleci/config.yml
```diff
@@ -18,31 +17,36 @@ commands:
...
-  - &python-docs
-    # using Python 3.6 as that's the version ReadTheDocs is running
+  - &python-docs # using Python 3.6 as that's the version ReadTheDocs is running
     image: circleci/python:3.6-node
 
   - &test-steps
     steps:
       - checkout
       - npm-install
+      - run: npm run build
```
honzajavorek: Should the build be an explicit step or should it be implicitly included as something like `pretest` script?

.npmignore
```diff
@@ -0,0 +1,2 @@
+# Prevent publishing source code to NPM
+lib
```
honzajavorek: Use package.json's `files`, please - https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

lib/Dredd.js
```diff
@@ -1,105 +1,141 @@
...
-  }, (error, contents) => {
-    if (error) { callback(error); return; }
-
-    const apiDescriptions = locations
-      .map((location, i) => ({ location, content: contents[i] }));
-    callback(null, apiDescriptions);
-  });
-}
+  if (typeof options === 'function') {
+    ;[options, callback] = [{}, options]
```
honzajavorek: What's the purpose of the semicolon here?

lib/Hooks.js
honzajavorek: That's black sorcery. Can we replace all expressions starting with a semicolon with some boring JavaScript code?

```js
this.logs = options.logs
this.logger = options.logger
```

lib/Dredd.js
artem-zakharchenko: To support initially invalid JavaScript syntax. I kept an intention to do as less changes to the existing code base as possible.

.circleci/config.yml
artem-zakharchenko: I've tried to perform a build once and store it, but since its quite fast, we've decided with @kylef that such setup may be an overkill (attaching a workspace with the build takes multiple times longer than the build itself). I'm okay to revisit this, if you find it necessary.

lib/Hooks.js
artem-zakharchenko: Yes, sure. Moreover, we can replace them in the original code as one of the first PRs:

https://github.com/apiaryio/dredd/blob/91ceef1a1b74191761cafc5d11ccb17d4d3b1e75/lib/Hooks.js#L18

honzajavorek:
👍 makes sense. If we could have big PRs at least consisting of commits each containing one kind-of-atomic change (like semicolons), that would also make the review easier. This PR contained some back-and-forth changes in the commits, which discouraged me to go commit-by-commit.

artem-zakharchenko:
@honzajavorek, yeah, that's true. It was a discovery road for me, so some things went back and forth. Next pull request will be cleaner based on the knowledge I've gather during this implementation.

honzajavorek:
I completely understand, my PRs usually look very similar 🙂 When they grow big, I roll up my sleeves and dive into breaking them down and rebasing into something smaller.

When I was alone in the team, @kylef was always nice to never be too harsh about my big PRs, but he got me positively motivated by naturally procrastinating reviews on them. I knew that small PRs will get through faster 😄

artem-zakharchenko:
Closed in favor of #1530

Issue #1524, opened 2019-09-30
Adopt TypeScript


artem-zakharchenko:
## Motivation:

The current state of the repository is not contributor-friendly. It takes intense amount of time to figure out certain parts of the logic, and the absence of strong types makes the learning curve, as well as the refactoring process, much steeper.

I propose to integrate TypeScript into the build pipeline to move forward predictable and self-documented code base. 

## Benefits:

1. TypeScript provides build-time verification that your code operates as intended.
1. TypeScript makes code self-documented, which is much more reliable than any kind of on-demand generated documentation.
1. We can emit TypeScript definitions we use in Dredd to allow TypeScript users to use Dredd more reliably.
1. Code base is easier to understand with types and cross-references.
1. Tests are easier to write since you also write them with TypeScript.
1. Vastly improved refactoring, due to all the points above.

## Roadmap:

The first attempts to introduce this change were done in #1497, but due to the size of the changes it was decided to split it into several smaller pull requests.

artem-zakharchenko:
Build setup and tests are adopted, it's left to rewrite a few basic modules in TypeScript.

honzajavorek:
@artem-zakharchenko do you want to close this one as done or do you want to turn it into a

- [ ] check
- [ ] list

tracking progress file by file? Or is there a "minimal viable adoption" we could define and when it is reached, it closes this issue?

rrhodes:
Any updates on this issue?

PR #1525, opened 2019-09-30
Fixes syntactically invalid JavaScript

artem-zakharchenko:

- Certain parts look like survivors after decaffeinate, and represent invalid JavaScript code

- Necessary prerequisite for #1524 
- Problematic spots found in #1497 

- [x] To write docs
- [x] To write tests

### DIFF
--- a/lib/Dredd.js
+++ b/lib/Dredd.js
@@ -27,17 +27,19 @@ function prefixErrors(decoratedCallback, prefix) {
 
 
 function readLocations(locations, options, callback) {
-  if (typeof options === 'function') { [options, callback] = [{}, options]; }
+  const usesOptions = typeof options !== 'function';
+  const resolvedOptions = usesOptions ? options : {};
+  const resolvedCallback = usesOptions ? callback : options;
 
   async.map(locations, (location, next) => {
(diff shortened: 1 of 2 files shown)


artem-zakharchenko:
These look like all places that were found to be auto-formatted with prepending a semicolon when introduced Prettier. Please review, @honzajavorek.

lib/Dredd.js
```diff
@@ -27,17 +27,19 @@ function prefixErrors(decoratedCallback, prefix) {
 
 
 function readLocations(locations, options, callback) {
-  if (typeof options === 'function') { [options, callback] = [{}, options]; }
+  const usesOptionsFunction = typeof options === 'function';
```
honzajavorek: Perhaps this could be called `usesOptions` - a boolean saying whether the caller provided optional `options` argument or not:

```js
const usesOptions = typeof options !== 'function';
```

honzajavorek:
The commits are `fix:`. From user's point of view, is it really a fix or is it refactor?

artem-zakharchenko:
@honzajavorek, thanks! Provided the suggestions:

* Marked commits as `refactor:`
* Used `usesOptions` variable, and reversed the logic to remain the same as before refactoring

PR #1529, opened 2019-10-01
Adds TypeScript

artem-zakharchenko:

See original motivation in #1497 

- More granular introduction of #1497 
- Implements #1524 

- [x] To write docs
- [x] To write tests


artem-zakharchenko:
As the first step I move `lib/options.json` from the `lib` directory to the root directory.

This is necessary due to how TypeScript compiles the source code directory. Basically, it will output the built code in to the dist folder, but it won't process JSON files by default. There are two options how to approach JSON processing:

## Set `resolveJsonModule: true`

With this option set in `tsconfig.json`, TypeScript will treat JSON files imports as a regular dependency and will compile JSON files into dist directory alongside the other built modules.

**The issue:** I haven't found a way to exclude certain JSON files (tried `files`, `exclude`, globs), which results into TypeScript also compiling `package.json`, as Dredd imports in source code. Because of the root-level JSON dependency TypeScript recreates the entire directory structure starting from the most distant dependency, which results into the following folder structure upon build:

```
- build
  - package.json // gets copied
  - lib/ // here's the source directory, copied
    - Dredd.js
- lib/
  - Dredd.js
- package.json
```

## Do not process JSON

Move JSON files into the directory that would have the same import paths for both source and dist files (which means their parent directory). This way the build directory structure represents the source code directory structure 1-1:

```
- build/
  - Dredd.js
- lib/
  - Dredd.js
- options.json
- package.json
```

I find the second option to be the lesser evil and, perhaps, even making `options.json` more evident to be a global source of truth for source code and tests.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+  "compileOnSave": true,
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es2017",
+    "lib": ["es2017"],
+    "typeRoots": ["node_modules/@types"],
+    "rootDir": "./lib",
+    "outDir": "./build",
+    "moduleResolution": "node",
+    "esModuleInterop": true,
```
artem-zakharchenko: Fixes the imports of ES modules that have the "default" export. Doesn't require to write:

```js
const foo = require('module').default
```

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "allowJs": true,
```
artem-zakharchenko: Configuring JS compatibility mode for now, as there is no intention to rewrite the entire Dredd to TypeScript in a single pull request.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "allowSyntheticDefaultImports": true
```
artem-zakharchenko: Preserves default module resolution, so you don't have to write:

```js
import * as path from 'path'
```

> As the `path` module has no default export, TypeScript would require you to import its exports using wildcard. While that's a valid code, I'd lean toward preserving as much backward compatibility as possible at this early stage of adoption.

artem-zakharchenko:
The next step is to add a basic TypeScript configuration and introduce a build step.

First step does not contain any actual modules written in TypeScript, so it acts as a proxy of the existing source code. It does, however, allow to write the source code in ES6, which I'm about to add (mainly `require` -> `import`). **The source code compilation is at place**.

appveyor.yml
```diff
@@ -1,15 +1,16 @@
 environment:
-  nodejs_version: "10"
```
honzajavorek: This is 12 now after merging https://github.com/apiaryio/dredd/pull/1495

artem-zakharchenko: Will be updated after this branch is rebased against the latest `master`.

artem-zakharchenko:
Preparing the next chunk of changes together with mocha+ts setup, as rewritten to ES6 source code won't be digested by the test suites, resulting into failed CI. To prevent this, I will push both source code rewrite and mocha+ts setup altogether (in separate commits).

lib/index.js
```diff
@@ -0,0 +1,3 @@
+import Dredd from './Dredd';
+
+module.exports = Dredd;
```
artem-zakharchenko: I'm adding explicit entry module so it can be exported properly.

By default, TypeScript will compile the source code into CJS, but preserve the default/named exports. This means that if you wish to import a distributed Dredd you cannot do:

```js
const Dredd = require('dredd')
```

You would have to do:

```js
const Dredd = require('dredd').default
```

To solve this, an explicit entry helps. It exports the default export as `module.exports`, allowing to preserve the same require/import as before.

artem-zakharchenko:
Tests adjusted to run with source code being in ES6/TypeScript.

That's achieved using `ts-node` module that's required by `mocha` before test run. Tests use their own `tsconfig.json` extending the default one used for the library's build. There are a few quirks about imports and proxyquring, refer to the changed files for more details.

lib/index.js
honzajavorek: This should probably be a comment in code.

test/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   env: {
-    mocha: true,
-    node: true,
```
honzajavorek: Why are these changes needed?

artem-zakharchenko: Left-overs from copying the test directory. Fixed pushed in e895117

lib/index.js
artem-zakharchenko: Added in 2f1a89d.

Issue #1519, opened 2019-09-27
Flaky test/unit/HooksWorkerClient-test.js


honzajavorek:
**Describe the bug**

After migrating to CircleCI, we discovered a new flaky test:

```
  1) Hooks worker client
       when hooks handler server is running and not modifying transactions
         when 'beforeEach' hook function is triggered
           "before each" hook in "when 'beforeEach' hook function is triggered":
     Error: done() called multiple times
      at runner.hooks.(anonymous function).(anonymous function) (test/unit/HooksWorkerClient-test.js:779:68)
      at EventEmitter.messageHandler (lib/HooksWorkerClient.js:346:11)
      at Socket.handlerClient.on (lib/HooksWorkerClient.js:286:40)
      at addChunk (_stream_readable.js:263:12)
      at readableAddChunk (_stream_readable.js:250:11)
      at Socket.Readable.push (_stream_readable.js:208:10)
      at TCP.onread (net.js:601:20)
```

PR #1531, opened 2019-10-03
Makes Dredd a monorepo

artem-zakharchenko:

- To manage Dredd and friends more effectively

- A part of #1186 

- [x] To write docs
- [x] To write tests


artem-zakharchenko:
**Current state:** Managing docs, moved to the root directory, fixing references to fixtures from `hooks/index.rst`. There were some issues with `docutils` python dependency being not installed. Should be solved with moving docs to the root.

packages/dredd/package.json
```diff
@@ -0,0 +1,107 @@
...
+    "prettify": "prettier --write lib/**/*.js test/**/*.js",
+    "prettify:check": "prettier --check lib/**/*.js test/**/*.js",
+    "test": "mocha \"./test/**/*-test.js\"",
+    "test:debug": "mocha --debug-brk \"./test/**/*-test.js\"",
+    "e2e": "npm run e2e:apib && npm run e2e:openapi2",
+    "e2e:apib": "cucumber-js",
+    "e2e:openapi2": "cucumber-js --world-parameters=\"{\\\"apiDescriptionFormat\\\": \\\"application/swagger+yaml\\\"}\"",
+    "ci:lint": "npm run lint",
+    "ci:test": "npm test",
+    "ci:smoke": "bash ./scripts/smoke.sh",
```
kylef: I don't think this comment is about the PR itself, but once we have dredd-transactions in the repo too, we will want to alter the smoke tests to test with the in-repo dredd-transactions.

When the time comes for this, you can take a look at how I solved similiar in API Elements where "Fury CLI" uses the local in-repo libraries as part of the smoke test: https://github.com/apiaryio/api-elements.js/blob/dbd6aafbd50fcb2fab43eae9d612561147308377/.circleci/smoke.sh

pksunkara:
LGTM.

artem-zakharchenko:
@honzajavorek please note that this pull request also brings Yarn as the default package manager to Dredd. The reasoning is the support for "workspaces", which grants an ability to properly install shared dependencies between multiple packages in a monorepo.

Due to that, I'd also update the docs (mainly contributing guidelines) to mention the Yarn preference. What do you think?

honzajavorek:
@artem-zakharchenko 

> Yarn as the default package manager to Dredd

For development, right? So we're talking about the "internals" page in the docs. Sure, reasonable.

For Dredd users we should still recommend npm.

artem-zakharchenko:
Yes, of course. Publicly nothing changes, Yarn is preferred for contributing only, not to fix package managers during dependencies management and CI. Well, in that case you're right, nothing to update (except of contributing docs in a separate pull request).

appveyor.yml
```diff
@@ -3,14 +3,16 @@ environment:
 install:
   - ps: Install-Product node 12
   - 'npm -g install npm@6'
```
honzajavorek: I guess we can remove this line now

docs/conf.py
```diff
@@ -116,6 +116,9 @@
 
 linkcheck_ignore = [
     'https://crates.io/crates/dredd-hooks',  # https://github.com/sphinx-doc/sphinx/pull/5140
+    # Exclude Dredd repository links, because they are checked
+    # as a part of another script.
```
honzajavorek: Would be nice to mention which one - `_extensions/ghlink_check.py`

appveyor.yml
artem-zakharchenko: Removed in https://github.com/apiaryio/dredd/pull/1531/commits/9abe641a834d0bb9609dae2e2df8c38bd106d8f2.

docs/conf.py
artem-zakharchenko: Mentioned in https://github.com/apiaryio/dredd/pull/1531/commits/5cc3afda02db6774f0256bb5de262335528f5905.

artem-zakharchenko:
Experiencing a flaky test (#1519). Re-running the workflow...

Issue #267, opened 2015-08-28
HTTP `LINK` and `UNLINK` is not sorted


obihann:
According to the dredd documentation sorted is missing support for `LINK` and `UNLINK`:

```
 --sorted, -s         Sorts requests in a sensible way so that objects are not
                       modified before they are created. Order: CONNECT,
                       OPTIONS, POST, GET, HEAD, PUT, PATCH, DELETE, TRACE.
```

I suggest the following, and I would be willing to help implement it:

```
 --sorted, -s         Sorts requests in a sensible way so that objects are not
                       modified before they are created. Order: CONNECT,
                       OPTIONS, POST, GET, HEAD, PUT, PATCH, LINK, UNLINK, DELETE, TRACE.
```

w-vi:
@obihann good suggestion I'll leave the decision on @netmilk how to approach this including the #268 issue as they are related and may be solved in one go.

obihann:
@w-vi @netmilk thanks for the reply, glad you agree its a useful idea. Honestly this is causing me a bit of trouble on a current project and I'm more than likely going to have to fork dredd and implement my own solution, so if you would like we can discuss the best way to handle this and then when I make the change I can submit a pull request.

w-vi:
@obihann Can we postpone the discussion to the next week when @netmilk is around?

obihann:
That works for me. I'm currently skipping my tests that relate to LINk and UNLINK, however they can stay in that state for a few more days haha.

obihann:
Just a poke to bring this back on the table. @w-vi @netmilk

PR #1545, opened 2019-10-08
Rewrites simple modules to TypeScript

artem-zakharchenko:

- To slowly begin to rewrite Dredd internal modules to TypeScript
- Rewrites and simplifies `sortTransactions` module

- Follows #1524 

- [x] To write docs
- [ ] To write tests
- [x] Add unit tests for `sortTransactions` module


packages/dredd/lib/general.ts
```diff
@@ -0,0 +1,82 @@
...
+    bodyEncoding: BodyEncoding;
+  };
+  skip: boolean;
+  fail: boolean;
+
+  test: TransactionTest;
+}
+
+export interface TransactionRequest {
+  method: RESTMethod;
```
honzajavorek:
```suggestion
  method: HTTPMethod;
```

packages/dredd/lib/general.ts
```diff
@@ -0,0 +1,82 @@
+export enum RESTMethod {
```
honzajavorek:
```suggestion
export enum HTTPMethod {
```

packages/dredd/test/unit/sortTransactions-test.ts
```diff
@@ -0,0 +1,104 @@
+import R from 'ramda';
+import { expect } from 'chai';
+import sortTransactions from '../../lib/sortTransactions';
+import { Transaction, RESTMethod } from '../../lib/general';
+
```
honzajavorek: I don't see a similar test written in JS being removed. Does it mean you want to keep it to be sure nothing got broken, or that it doesn't exist and this part isn't tested?

honzajavorek:
If you add two methods, you can close this one https://github.com/apiaryio/dredd/issues/267

packages/dredd/test/unit/sortTransactions-test.ts
artem-zakharchenko: I couldn't find any existing test that would cover `sortTransactions`. To my knowledge this adds the missing unit test, so there is no JavaScript predecessor to remove. Please, point me if you know where this behavior may be tested already. Thanks.

packages/dredd/lib/general.ts
artem-zakharchenko: Resolved in 875fc45.

artem-zakharchenko:
@honzajavorek thanks for the review.

Are you speaking about adding `LINK` and `UNLINK` to `HTTPMethod`? Are those even valid methods per spec?

honzajavorek:
They are used with proxies.

artem-zakharchenko:
@honzajavorek I think we can cover LINK/UNLINK separately. Thanks!

PR #1550, opened 2019-10-09
Configures scripts and CI for monorepo

artem-zakharchenko:

- [ ] To write docs
- [ ] To write tests


.circleci/config.yml
```diff
@@ -92,16 +95,16 @@ jobs:
...
   release:
     # the GH_TOKEN and NPM_TOKEN environment variables are expected to be set
     # in https://circleci.com/gh/apiaryio/dredd/edit#env-vars
     docker: [<<: *node12]
     steps:
       - checkout
-      - <<: *install
-      - run: yarn ci:release
+      # - <<: *install
+      # - run: yarn ci:release
```
honzajavorek: My hunch tells me this commented out piece can possibly stay here for longer than what we'd anticipate. I'd appreciate if there's a comment why it's commented out (the decision behind it).

pksunkara: It's commented out because semantic release needs to be figured out how to work with lerna. You lose the monorepo benefits if your package.json doesn't have proper `version`. Semantic release needs `0.0.0`.

artem-zakharchenko: Added the comment. Thanks!

PR #1551, opened 2019-10-09
Adds "dredd-transactions" to the monorepo

artem-zakharchenko:
#### :rocket: Why this change?

- Adds `dredd-transactions` package under the Dredd's monorepo
- Adjusts the scripts and CI to apply for multiple projects in the monorepo

#### :white_check_mark: What didn't I forget?

- [x] To write docs
- [x] To write tests
- [x] To put [Conventional Changelog](https://dredd.org/en/latest/internals.html#sem-rel) prefixes in front of all my commits and run `npm run lint`

PR #1564, opened 2019-10-18
Fix dependencies bundling for yarn workspaces

honzajavorek:

This improves the hack which gets rid of protagonist (i.e. C++ compilation of drafter, the API Blueprint parser). When we moved dredd-transactions to the monorepo, where dependencies are managed by yarn workspaces, the hack to get rid of protagonist stopped to work. The hack relies on 'bundledDependencies', but those stopped to work. 'npm pack' cannot find the dependencies to bundle if they're in the root 'node_modules' directory. 'yarn pack' doesn't work either, it has bugs and it doesn't support 'bundledDependencies' or the workspaces very well - see https://github.com/yarnpkg/yarn/issues/6794

This change overcomes the limitation by improving the prepack.js script. It now iterates over the 'bundledDependencies' and symlinks them to the local 'node_modules' directory where 'npm pack' can pick them up. After packing is done, there's a new postpack.js script, which makes sure the symlinks get cleaned up. The resulting .tgz then contains the dependencies bundled - without protagonist.

https://github.com/apiaryio/dredd/pull/1551

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/packages/dredd-transactions/package.json
+++ b/packages/dredd-transactions/package.json
@@ -30,6 +30,7 @@
     "fury-adapter-apib-parser": "0.16.0",
     "fury-adapter-oas3-parser": "0.9.1",
     "fury-adapter-swagger": "0.27.2",
+    "rimraf": "3.0.0",
     "uri-template": "1.0.1"
   },
   "bundledDependencies": [
@@ -44,7 +45,6 @@
     "mocha": "6.2.1",
     "mocha-lcov-reporter": "1.3.0",
     "proxyquire": "2.1.3",
-    "rimraf": "3.0.0",
     "sinon": "7.5.0"
   },
   "keywords": [
--- a/packages/dredd-transactions/package.json
+++ b/packages/dredd-transactions/package.json
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "prepack": "node ./scripts/prepack.js",
+    "postpack": "node ./scripts/postpack.js",
     "build": "exit 0",
     "clean": "rimraf ./coverage",
     "lint": "eslint --ignore-path .gitignore .",
--- a/packages/dredd-transactions/package.json
+++ b/packages/dredd-transactions/package.json
@@ -30,7 +30,6 @@
     "fury-adapter-apib-parser": "0.16.0",
     "fury-adapter-oas3-parser": "0.9.1",
     "fury-adapter-swagger": "0.27.2",
-    "rimraf": "3.0.0",
     "uri-template": "1.0.1"
   },
   "bundledDependencies": [
@@ -45,6 +44,7 @@
     "mocha": "6.2.1",
     "mocha-lcov-reporter": "1.3.0",
     "proxyquire": "2.1.3",
+    "rimraf": "3.0.0",
     "sinon": "7.5.0"
   },
   "keywords": [
--- a/packages/dredd-transactions/scripts/prepack.js
+++ b/packages/dredd-transactions/scripts/prepack.js
@@ -4,6 +4,7 @@
 
 const fs = require('fs');
 const path = require('path');
+/* eslint-disable import/no-extraneous-dependencies */
 const rimraf = require('rimraf');
 
 const PACKAGE_DIR = path.resolve(__dirname, '..');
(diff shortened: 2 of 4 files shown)


artem-zakharchenko:
Theoretically, it's possible to use [Yarn custom resolutions]() to hijack which version of `protagonist` should be installed. It can work with 4th party dependencies too:

```
{
  "resolutions": {
    "fury-adapter-xxx/drafter/protagonist": "https://registry.npmjs.org/package-that-cannot-be-installed/-/package-that-cannot-be-installed-0.0.0.tgz"
  }
}
```

The question is whether Yarn supports such nested path (it's a 5th party dependency in the end). If it does, replacing the tarball that should get installed by a non-existing package should hack the packing and omit `protagonist` altogether. This will also affect the local installations of the package.

packages/dredd-transactions/package.json
```diff
@@ -29,6 +30,7 @@
     "fury-adapter-apib-parser": "0.16.0",
     "fury-adapter-oas3-parser": "0.9.1",
     "fury-adapter-swagger": "0.27.2",
+    "rimraf": "3.0.0",
```
pksunkara: This should still be a dev dep

packages/dredd-transactions/scripts/prepack.js
```diff
@@ -4,14 +4,60 @@
...
+ * the package with npm and uses 'bundledDependencies'. 'npm pack' is not able
+ * to find packages if they're put into the root 'node_modules' by yarn and
+ * wouldn't be able to bundle them.
+ *
+ * @param {string} dependencyName Dependency tree root
+ */
+function symlinkDependencyTreeToLocalNodeModules(dependencyName) {
+  const localDependencyPath = path.join(PACKAGE_DIR, 'node_modules', dependencyName);
+  if (!fs.existsSync(localDependencyPath)) {
+    fs.symlinkSync(`../../../node_modules/${dependencyName}`, localDependencyPath);
```
pksunkara: Maybe it's better to get the root of the project here, instead of doing `../../..` in case we change the directories later.

packages/dredd-transactions/scripts/prepack.js
```diff
@@ -4,14 +4,60 @@
...
-
-const drafterPackageJsonPath = path.resolve(path.dirname(__filename), '../../../', 'node_modules/drafter/package.json')
+writePackageJson(DRAFTER_PATH, drafterPackageData);
 
-// prettier-ignore
-fs.writeFileSync(drafterPackageJsonPath, json);
+// get rid of protagonist everywhere
+[
+  path.join(PACKAGE_DIR, 'node_modules', 'protagonist'),
+  path.join(PACKAGE_DIR, '..', '..', 'node_modules', 'protagonist'),
```
pksunkara: Same here about root project dir.

packages/dredd-transactions/package.json
artem-zakharchenko: Hm, per eslint `rimraf` should be in dependencies:

```
/home/circleci/project/packages/dredd-transactions/scripts/prepack.js
  7:16  error  'rimraf' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```

artem-zakharchenko: Which is clearly a mistake, since ESLint doesn't care what for the package is required. I suggest we disable eslint rule for `rimraf` in this case.

packages/dredd-transactions/scripts/prepack.js
artem-zakharchenko: Since `dredd-transactions` package is always the direct child of the `dredd/packages`, the only thing that is variable is the location of the `prepack.js` script, which casts module requires.

I don't think it's worthy to invest time into this, as `scripts` are purely technical, and touching them should raise expectations that things may break.

honzajavorek: Actually, the first argument of `fs.symlinkSync()` must be a relative path from the resulting symlink's location:

image


https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback

I wouldn't be too concerned about this as if things change, the smoke test is our insurance.

honzajavorek:
Thanks for the additional fixes. I'm merging this.

PR #1622, opened 2019-12-06
Updates Gavel for JSON Schema Draft 6/7 support

artem-zakharchenko:
> **Currently Dredd release is blocked by https://github.com/lerna/lerna/pull/2358**

- To support JSON Schema Draft 6/7 validation in Dredd

- Gavel implementation https://github.com/apiaryio/gavel.js/pull/355

- [x] To write docs
- [x] To write tests
- [x] Add an integration test asserting JSON Schema Draft 6/7 usage
- [x] Add an integration test with a single Boolean value JSON Schema to be inferred as Draft 7

### DIFF
--- a/packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
+++ b/packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
@@ -6,13 +6,13 @@ FORMAT: 1A
 
 # Machines collection [/machines/{id}]
     + Parameters
-        + id (number, `1`)
+        + id: 1 (number)
 
 ## Get Machines [GET]
 
 + Request (application/json)
     + Parameters
-        + id (number, `2`)
+        + id: 2 (number)
 
 + Response 200 (application/json; charset=utf-8)
 
--- a/packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
+++ b/packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
@@ -1,0 +1,21 @@
+FORMAT: 1A
+
+# Machines API
+
+# Group Machines
+
+# Machines collection [/machines/{id}]
+    + Parameters
+        + id (number, `1`)
+
+## Get Machines [GET]
+
++ Request (application/json)
+    + Parameters
+        + id (number, `2`)
+
++ Response 200 (application/json; charset=utf-8)
+
+    + Schema
+
+            true
--- a/packages/dredd/test/fixtures/json-schema-draft-7.apib
+++ b/packages/dredd/test/fixtures/json-schema-draft-7.apib
@@ -1,0 +1,32 @@
+FORMAT: 1A
+
+# Machines API
+
+# Group Machines
+
+# Machines collection [/machines]
+
+## Get Machines [GET]
+
+- Response 200 (application/json; charset=utf-8)
+
+    + Schema
+        
+            {
+                "$schema": "http://json-schema.org/draft-07/schema#",
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "type": {
+                            "type": "string",
+                            "enum": ["bulldozer"]
+                        },
+                        "name": {
+                            "not": {
+                                "type": "number"
+                            }
+                        }
+                    }
+                }
+            }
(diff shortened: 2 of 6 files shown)


artem-zakharchenko:
There appears to be something wrong with CircleCI as no jobs are being run.

kylef:
I think the commit message should be more in the line of `feat:` that `chore:`. Since this is adding features to Dredd, `chore` wouldn't warrant a release.

artem-zakharchenko:
@kylef nice catch, renamed the commit. Proceeding with adding the integration test to assert JSON Schema Draft 6/7 support.

packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
```diff
@@ -0,0 +1,21 @@
+FORMAT: 1A
+
+# Machines API
+
+# Group Machines
+
+# Machines collection [/machines/{id}]
+    + Parameters
+        + id (number, `1`)
```
kylef: This is a much older parameter syntax in API Blueprint from before it was aligned with MSON. If I recall correctly, this older synatx is undocumented in API Blueprint. Wondering where you found it? (perhaps in other fixtures?)

I'd recommend moving to the newer syntax:

```suggestion
        + id: 1 (number)
```

packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
```diff
@@ -0,0 +1,21 @@
...
+
+## Get Machines [GET]
+
++ Request (application/json)
+    + Parameters
+        + id (number, `2`)
```
kylef:
```suggestion
        + id: 2 (number)
```

packages/dredd/test/fixtures/json-schema-draft-7.apib
```diff
@@ -0,0 +1,32 @@
...
+        
+            {
+                "$schema": "http://json-schema.org/draft-07/schema#",
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "type": {
+                            "type": "string",
+                            "enum": ["bulldozer"]
```
kylef: We could use `const` instead of `enum` because const was added in draft 6 or 7 so it can ensure that this is treated as a newer draft since it won't be valid in draft 3/4.

```suggestion
                            "const": "bulldozer"
```

packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
artem-zakharchenko: Got it. Copy pasted from another blueprint, perhaps it was different intentionally. Will fix.

packages/dredd/test/fixtures/json-schema-draft-7.apib
artem-zakharchenko: That's a wonderful idea. Will change.

artem-zakharchenko: Changed in https://github.com/apiaryio/dredd/pull/1622/commits/dfd485938c2b764ff015b1acf28885e1af0a8f55.

packages/dredd/test/fixtures/json-schema-draft-7-boolean.apib
artem-zakharchenko: Fixed in https://github.com/apiaryio/dredd/pull/1622/commits/6534cf61609145345da38a0674c220a37d825204.

artem-zakharchenko:
Thanks for the review, @kylef. Should we disclose somewhere the support of new versions of JSON Schema? In Dredd docs, perhaps?

kylef:
@artem-zakharchenko Look's like it's documented at https://github.com/apiaryio/dredd/blob/6652b71b660aec13ab97ccbc61c02e1931f20f46/docs/how-it-works.rst#api-blueprint, we can add Draft 6/7 to the list there for point 1 of API Blueprint. The links for draft 6 and 7 are already in `docs/_links.rst` for some reason.

artem-zakharchenko:
@kylef JSON Schema Draft 6 and 7 support is added to "How It Works > API Blueprint" section (https://github.com/apiaryio/dredd/pull/1622/commits/e015c8e207ae6f5b4ddde5c25ef65fc774d4a739).

PR #1648, opened 2020-01-20
Adds nullable enum integration test

### LINKED ISSUE #507: Enums can't be nullable
Problems with `nullable` enums were reported by multiple users in https://github.com/apiaryio/mson/issues/61. We should investigate the problem.

artem-zakharchenko:

- Related issue seems to be fixed, but requires an integration test to be verified

- Closes #507 

- [ ] To write docs
- [ ] To write tests


artem-zakharchenko:
I'm not sure if that's the best way to write such test. I haven't found an example of integration test in Dredd that would test multiple responses (with actual bodies).

artem-zakharchenko:
@kylef, could you please look at this and verify if such test on Dredd side makes sense?

packages/dredd/test/integration/regressions/regression-507-test.js
```diff
@@ -0,0 +1,86 @@
...
+      });
+    });
+};
+
+describe('Regression: Issue #507', () => {
+  describe('given actual body matches an enum', () => {
+    let runtimeInfo;
+
+    before(async () => {
+      runtimeInfo = await bootstrap((app) => {
```
artem-zakharchenko: What if we restructure the API description document to use 1 request and 1 response? I've learned that Dredd would hit both req/res in that case, and we could use the same server instance.

PR #1651, opened 2020-01-24
test: adds OA2 produces/consumes integration test

### LINKED ISSUE #531: Dredd should support all +json media types, not just application/json
See https://github.com/apiaryio/dredd/pull/529#discussion_r67153232 for details, but basically, there are many places in Dredd where we rely purely on `application/json`. We should ensure that Dredd is able to treat all `+json` media types as JSON, consistently everywhere in its stack.

artem-zakharchenko:

Adds missing integration test for OA2 `produces`/`consumes` properties. 

- Closes #531 

- [x] To write docs
- [x] To write tests

### DIFF
--- a/packages/dredd/test/integration/regressions/regression-531-test.js
+++ b/packages/dredd/test/integration/regressions/regression-531-test.js
@@ -1,0 +1,62 @@
+import { assert } from 'chai';
+
+import { runDreddWithServer, createServer } from '../helpers';
+import Dredd from '../../../lib/Dredd';
+
+describe(`Sending and receiving "application/hal+json" according to produces/consumes in OpenAPI 2.0`, () => {
+  let runtimeInfo;
+
+  before((done) => {
+    const app = createServer();
+    app.get('/articles', (req, res) =>
+      res
+        .set({
+          'content-type': 'application/hal+json',
+        })
+        .json([
+          {
+            id: 1,
+            title: 'Creamy',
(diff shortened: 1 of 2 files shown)


packages/dredd/test/integration/regressions/regression-531-test.js
```diff
@@ -0,0 +1,62 @@
...
+      res.status(201).json({
+        id: 2,
+        title: 'Crispy schnitzel',
+        text: 'Prepare eggs...',
+      });
+    });
+
+    const dredd = new Dredd({
+      options: {
+        path: './test/fixtures/531-produces-consumes.yaml',
```
kylef: Should we use `path.join`? wondering if this will break on windows and other platforms where the path separator is not `/` but something like `\`.

artem-zakharchenko: There are other tests that provide `path` option like this. I'll check the source code if there's any `path.normalize(options.path)` or similar to be performed on the Dredd's side.

If not, let's raise this as a new issue and adjust all values of `path`. What do you think?

artem-zakharchenko: I've went through Dredd's codebase and haven't found a place where it would have normalized the `options.path`. Raising an issue: https://github.com/apiaryio/dredd/issues/1653

PR #1658, opened 2020-02-03
Release dredd to 12.2.1, dredd-transactions to 9.1.1

artem-zakharchenko:

- To release a next version of `dredd` and `dredd-transactions` that contain updated Fury adapters

- [ ] To write docs
- [ ] To write tests

### DIFF
--- a/packages/dredd/package.json
+++ b/packages/dredd/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dredd",
-  "version": "12.2.0",
+  "version": "12.2.1",
   "description": "HTTP API Testing Framework",
   "main": "build/index.js",
   "bin": {
@@ -38,7 +38,7 @@
     "chai": "4.2.0",
     "clone": "2.1.2",
     "cross-spawn": "7.0.0",
-    "dredd-transactions": "9.1.0",
+    "dredd-transactions": "9.1.1",
     "gavel": "^8.2.0",
     "glob": "7.1.5",
     "html": "1.0.0",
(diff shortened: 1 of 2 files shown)


artem-zakharchenko:
We should first update Dredd to include the latest parsing dependencies, and then proceed with releasing the next major version of Dredd that would contain `gavel@9` and removal JSON Schema Draft 3 support.

packages/dredd/package.json
```diff
@@ -38,7 +38,7 @@
     "chai": "4.2.0",
     "clone": "2.1.2",
     "cross-spawn": "7.0.0",
-    "dredd-transactions": "9.1.0",
+    "dredd-transactions": "^9.1.1",
```
kylef: Is this intentional adding the `^`?

artem-zakharchenko: No, that's been added automatically after running `lerna version`.

artem-zakharchenko: Changed that to an exact dependency version. Since Lerna bumps dependency versions of of a dependent package upon `lerna version`, it should not matter to include a range of dependencies or a fixed version number.

packages/dredd/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dredd",
-  "version": "12.2.0",
+  "version": "12.2.1",
```
kylef: Should this be a patch version? iirc we've removed JSON Schema Draft 3 support

artem-zakharchenko: `12.2.1` contains a deprecation warning upon using JSON Schema Draft 3, but the support is still there. Since we've discussed to update to the next major Dredd version that would include `gavel@9`, I'll bump these dependencies versions once they are released.

PR #1725, opened 2020-04-20
Update fury toolchain

kylef:
This brings support for OAS 3 default responses, and parsing of OAS 3 "servers" (this only affects Dredd in terms that the OAS document gets validated and warnings are emitted for servers).

This also brings in Drafter 5.0 as a transitive dependency, this includes changes to generated JSON Schema (now draft 7) and updated URI Template parser which causes extra warnings in the parse result which need test adaptation.


kylef:
@opichals good point, I've added 585ba18a to update ramda / ramda typings.

PR #1745, opened 2020-04-29
Update fury-adapter-apib-parser

kylef:
Brings in https://github.com/apiaryio/api-elements.js/pull/438


kylef:
@opichals had to change the prepack script to count for drafter package logic being in fury-adapter-apib-parser so that the smoke tests pass.

PR #1862, opened 2020-11-27
chore: bump dredd version to 14.0.0

artem-zakharchenko:
- Removes NodeJS@8 support (EOL).
- Updates dependencies (incl. security issues fixes).

### DIFF
--- a/packages/dredd-transactions/package.json
+++ b/packages/dredd-transactions/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dredd-transactions",
-  "version": "9.2.2",
+  "version": "10.0.0",
   "description": "Compiles HTTP Transactions (Request-Response pairs) from an API description document",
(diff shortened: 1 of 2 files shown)


packages/dredd-transactions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dredd-transactions",
-  "version": "9.2.2",
+  "version": "10.0.0",
```
kuba-kubula, line 3: Would that work as a dependency even though it is not released yet?

kuba-kubula (approved):

packages/dredd-transactions/package.json
artem-zakharchenko: Yes, that should work without issues as both packages will be published as a part of this release.

PR #1864, opened 2020-11-30
chore: logs symlink existence in "postpack" script

artem-zakharchenko:
Adds a log whether a listed symlink exists, so we could observe the `postpack.js` script behavior.

### DIFF
--- a/packages/dredd-transactions/package.json
+++ b/packages/dredd-transactions/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dredd-transactions",
-  "version": "9.2.2",
+  "version": "10.0.0",
   "description": "Compiles HTTP Transactions (Request-Response pairs) from an API description document",
   "main": "index.js",
   "engines": {
--- a/packages/dredd/package.json
+++ b/packages/dredd/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dredd",
-  "version": "13.2.0",
+  "version": "14.0.0",
   "description": "HTTP API Testing Framework",
   "main": "build/index.js",
   "bin": {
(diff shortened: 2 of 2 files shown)


packages/dredd-transactions/scripts/postpack.js
```diff
@@ -13,7 +13,12 @@ if (fs.existsSync(SYMLINKS_LOG)) {
     .split('\n')
     .filter(line => line.trim())
     .map(dependencyName => path.join(PACKAGE_DIR, 'node_modules', dependencyName))
-    .filter(symlinkPath => fs.existsSync(symlinkPath))
+    .map((a) => console.log(a) || a)
+    .filter(symlinkPath => {
+      const exists = fs.existsSync(symlinkPath)
+      console.log('checking if %s exists', symlinkPath, exists)
```
artem-zakharchenko, line 19: The issue is that symlink to a scoped package (i.e. `@apielements/apib-parser`) never resolves, even if it exists.

packages/dredd-transactions/scripts/prepack.js
```diff
@@ -9,8 +9,13 @@ const rimraf = require('rimraf');
 
 const PACKAGE_DIR = path.resolve(__dirname, '..');
 const SYMLINKS_LOG = path.join(PACKAGE_DIR, 'prepack-symlinks.log');
-const APIB_PARSER_PATH = path.join(PACKAGE_DIR, 'node_modules', '@apielements', 'apib-parser');
 
+// Resolve "@apielements/apib-parser" module location against both
+// root-level and relative "node_modules" directory.
+const APIB_PARSER_PATH = [
+  path.join(PACKAGE_DIR, 'node_modules', '@apielements', 'apib-parser'),
+  path.join(PACKAGE_DIR, '../..', 'node_modules', '@apielements', 'apib-parser')
```
artem-zakharchenko, line 17: With this addition it symlinks the `@apielements/apib-parser` without exceptions on my machine. I'm preserving the previous logic for backward-compatibility, yet it always installs `@apielements/apib-parser` on the root-level of the monorepo in all these scenarios:

- using `lerna bootstrap`
- explicitly installing `lerna add @apielements/apib-parser --scope=dredd-transactions`
- removing `yarn.lock` and trying both of the above.

PR #1866, opened 2020-11-30
chore: resolve "apib-parser" against root-level "node_modules"

artem-zakharchenko:


artem-zakharchenko:
We should always operate with a clean `prepack-symlinks.log` file when running the `prepack` script. If `prepack`/`postpack` fails we are risking of being left with a malformed symlinks log file that would block further executions of the command locally.

kuba-kubula (approved):

Issue #1621, opened 2019-12-06
Establish a release process


artem-zakharchenko:
With Dredd becoming a monorepo (#1531) we need to adjust the release process. We are still going to use semantic release, but now integrated with Lerna for publishing. Monorepo versioning kind was agreed to be independent.

## Pre-requisites

- https://github.com/lerna/lerna/pull/2358 (split tagging and committing in Lerna)

PR #1867, opened 2020-11-30
chore: add release docs

artem-zakharchenko:
Adds the release process documentation.

- Documents #1621

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,32 @@
+# Release process
+
+- [Release reference #1](https://github.com/apiaryio/dredd/pull/1626)
+- [Release reference #2](https://github.com/apiaryio/dredd/pull/1862)
+
+## 1. Create a release pull request
+
+```bash
+$ git checkout -b <user>/dredd-14.0.0
+$ npx lerna version --no-push --no-git-tag-version 
+$ git add .
+$ git push -u origin <user>/dredd-14.0.0
+```
+
+## 2. Post-merge process
+
+**Once the release pull request is approved and merged**, proceed using the following instructions:
+
+```bash
+$ git checkout master
+$ git pull
+
+# Publish the packages respecting their versions from "package.json".
+$ npx lerna publish from-package
+
+# Tag the packages manually.
+$ git tag -a -m dredd@12.2.0 dredd@12.2.0 
+$ git tag -a -m dredd-transactions@9.1.0 dredd-transactions@9.1.0
+
+# Push the new tags to master.
+$ git push --tags
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -23,9 +23,9 @@ $ git pull
 # Publish the packages respecting their versions from "package.json".
 $ npx lerna publish from-package
 
-# Tag the packages manually.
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,35 @@
...
+```bash
+$ git checkout -b <user>/dredd-14.0.0
+$ npx lerna version --no-push --no-git-tag-version 
+$ git add .
+$ git push -u origin <user>/dredd-14.0.0
+```
+
+## 2. Post-merge process
+
+**Once the release pull request is approved**, proceed using the following instructions:
```
kuba-kubula, line 17: Should `approved and merged`?

kuba-kubula:
```suggestion
**Once the release pull request is approved and merged**, proceed using the following instructions:
```

kuba-kubula (approved):

RELEASE.md
artem-zakharchenko: True. Will correct that 👍

RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+$ git push -u origin <user>/dredd-14.0.0
+```
+
+## 2. Post-merge process
+
+**Once the release pull request is approved and merged**, proceed using the following instructions:
+
+```bash
+$ git checkout master
+$ git pull
```
kuba-kubula, line 21: no `npm install` or `yarn install` post-checkout?

kuba-kubula: What I mean is that unless you don't bundle any packages, it'll work. But AFAIK Dredd or Dredd-Transactions have bundled one or two packages? So the install of packages might be required. But anyway. Good enough to merge this Docs. It's already a good improvement to "nothing" 👍

RELEASE.md
```diff
@@ -0,0 +1,32 @@
+# Release process
```
kylef, line 1: Perhaps it might make sense to put this together with the other internal contributions documentation? (https://dredd.org/en/latest/internals.html#contributing / https://github.com/apiaryio/dredd/blob/master/docs/internals.rst)

RELEASE.md
artem-zakharchenko: That's a good point. From what I see `npx lerna publish` calls `npm pack` on each package, and we have pre- and postpack scripts specified where needed:

https://github.com/apiaryio/dredd/blob/11e821a482745e89315e0832ed6c52c751755e73/packages/dredd-transactions/package.json#L10-L11

You'd have to have dependencies installed, that's true. If they are missing you'd get an exception. The risk is to bundle a mismatched version of the package. 

Do you suggest to add `yarn install` as an error-safe step to ensure properly bundled dependencies?

RELEASE.md
artem-zakharchenko: Would it be relevant? That documentation addressed both internal and external contributors. The latter won't be able to publish to NPM to finish the release process. We'd get more reliability in the process if only the internal contributors release Dredd (less changes of changes to master while amidst the release).

RELEASE.md
kuba-kubula: @artem-zakharchenko adding `yarn install` as a pre-requisite for the release? Yep, exactly.

PR #1895, opened 2021-02-04
fix(deps): update uuid deprecated

ipv4sec:
when i run `dredd` command , my terminal show 

```
(node:11604) DeprecationWarning: Deep requiring like `const uuidv4 = require('uuid/v4');` is deprecated as of uuid@7.x. Please require the top-level module when using the Node.js CommonJS module or use ECMAScript Modules when bundling for the browser. See https://github.com/uuidjs/uuid#deep-requires-now-deprecated for more information.
```

this PR just update uuid deprecated


artem-zakharchenko (approved):
This looks great. Thank you for fixing this, @ipv4sec!

Issue #2173, opened 2023-01-24
Issues on Paramaeters Example with Swagger.JSON


YosaRama:
**Describe the bug**
I got error for `no example` on my endpoint when I add example inside schema object, currently I used OpenAPI v.3.0.0, and it will be look something like this : 
![Screenshot 2023-01-24 at 11 35 36](https://user-images.githubusercontent.com/67961793/214207543-bf94864c-f39c-4e50-a9f0-337978484db5.png)

but when I take out `example` from schema (schema.example) to directly one level with schema, it will be work fine. here attachment for that :
![Screenshot 2023-01-24 at 11 37 18](https://user-images.githubusercontent.com/67961793/214207559-e5d5cb7e-a1dd-4187-b9bb-6478128bc446.png)

if look at swagger documentation, single `example` will be inside the `schema` and multiple `example` will be outside `schema` but with `examples` tag, how can I running this test with `example` inside `schema` instead of writing example outside of schema object ?

**To Reproduce**
I just called in a simple way `npx dredd swagger.json https://zoho-subscriptions-integration-ttzxdcebuq-as.a.run.app/`, and write example inside schema object

**Expected behavior**
Writing example inside schema object for OpenAPI v.3.0.0, same as documentation `Adding example` from swagger.io
![Screenshot 2023-01-24 at 11 44 45](https://user-images.githubusercontent.com/67961793/214208141-bb4eb2fe-aa7d-4c97-bbcb-f2545c9b7afb.png)

**What is in your `dredd.yml`?**
Not using dredd.yml, directly running on command line

**What's your `dredd --version` output?**
![Screenshot 2023-01-24 at 11 45 31](https://user-images.githubusercontent.com/67961793/214208211-3ef06bb3-dca0-423f-a6f6-ecdc58c0cd9f.png)

**Does `dredd --loglevel=debug` uncover something?**
![Screenshot 2023-01-24 at 11 47 34](https://user-images.githubusercontent.com/67961793/214208409-2e66df94-f531-4298-ac40-f7b098f27ca6.png)

PR #2214, opened 2023-05-04
fix: issue #2173 OpenApi3.0.0

NamesNotRick:
JSON parsing (validateParams) doesn't support the new OpenApiV3.0.0 JSON formatting
Issue #2173 

- [x] To write docs
- [x] To write tests

### DIFF
--- a/packages/dredd-transactions/compile/compileURI/validateParams.js
+++ b/packages/dredd-transactions/compile/compileURI/validateParams.js
@@ -5,32 +5,46 @@ module.exports = function validateParams(params) {
     let text;
     const param = params[paramName];
 
+    
     if (param.required && !(typeof param.example !== 'undefined' && param.example !== '') && !(typeof param.default !== 'undefined' && param.default !== '')) {
       text = `Required URI parameter '${paramName}' has no example or default value.`;
       result.errors.push(text);
     }
 
-    switch (param.type) {
-      case 'number':
-        if (Number.isNaN(parseFloat(param.example))) {
-          text = `URI parameter '${paramName}' is declared as 'number' but it is a string.`;
-          result.errors.push(text);
-        }
-        break;
-      case 'boolean':
-        if ((param.example !== 'true') && (param.example !== 'false')) {
-          text = `URI parameter '${paramName}' is declared as 'boolean' but it is not.`;
-          result.errors.push(text);
-        }
-        break;
-      default:
-        break;
+    if (param.schema && param.schema.example  && typeof param.example === 'string'){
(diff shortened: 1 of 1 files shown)


packages/dredd-transactions/compile/compileURI/validateParams.js
```diff
@@ -5,32 +5,46 @@ module.exports = function validateParams(params) {
...
-      case 'number':
-        if (Number.isNaN(parseFloat(param.example))) {
-          text = `URI parameter '${paramName}' is declared as 'number' but it is a string.`;
-          result.errors.push(text);
-        }
-        break;
-      case 'boolean':
-        if ((param.example !== 'true') && (param.example !== 'false')) {
-          text = `URI parameter '${paramName}' is declared as 'boolean' but it is not.`;
+    if (param.schema && param.schema.example  && typeof param.example === 'string'){
```
kylef, line 14: The `param` structure does not contain a `schema` property. See the implementation of [`compileParams`](https://github.com/apiaryio/dredd/blob/805a797f29915c33c4c6136e5ff0da28ed735837/packages/dredd-transactions/compile/compileURI/compileParams.js#L39) which the result is [passed to `validateParams`](https://github.com/apiaryio/dredd/blob/805a797f29915c33c4c6136e5ff0da28ed735837/packages/dredd-transactions/compile/compileURI/index.js#L38).

```js
return {
  [name]: {
    required: getRequired(memberElement),
    default: getDefault(valueElement),
    example: getExample(valueElement),
    values: getValues(valueElement),
  },
};
```

kylef (changes requested):
I don't believe this pull request does what you expect it does. As `schema` is not a property passed down by the caller. It's not quite that simple to implement schema support unfortunately.

Schema Object's inside a Parameter Object in OpenAPI 3 is not supported in the underlying parser and therefore Dredd will not be able to access it. Dredd solely deals with two internal formats: API Elements, and then Dredd's own format (this the `validateParams` is taking Dredd's own format). Dredd itself doesn't really interoperate with underlying description formats like API Blueprint, Swagger 2 or OpenAPI. Instead it uses API Elements parsers to return ["API Elements"](https://apielements.org/en/latest/) which provides a uniform structure for numerous description formats (so that Dredd does not have to deal with the differences between formats). Then dredd-transactions package converts API Element's structures from API Element's into Dredd transaction format. That is what the `compileParams` function is doing.

It would be appropriate to add this into the OpenAPI 3 parser used, there's an existing issue tracking this: https://github.com/apiaryio/api-elements.js/issues/226

Issue #2294, opened 2024-03-24
Is `dredd` dead? Please don't abandon this amazing project!


jfoliveira:
Hey @kylef @abtris @kuba-kubula @nadade @ApiaryBot 
I see you are the ones who reviewed/approved the latest merged PRs on this repo.

As dredd has no new releases since 2021, could you please point who is the repo maintainer and what's the state of this project?

**It seems abandoned**:
* No interaction on the open pull requests (48) and issues (209) at the time of this writing

Could you share your plans, point to the currrent maintainers and maybe tag issues as `good first issues` so other community members might be motivated to collaborate?

dblock:
I have fixed CI in https://github.com/apiaryio/dredd/pull/2295 and am willing to do more maintenance work if someone wants to help out a little with merging my and other PRs. We can fork the project of course if there's no activity here, but would rather not.

We are attempting to use dredd in https://github.com/opensearch-project/opensearch-api-specification, specifically https://github.com/opensearch-project/opensearch-api-specification/pull/286.

PR #2295, opened 2024-05-09
chore: remove Circle-CI, fix GHA

dblock:

Build goes to green again.

- Removes Circle-CI. Looks dead? Unused?
- Moves smoke tests to GitHub Actions.
- Run tests/lints/etc. on `pull_request`.
- Renames jobs to tests (plural) to be consistent.
- Locks sphinx doc dependencies to those compatible with v4.3.0.
- https://github.com/ungrim97/Dredd-Hooks doesn't have functioning CI anymore (Travis CI is RIP), so links in docs removed
- Fixes `yarn docs:lint`, but loses some anchors that no longer work and I couldn't find easy replacements for

See https://github.com/dblock/dredd/actions for everything passing.

#2294 

- [x] To write docs
- [x] To write tests


.github/workflows/run-smoke-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [12.x]
+
+    steps:
+    - uses: supercharge/mongodb-github-action@1.6.0
```
kylef: > supercharge/mongodb-github-action@1.6.0 is not allowed to be used in apiaryio/dredd.

We're not going to be able to use this, perhaps it is worth removing this file/smoke tests from CI. Arguably merging this would still be better than current situation without smoke tests.

dblock: Not a problem. Updated to use GitHub's services, I was lazy and copy pasting from an older project ;)

Passing build in https://github.com/dblock/dredd/actions/runs/9043948597/job/24852077761.

